PR #16, opened 2025-07-21
[CFX-3136] Use Viper for Config Handling

alevan10:
# RATIONALE

We had been using a custom written handler for pulling and saving config information. Switching to [viper](https://github.com/spf13/viper) makes more sense, since it was written to be used alongside cobra, our CLI framework of choice.

We've also taken the opportunity to start adding in unit tests using the [testify](https://github.com/stretchr/testify) framework.

## CHANGES
- Switch to using `viper` for reading, updating, and writing config file information
- Add testify and tests for config file generation and reading
- Add a few new tasks for dev environment generation


cmd/config/config_test.go
```diff
@@ -0,0 +1,83 @@
...
+
+	readYamlData := make(map[string]string)
+
+	err := ReadConfigFile("")
+	suite.NoError(err, "Expected no error when reading config file without a previous file") //nolint: testifylint
+
+	host := viper.GetString("host")
+	suite.Equal(host, readYamlData["host"], "Expected config file to have the same host")
+
+	token := viper.GetString("host")
```
c-h-russell-walker, line 81:
```suggestion
	token := viper.GetString("token")
```

cmd/config/config.go
```diff
@@ -0,0 +1,85 @@
...
+			return fmt.Errorf("config file must have .yaml or .yml extension: %s", filePath)
+		}
+
+		dir := filepath.Dir(filePath)
+		filename := filepath.Base(filePath)
+		viper.SetConfigName(filename)
+		viper.AddConfigPath(dir)
+	} else {
+		if err := CreateConfigFileDirIfNotExists(); err != nil {
+			panic(err)
```
c-h-russell-walker, line 73: Does it make sense to return this `err` here?

```suggestion
			return err
```

I see (what looks to me) the only place `ReadConfigFile` is called in `cmd/auth/auth.go` panics if it gets an error back from this function.

cmd/auth/auth.go
```diff
@@ -184,67 +94,65 @@ func verifyAPIKey(datarobotHost string, apiKey string) (bool, error) {
...
-				// Set the DataRobot API key to be an empty string
-				if err := setValueInConfigFile(DataRobotAPIKey, ""); err != nil {
-					panic(err)
-				}
-			} else {
-				fmt.Println("Exiting without overwriting the API key.")
-				return nil
-			}
+		if strings.ToLower(strings.Replace(selectedOption, "\n", "", -1)) == "y" {
+			// Set the DataRobot API key to be an empty string
```
c-h-russell-walker, line 129: Is this comment correct?
Sorry maybe I'm wrong but it looks like `currentKey` is _not_ an empty string at this point?

cmd/auth/auth.go
```diff
@@ -184,67 +94,65 @@ func verifyAPIKey(datarobotHost string, apiKey string) (bool, error) {
...
 	}
 
-	if fileInfo.Size() > 0 { //nolint: nestif
-		currentKey, err := GetAPIKey()
-		if err != nil {
-			panic(err)
-		}
+	currentKey := viper.GetString(DataRobotAPIKey)
+	if currentKey == "" {
+		panic("API key is empty, please log in to DataRobot first.")
```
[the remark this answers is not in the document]
c-h-russell-walker: Oh cool - just got to `waitForAPIKeyCallback` below - ha, closing comment thread.

c-h-russell-walker (changes requested):
Looking great.
Really like the newer looking code structure using viper - also great you have these test additions.

I think we should use those two respective native GoLang `Join` functions in a few places where we're either dealing with files/dirs or URLs

cmd/config/config.go
alevan10: ah that's a good point, thanks

Taskfile.yaml
```diff
@@ -32,6 +32,19 @@ tasks:
...
+    silent: true
+    desc: "Initialize development environment"
+    deps: [lint]
+    cmds:
+      - |
+        export GOBIN={{.BIN_DIR}}
+        echo "🔧 Initializing development environment…"
+        mkdir -p {{.BIN_DIR}}
+        echo "🛠️ Installing tools…"
+        go get github.com/stretchr/testify@latest
```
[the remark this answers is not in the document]
alevan10: I'm not sure `testify` is added by `lint`, is it? In either case, do we want linting and tests to be associated like that, or should we move the linting install process into this function and keep `task lint/test` as functions that only lint or test respectively?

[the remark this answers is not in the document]
c-h-russell-walker: Hmmm TBH I'm not sure here 🤷

alevan10:
> Looking great. Really like the newer looking code structure using viper - also great you have these test additions.
> 
> I think we should use those two respective native GoLang `Join` functions in a few places where we're either dealing with files/dirs or URLs

Yup, you're right, I'm not OS agnostic right now

cmd/auth/auth.go
c-h-russell-walker: Sorry this comment still says we're setting it to an empty string but is `currentKey` actually an empty string here?

yuriyhrytsyuk (approved):

c-h-russell-walker (approved):
Looks good! Thanks for the additional updates.

PR #18, opened 2025-07-23
[CFX-3283] Better file structure in cmd folder

yuriyhrytsyuk:
# RATIONALE


cmd/auth/constants.go
```diff
@@ -0,0 +1,14 @@
...
+// and its affiliates.
+// The copyright notice above does not evidence any actual or intended
+// publication of such source code.
+
+package auth
+
+const (
+	DataRobotURL    = "endpoint"
+	DataRobotAPIKey = "token"
+)
```
yuriyhrytsyuk, line 14: Putting it here for now. `internal/config` might be better place for this.

cmd/auth/auth.go
```diff
@@ -105,103 +98,9 @@ func writeConfigFile() {
...
-
-	return nil
-}
-
 func GetAPIKey() (string, error) {
 	// Returns the API key if there is one, otherwise returns an empty string
 	key := viper.GetString(DataRobotAPIKey)
 
 	return key, nil
 }
```
yuriyhrytsyuk, line 106: Leaving this here for now. internal/config might be a better place for this.

PR #19, opened 2025-07-23
[CFX-3136] Fix Path Joining and Load Config on `init`

alevan10:
# RATIONALE

While working on the template TUI for the CLI, I encountered a couple bugs I introduced that needed some fixing, as well as reading some suggestions that the `viper` config loading should happen in the CLI's `init` so that every package has access to its data.

## CHANGES
- Switch from `path.Join` to `url.JoinPath` for joining the base URL with `/applicationTemplates`

### DIFF
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -41,8 +41,9 @@ func init() {
 
 	err := config.ReadConfigFile("")
 	if err != nil {
-		panic(err)
+		log.Fatal(err)
 	}
+
 	RootCmd.AddCommand(
 		auth.AuthCmd,
 		templates.TemplatesCmd,


cmd/root.go
```diff
@@ -38,6 +39,10 @@ func Execute() error {
 func init() {
 	cobra.OnInitialize(initConfig)
 
+	err := config.ReadConfigFile("")
+	if err != nil {
+		panic(err)
+	}
```
yuriyhrytsyuk, line 45:
```suggestion
	err := config.ReadConfigFile("")
	if err != nil {
		log.Fatal(err)
	}

```

PR #24, opened 2025-08-15
[CFX-3417] Pretty up the "you can close this page"

yuriyhrytsyuk:
# RATIONALE

## CHANGES

Screenshot 2025-08-15 at 12 06 56


internal/assets/assets.go
```diff
@@ -0,0 +1,28 @@
+// Copyright 2025 DataRobot, Inc. and its affiliates.
+// All rights reserved.
+// DataRobot, Inc. Confidential.
+// This is unpublished proprietary source code of DataRobot, Inc.
```
c-h-russell-walker, line 4: Oh I just realized something likely totally unrelated to this PR - if this is OSS then should this line be different?

`// This is unpublished proprietary source code of DataRobot, Inc.`

It _is_ "published" right?

PR #26, opened 2025-08-22
[CFX-3405] TUI Testing Framework

alevan10:
# RATIONALE

In order to help with ensuring we don't break the TUI during development, we want to add in the `teatest` testing framework for `BubbleTea`.

## CHANGES

- Reorganize files
- Add in additional tests for TUI functionality

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -8,6 +8,8 @@
 *.so
 *.dylib
 
+*.golden -text
+
 # Test binary, built with `go test -c`
 *.test
 
--- a/cmd/templates/setup/loginModel.go
+++ b/cmd/templates/setup/loginModel.go
@@ -22,6 +22,7 @@ import (
 	"github.com/charmbracelet/log"
 	"github.com/datarobot/cli/cmd/auth"
 	"github.com/datarobot/cli/internal/assets"
+	"github.com/datarobot/cli/internal/base_auth"
 	"github.com/spf13/viper"
 )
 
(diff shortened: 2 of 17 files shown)


.gitignore
```diff
@@ -8,6 +8,8 @@
 *.so
 *.dylib
 
+*.golden -text
```
c-h-russell-walker, line 11: Random question as I review - is this correct with the space in here? (before `-text`)

internal/base_auth/auth.go
```diff
@@ -0,0 +1,73 @@
+package base_auth
```
c-h-russell-walker, line 1: Random thing I'm curious about:

I wondered if it's common practice in Go world to maybe use another format for packages as opposed to snake_case. I would have _guessed_ camelCase

Buuut at least judging from two other repos we have it's (wild to me) just all lowercase?
https://github.com/datarobot/oauth-providers-service/blob/main/internal/httperror/error.go#L13
https://github.com/datarobot/nbx-git-helper/blob/main/internal/gitconfig/githelper.go#L13

Curious if there's actually a preference to do all lowercase like that 🤔 

Huh - likve tweeting my thoughts - Go says:
> Good package names are short and clear. They are lower case, with no under_scores or mixedCaps.

https://go.dev/blog/package-names

cmd/templates/setup/loginModel.go
```diff
@@ -98,30 +122,112 @@ func waitForAPIKey(apiKeyChan chan string, server *http.Server, successCmd tea.C
...
+					if err := lm.server.Shutdown(context.Background()); err != nil {
+						log.Errorf("error during shutdown: %v", err)
+					}
+				}
+				return lm, tea.Quit
+			}
+		case "1":
+			// User pressed '1' to select app.datarobot.com
+			return lm, func() tea.Msg {
+				datarobotUrl := "https://app.datarobot.com"
```
c-h-russell-walker, line 172: nit: I feel like these strings (for three cloud envs) should be constants

alevan10: yeah, that's not a bad idea

yuriyhrytsyuk (approved):

PR #58, opened 2025-09-18
[CFX-3597] Add test for multiple selection

yuriyhrytsyuk:
# RATIONALE


alevan10 (approved):
those tests are much cleaner, thanks for the update!

PR #446, opened 2026-04-21
[CFX-5377] Add error handling for newLLMListPrompt

taras-pokornyy:
# RATIONALE
As a user running the dr dotenv setup wizard to configure a template that uses the DataRobot LLM Gateway, I want to see a clear error message when the CLI is unable to retrieve the list of available LLMs, so that I can quickly understand and resolve the root cause (e.g., auth misconfiguration, missing credentials, or a network issue) instead of staring at an empty, unresponsive wizard screen.

## CHANGES
Added error handling for newLLMListPrompt:

401 Unauthorized
- Once GetLLMs returns err 401 Unauthorized 👆 
<hr>

404 Not Found
- Once GetLLMs returns err  404 Not Found 👆 
<hr>

Client Timeout
- Once GetLLMs returns err  Request timeout (Client.Timeout) 👆 
<hr>

No active LLMs found in the catalog
- Once GetLLMs returns empty list llms.LLMs 👆 
<hr>

Misc Errors
- In case of other errors
<hr> 

Also added tests for changes that were made

## PR Automation

> [!IMPORTANT]
>

---

> [!NOTE]
> **Medium Risk**
> Changes error propagation from `drapi.Get` to return a typed `HTTPError` and updates the dotenv wizard to branch UI behavior on HTTP/network failures, which could affect any callers that relied on previous error strings or handling.

**No risk**
`drapi.Get` is used only in one place 😈 
> 
> **Overview**
> Improves the `dr dotenv` setup wizard’s LLM Gateway catalog prompt to **fail gracefully**: `newLLMListPrompt` now converts LLM catalog fetch failures (401/403, 404, timeouts, and other errors) and empty catalogs into an *error prompt* with user-facing guidance, while logging the underlying error.
> 
> Adds `drapi.HTTPError` from `drapi.Get` for non-200 responses so callers can inspect status codes via `errors.As`, and updates `promptModel.View()` to render error prompts with `tui.ErrorStyle` and suppress the text input when in an error state. Includes new tests covering the new error cases and view rendering behavior.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit e521c71a475ddc528d13489f5cdf984789bbb933. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/cmd/dotenv/promptModel.go
+++ b/cmd/dotenv/promptModel.go
@@ -49,6 +49,7 @@ var (
 const (
 	cursorStyle = '•'
 
+	typeError			 = "error"
 	errMsgPrefix         = "😞 Unable to retrieve the list of available LLMs.\n\n"
 	errMsgAuthFailed     = "🔐 Authentication failed. Please check your credentials and try again."
 	errMsgNotFound       = "🔍 Requested resource not found. Please contact support for assistance."
@@ -197,7 +198,7 @@ func newLLMListPrompt(prompt envbuilder.UserPrompt, successCmd tea.Cmd) (promptM
 		var (
 			netErr     net.Error
 			httpErr    *drapi.HTTPError
-			StatusCode int
+			statusCode int
 		)
 
 		helpMsg := errMsgPrefix
@@ -205,12 +206,12 @@ func newLLMListPrompt(prompt envbuilder.UserPrompt, successCmd tea.Cmd) (promptM
 		// Check if the error is a network timeout or an HTTP error to provide more specific feedback
 		// Treat net.Error.Timeout() as an HTTP 408 Request Timeout for user-friendly messaging
 		if errors.As(err, &netErr) && netErr.Timeout() {
-			StatusCode = http.StatusRequestTimeout
+			statusCode = http.StatusRequestTimeout
 		} else if errors.As(err, &httpErr) {
-			StatusCode = httpErr.StatusCode
+			statusCode = httpErr.StatusCode
 		}
 
-		switch StatusCode {
+		switch statusCode {
 		case http.StatusUnauthorized, http.StatusForbidden:
 			helpMsg += errMsgAuthFailed
 		case http.StatusNotFound:
@@ -222,7 +223,7 @@ func newLLMListPrompt(prompt envbuilder.UserPrompt, successCmd tea.Cmd) (promptM
 		}
 
 		errPrompt := prompt
-		errPrompt.Type = "error"
+		errPrompt.Type = typeError
 
 		errPrompt.Help = helpMsg
 
@@ -233,7 +234,7 @@ func newLLMListPrompt(prompt envbuilder.UserPrompt, successCmd tea.Cmd) (promptM
 		log.Warn("No active LLMs found in the catalog")
 
 		errPrompt := prompt
-		errPrompt.Type = "error"
+		errPrompt.Type = typeError
 		errPrompt.Help = errMsgNoLLMs
 
 		return promptModel{prompt: errPrompt, successCmd: successCmd}, nil
@@ -377,7 +378,7 @@ func (pm promptModel) View() string {
 		if pm.prompt.Multiple {
 			sb.WriteString(tui.DimStyle.Render("space to toggle • enter to answer • "))
 		}
-	} else if pm.prompt.Type.String() == "error" {
+	} else if pm.prompt.Type.String() == typeError {
 		sb.WriteString("\n\n")
 	} else {
 		sb.WriteString(pm.input.View())
--- a/cmd/dotenv/promptModel.go
+++ b/cmd/dotenv/promptModel.go
@@ -15,8 +15,11 @@
 package dotenv
 
 import (
+	"errors"
 	"fmt"
 	"io"
+	"net"
+	"net/http"
 	"slices"
 	"strings"
 
@@ -45,6 +48,13 @@ var (
 
 const (
 	cursorStyle = '•'
+
+	errMsgPrefix         = "😞 Unable to retrieve the list of available LLMs.\n\n"
+	errMsgAuthFailed     = "🔐 Authentication failed. Please check your credentials and try again."
+	errMsgNotFound       = "🔍 Requested resource not found. Please contact support for assistance."
+	errMsgTimeout        = "⏳ Request timed out. Please check your network connection and try again."
+	errMsgNoLLMs         = "🤷 No available LLMs found. Please contact support for assistance."
+	errMsgContactSupport = "👥 Please try again or contact support if the issue persists."
 )
 
 type item envbuilder.PromptOption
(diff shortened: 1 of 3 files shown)


cmd/dotenv/promptModel.go
```diff
@@ -181,7 +182,33 @@ func llmsToPromptOptions(llms []drapi.LLM) []envbuilder.PromptOption {
...
+		errPrompt.Type = "llmgw_error"
+		helpMsg := "😞 Unable to retrieve the list of available LLMs.\n\n"
+
+		if strings.Contains(err.Error(), "Unauthorized") {
+			helpMsg += "🔐 Authentication failed. Please check your credentials and try again."
+		} else if strings.Contains(err.Error(), "Not Found") {
+			helpMsg += "🔍 Requested resource not found. Please contact support for assistance."
+		} else if strings.Contains(err.Error(), "Timeout") {
+			helpMsg += "⏳ Request timed out. Please check your network connection and try again."
+		}
```
c-h-russell-walker: First, this PR looks great so far - and I really appreciate the screenshots in the description!

<hr>

This block of code stands out to me since we're using "magic strings" and checking strings in strings - which feels a little brittle to me.

Is it possible (even if you need to slightly alter `GetLLMs()` and/or `GetJSON()` so you can check against integers here that represent the status codes? `401`, `404` etc. ?

Also I _think_ that a more "go-thonic" (kinda like pythonic!) approach here could be using a match/case statement instead of if/else if blocks - what do you think?

cmd/dotenv/promptModel.go
```diff
@@ -181,7 +182,35 @@ func llmsToPromptOptions(llms []drapi.LLM) []envbuilder.PromptOption {
 func newLLMListPrompt(prompt envbuilder.UserPrompt, successCmd tea.Cmd) (promptModel, tea.Cmd) {
 	llms, err := drapi.GetLLMs()
 	if err != nil {
-		return promptModel{}, nil
+		log.Errorf("Error retrieving LLMs: %s", err.Error())
+
+		errPrompt := prompt
+		errPrompt.Type = "llmgw_error"
+		helpMsg := "😞 Unable to retrieve the list of available LLMs.\n\n"
```
c-h-russell-walker: Looks like this _is_ the default message but will never be used because of the additional `else` I see you just committed - am I right or maybe I need more coffee?

I _do_ think though that a match/case is a better case here if we can match against status code integers returned as I mentioned - so maybe with a refactor to match/case this can be taken care of - might even be clearer?

cmd/dotenv/promptModel.go
taras-pokornyy: Good idea! Thanks

I would like to avoid magic strings, but not sure how to do that in this particular case...
Should those strings be constants or just vars inside the func?

c-h-russell-walker: From my quick glance at the underlying HTTP related functions I feel like maybe you'll want to instead of returning either success or an error you return a tuple in those cases and return the integer status code as second item - then you'll have that where/when you need it to do a strict match against integers - make sense?

cmd/dotenv/promptModel.go
taras-pokornyy: Yeah 😅 
Additional `else` was just added due to CursorBot comment while you were writing a comment .

Going to try to use `switch/case`. The only concern is that I haven't ever received status code on request timeout ...

c-h-russell-walker: Ah true yeah - well when considering returning status codes in those http-related let's try to figure that out somehow ... 🤔

cmd/dotenv/promptModel.go
```diff
@@ -181,7 +192,51 @@ func llmsToPromptOptions(llms []drapi.LLM) []envbuilder.PromptOption {
 func newLLMListPrompt(prompt envbuilder.UserPrompt, successCmd tea.Cmd) (promptModel, tea.Cmd) {
 	llms, err := drapi.GetLLMs()
 	if err != nil {
-		return promptModel{}, nil
+		log.Errorf("Error retrieving LLMs: %s", err.Error())
+
+		var (
+			netErr     net.Error
+			httpErr    *drapi.HTTPError
+			StatusCode int
```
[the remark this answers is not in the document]
ajalon1: Yeah, move to statusCode

cmd/dotenv/promptModel.go
```diff
@@ -181,7 +192,51 @@ func llmsToPromptOptions(llms []drapi.LLM) []envbuilder.PromptOption {
 func newLLMListPrompt(prompt envbuilder.UserPrompt, successCmd tea.Cmd) (promptModel, tea.Cmd) {
 	llms, err := drapi.GetLLMs()
 	if err != nil {
-		return promptModel{}, nil
+		log.Errorf("Error retrieving LLMs: %s", err.Error())
```
ajalon1, line 196: this will be helpful in the log files

cmd/dotenv/promptModel_test.go
```diff
@@ -15,13 +15,50 @@
...
+	}))
+
+	viper.Set(config.DataRobotURL, srv.URL+"/api/v2")
+	viper.Set(config.DataRobotAPIKey, "test-token")
+
+	t.Cleanup(func() {
+		srv.Close()
+		viper.Reset()
+	})
+}
```
ajalon1, line 60: Very nice.

cmd/dotenv/promptModel_test.go
ajalon1, line 1: We use a lot of parametrized unit tests in this repo, but I actually rather like that these are not parametrized. It makes the tests themselves easier to read.

cmd/dotenv/promptModel.go
```diff
@@ -181,7 +192,51 @@ func llmsToPromptOptions(llms []drapi.LLM) []envbuilder.PromptOption {
...
+		errPrompt.Help = helpMsg
+
+		return promptModel{prompt: errPrompt, successCmd: successCmd}, nil
+	}
+
+	if len(llms.LLMs) == 0 {
+		log.Warn("No active LLMs found in the catalog")
+
+		errPrompt := prompt
+		errPrompt.Type = "error"
```
c-h-russell-walker: Can we make this use of `"error"` magic string maybe a `const` ?

cmd/dotenv/promptModel.go
```diff
@@ -181,7 +192,51 @@ func llmsToPromptOptions(llms []drapi.LLM) []envbuilder.PromptOption {
...
+		switch StatusCode {
+		case http.StatusUnauthorized, http.StatusForbidden:
+			helpMsg += errMsgAuthFailed
+		case http.StatusNotFound:
+			helpMsg += errMsgNotFound
+		case http.StatusRequestTimeout, http.StatusGatewayTimeout:
+			helpMsg += errMsgTimeout
+		default:
+			helpMsg += err.Error() + "\n\n" + errMsgContactSupport
+		}
```
ajalon1: Good use of switch/case.

ajalon1 (approved):
There's definitely room for making the error handler generic enough to reuse with any drapi call, and with any bubbletea renderer.

But I don't think we need to engineer this further, yet.

internal/drapi/get.go
```diff
@@ -17,14 +17,24 @@ package drapi
...
 
+// HTTPError is returned by Get when the server responds with a non-200 status code.
+// Callers can extract the status code with errors.As to make decisions without string matching.
+type HTTPError struct {
+	StatusCode int
+}
+
+func (e *HTTPError) Error() string {
+	return fmt.Sprintf("HTTP error: %d %s", e.StatusCode, http.StatusText(e.StatusCode))
+}
```
c-h-russell-walker, line 38: I may be missing something but is this function used anywhere?

cmd/dotenv/promptModel.go
```diff
@@ -316,6 +377,8 @@ func (pm promptModel) View() string {
 		if pm.prompt.Multiple {
 			sb.WriteString(tui.DimStyle.Render("space to toggle • enter to answer • "))
 		}
+	} else if pm.prompt.Type.String() == "error" {
```
c-h-russell-walker: This I think would benefit from what I mentioned before using a `const` instead of a magic string

cmd/dotenv/promptModel.go
```diff
@@ -181,7 +192,51 @@ func llmsToPromptOptions(llms []drapi.LLM) []envbuilder.PromptOption {
...
+		// Check if the error is a network timeout or an HTTP error to provide more specific feedback
+		// Treat net.Error.Timeout() as an HTTP 408 Request Timeout for user-friendly messaging
+		if errors.As(err, &netErr) && netErr.Timeout() {
+			StatusCode = http.StatusRequestTimeout
+		} else if errors.As(err, &httpErr) {
+			StatusCode = httpErr.StatusCode
+		}
+
+		switch StatusCode {
+		case http.StatusUnauthorized, http.StatusForbidden:
```
c-h-russell-walker, line 215: I think this whole update you made is awesome!

Love too that in match/case you can group multiple HTTP status codes together for same message - likely gives us good coverage on error messaging.

cmd/dotenv/promptModel.go
taras-pokornyy: Have added URL to err string also  🤓

c-h-russell-walker (approved):
Really great updates 👍

internal/drapi/get.go
taras-pokornyy: Yes
It is used in https://github.com/datarobot-oss/cli/pull/446/changes#diff-edf6278f536965611068548c18281a1a473572d82ab4d0e400a6b8635c821587R196

On another hand we have to implement Error() once it is custom Error

internal/drapi/get.go
```diff
@@ -67,7 +79,8 @@ func Get(url, info string) (*http.Response, error) {
 
 	if resp.StatusCode != http.StatusOK {
 		resp.Body.Close()
-		return nil, errors.New("Response status code is " + resp.Status + ".")
+
+		return nil, &HTTPError{StatusCode: resp.StatusCode}
```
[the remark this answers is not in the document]
taras-pokornyy: Lost that change at some point

taras-pokornyy:
@ajalon1 , @c-h-russell-walker , last comments before merge...

c-h-russell-walker (approved):
This PR is great - thanks for all there iterative updates!

PR #453, opened 2026-04-23
[RAPTOR-16920] workload artifact create command

wojtekwdr:
# RATIONALE

Continues the RAPTOR-16918/-16919 series (workload artifact `get` and `list`) by adding the write path: `dr workload artifact create`. Users can POST a JSON spec file to the Workload API and see the created artifact back in either human-readable or machine-parseable form, without hand-crafting a cURL against `/api/v2/artifacts/`.

## CHANGES

- New `dr workload artifact create --spec-file <path> [--output json]` subcommand under the `workload` feature gate.
- Strict client-side spec validation (`DisallowUnknownFields` + required-field checks for `name`, `spec.containerGroups`, and ≥1 container per group) so malformed specs fail fast before hitting the server. Raw spec bytes are forwarded verbatim to preserve fields the typed struct doesn't yet model.
- New `drapi.Post` / `drapi.PostJSON` helpers mirroring the existing `Get` / `GetJSON` shape (200/201 success, `HTTPError` on non-success, memoized token, standard headers, request-body restoration after `RedactedReqInfo`).
- `internal/workload` additions: `ArtifactCreateRequest` types, `ValidateCreateRequest`, `CreateArtifact`.
- Unit tests for spec validation, file reading, output formatting, and `Post` / `PostJSON` success + error paths.

## NOTES
- Built on top of #451 (list command). Rebase on `main` once #451 merges.

## PR Automation

**Comment-Commands:** Trigger CI by commenting on the PR:

- `/trigger-smoke-test` or `/trigger-test-smoke` — Run smoke tests
- `/trigger-install-test` or `/trigger-test-install` — Run installation tests

**Labels:** Apply labels to trigger workflows:

- `run-smoke-tests` or `go` — Run smoke tests on demand (only works for non-forked PRs)

---

> [!NOTE]
> **Medium Risk**
> Adds a new write-path API call and new HTTP POST helper; failures could affect artifact creation or request logging/body handling, though changes are scoped and well-tested.
> 
> **Overview**
> Adds a new `dr workload artifact create --spec-file <file>` command that reads a JSON spec, performs strict client-side validation (unknown-field rejection + required invariants), POSTs to `/api/v2/artifacts/`, and renders the created artifact in `text` or `json`.
> 
[...]

PR #456, opened 2026-04-28
[RAPTOR-16921][RAPTOR-16923] added dr workload code init command, added .wapi state management logic

wojtekwdr:
# RATIONALE

`dr workload code` is the **code sync layer** for the DataRobot code-to-workload (c2w) loop: developers edit code locally in their IDE, the CLI keeps the working directory in agreement with the remote artifact's code. This PR ships the first user-visible "code" leaf, `dr workload code init`, together with the local state-management library that all future `code` subcommands (`sync`, `versions`, `checkout`) will consume.

`init` is the entry point: it links a project directory to an existing artifact (created via `dr workload artifact create` or in the UI) by writing a machine-managed `.wapi/` state directory at the project root, conceptually `.git/` for c2w. Subsequent commands operate on this state to detect drift, apply changes, and version the code.

Stacked on top of #453 so it can reuse the typed `--output-format` flag, the artifact-status constants, and shared formatting helpers. Will rebase to `main` once #453 merges.

### State-management library (`.wapi/`)

A new private package owns the on-disk state that c2w commands read and write:

- A `.wapi/` directory at the project root, holding `config.json` (artifact identity + last-synced pointer), `manifest.json` (the BASE snapshot used by the future three-way diff), `history.log` (append-only JSONL operation log with size-based rotation), and `.gitignore` (so the `.wapi/` tree never lands in git).
- A `.wapiignore` template dropped at the project root on first init (matches `.gitignore`-style patterns; users own the file thereafter).
- Atomic file writes throughout (temp-file + rename) so partial failures never leave a half-written config or manifest.
- Nullable fields (`catalogId`, `lastSyncedVersionId`, `syncedAt`, `syncedVersionId`) round-trip as JSON `null` rather than empty strings, since the sync engine relies on these distinctions.
- Idempotent project-directory creation: if `--dir <path>` points at a non-existent directory, init creates it (and any missing parents) the way `git init <newdir>` does.
- Sentinels for the three expected error modes: already-linked, not-initialized, and corrupted state file.

### `dr workload code init`

```
dr workload code init [<artifact-id>] [flags]
  --dir <path>             # project directory (default: '.', auto-created if missing)
  --yes / -y               # non-interactive (also via DATAROBOT_CLI_NON_INTERACTIVE=true)
  --sync                   # not yet implemented; rejected before any FS write
  --output-format text|json
```

User-facing behaviour:

- **TTY-aware prompts** for the project directory and artifact ID when omitted; suppressed by `--yes` or when stdin is not a terminal.
- **Branching on the artifact's state:**
  - Artifact already has committed code: init records its catalog ID; the first `sync` will three-way diff against the remote manifest.
  - Artifact is empty (no committed code yet): init writes `catalogId: null`; the first `sync` will create the catalog and upload the local files as the initial version.
- **Friendly error mapping:** 404 becomes "Artifact `<id>` not found."; locked artifacts are refused with an explicit immutability message.
- **Already-linked re-runs** are detected before any HTTP call and produce a message that names the *existing* artifact (the one in `config.json`), not the one the user just typed, and never overwrite the existing config.
- **JSON output mode** emits a parseable result object (artifact metadata plus the linked directory) suitable for piping into `jq` in CI.

## TESTING

- **Unit:** lint clean; full package test suite passes under `-race`. Pure helpers are table-driven; the cobra layer is exercised via arg-validation tests; the end-to-end RunE is driven against a stubbed HTTP boundary on a temporary directory.
- **Manual smoke against `staging.datarobot.com`:** a 24-test suite covering help text, input validation, success paths (existing-code + empty artifact), guards (locked / 404 / malformed-id), already-linked re-run, `--dir` variants (including non-existent paths now auto-created), non-TTY behaviour, the `DATAROBOT_CLI_NON_INTERACTIVE` env var, and feature-gate on/off. **24/24 pass.**

## NOTES

- **Behind feature gate `DATAROBOT_CLI_FEATURE_WORKLOAD=true`** (inherited from the parent `workload` command).
- **Blocked by #453's merge.** Will rebase to `main` once #453 lands.
- **`--sync` flag is reserved.** Registered so it appears in `--help`, but rejected with an explicit error before any filesystem write. Avoids silent partial behaviour for automation that might assume init+sync is atomic. Real implementation arrives with the sync engine.
- **Out of scope:** sync engine, three-way diff, conflict copies, OCC, version listing, checkout. All future work.

## PR Automation

> [!IMPORTANT]
>

---

> [!NOTE]
> **Medium Risk**
> Adds new file-system state creation and persistence logic (atomic writes + log rotation) plus a new user-facing CLI entrypoint; failures or path-handling bugs could affect local project directories, but no auth or backend mutation logic is changed.
> 
> **Overview**
> Adds a new `dr workload code` command group (wired into `dr workload`) with an initial `init` subcommand to link a local directory to an existing workload artifact and persist local sync state.
> 
> Introduces a new internal `.wapi` state-management package that creates and maintains `.wapi/` (atomic `config.json` + `manifest.json` writes, append-only `history.log` with 1MB rotation, and `.wapi/.gitignore`) and drops a default `.wapiignore` template at the project root. The `init` flow is TTY/non-interactive aware, supports `--dir`, `--yes` and `--output-format json`, maps 404s to a friendly error, and refuses locked artifacts via a new `Artifact.IsLocked()` helper.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 0763591b2000995da85eeb68758df6bb235a716c. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


damjandr (approved):
Overall it looks okay, but it would be nice if someone from the CLI/Golang team could take a look

cmd/workload/code/init/cmd.go
```diff
@@ -0,0 +1,163 @@
...
+	workload.AddOutputFlag(c, &outputFormat)
+
+	return c
+}
+
+func runInit(cmd *cobra.Command, args []string, outputFormat workload.OutputFormat) error {
+	yes, _ := cmd.Flags().GetBool("yes")
+
+	if !yes {
+		if v, ok := os.LookupEnv("DATAROBOT_CLI_NON_INTERACTIVE"); ok {
```
victorborshak, line 80: We used `viperx.BindPFlag("yes", cmd.Flags().Lookup("yes"))` and `viperx.BindEnv("yes", "DATAROBOT_CLI_NON_INTERACTIVE")` in the other places. IMO might be good here too for consistency.

For example:
https://github.com/datarobot-oss/cli/blob/cd93a7f170670172723e55b5d9f9fb0fe873675c/cmd/dependencies/install/cmd.go#L42
https://github.com/datarobot-oss/cli/blob/main/cmd/dotenv/cmd.go#L231

victorborshak (approved):
Looks good to me. Left a minor commend about using `viperx`.

cmd/workload/code/init/cmd.go
wojtekwdr: will be fixed in follow-up PR, thanks for feedback!

PR #458, opened 2026-04-29
fix: only persistent flags in viper

ajalon1:
# RATIONALE

@victorborshak raised the alarm on the fact that Viper writes out all configuration properties to `drconfig.yaml` anytime `viper.WriteConfig()` is invoked. This is the documented behavior for the function, but not what we actually want. We are seeing things like `--yes` get written into drconfig.yaml.

When it comes down to it, we only want a few flags to be persisted.

## CHANGES

- in the root cmd, remove `BindPFlags()`. This is what was binding every single flag to viper. I *think* we added this in just to make it easy to configure everything, but I haven't seen proper use cases for this yet. We already explicitly bind PersistentFlags (as defined in `RootCmd.PersistentFlags`), which are the ones we really care about anyway.

- initializeConfig() now only wires up DATAROBOT_CLI_ env vars, as well as explicit, legacy env vars, as well as the config file. We do NOT call viper.BindPFlags anymore.

- factory.ai - generated dev docs

## PR Automation

> [!IMPORTANT]
>

### DIFF
--- a/cmd/dotenv/cmd.go
+++ b/cmd/dotenv/cmd.go
@@ -146,7 +146,12 @@ This wizard will help you:
 		variables, contents := envbuilder.VariablesFromLines(dotenvFileLines)
 
 		showAllPrompts, _ := cmd.Flags().GetBool("all")
-		yes := viper.GetBool("yes")
+		// Read --yes directly from flags rather than viper to avoid the
+		// flag value leaking into viper.AllSettings() (and from there into
+		// drconfig.yaml on writes). The DATAROBOT_CLI_NON_INTERACTIVE
+		// environment variable still flows through viper via BindEnv below.
+		yesFlag, _ := cmd.Flags().GetBool("yes")
+		yes := yesFlag || viper.GetBool("yes")
 
 		needsPulumi, pulumiLoggedIn, needsPassphrase := CheckPulumiSetup(repositoryRoot, variables)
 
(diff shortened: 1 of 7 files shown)


cmd/dotenv/cmd.go
```diff
@@ -195,8 +200,10 @@ func init() {
 	SetupCmd.Flags().BoolP("yes", "y", false, "Skip interactive prompts and use defaults (useful for automation).")
 	SetupCmd.MarkFlagsMutuallyExclusive("yes", "all")
 
-	// Bind flag to viper to enable env var support (DATAROBOT_CLI_NON_INTERACTIVE)
-	_ = viper.BindPFlag("yes", SetupCmd.Flags().Lookup("yes"))
+	// Bind only the env var (DATAROBOT_CLI_NON_INTERACTIVE) to viper.
+	// The --yes flag itself is read directly from cmd.Flags() in RunE so
+	// that an explicit --yes does not leak into viper.AllSettings() and
+	// get persisted to drconfig.yaml on subsequent config writes.
```
ajalon1, line 206: We can also remove the binding to DATAROBOT_CLI_NON_INTERACTIVE, if we really only want `--yes` the CLI arg to do this.

cmd/dotenv/cmd.go
```diff
@@ -146,7 +146,12 @@ This wizard will help you:
 		variables, contents := envbuilder.VariablesFromLines(dotenvFileLines)
 
 		showAllPrompts, _ := cmd.Flags().GetBool("all")
-		yes := viper.GetBool("yes")
+		// Read --yes directly from flags rather than viper to avoid the
+		// flag value leaking into viper.AllSettings() (and from there into
+		// drconfig.yaml on writes). The DATAROBOT_CLI_NON_INTERACTIVE
+		// environment variable still flows through viper via BindEnv below.
+		yesFlag, _ := cmd.Flags().GetBool("yes")
+		yes := yesFlag || viper.GetBool("yes")
```
ajalon1, line 154: This OR statement smells to me, so if we don't need the env var, we should remove it and just rely directly on cmd.Flags().GetBool

PR #459, opened 2026-04-29
feat: when viper writes to drconfig, allowlist certain fields

ajalon1:
# RATIONALE

We only want viper to write certain configuration properties to drconfig.yaml when `WriteConfig()` and `SafeWriteConfig()` are invoked. `--yes` is not one of them.

I think the ideal way to handle this would be to use separate viper instances to separate reading and writing of drconfig.yaml, but that is a larger refactor that needs more discussion and thought. This seems like a simpler way of handling things for now, with a bit of a hack.

## CHANGES

1. Create a new `internal/config/viperx` module that wraps a viper instance.
2. This module exports a subset of the viper interface that is currently, actually used.
3. swap all imports of `github.com/spf13/viper` with `internal/config/viperx`.
4. Adds [depguard](https://github.com/OpenPeeDeeP/depguard) as a dependency.
5. Creates `depguard` rule that denies use of `github.com/spf13/viper` package outside of `internal/config/**`.

## TESTING

The new depguard linter rule looks like this in practice:

```
╰─ ❌201 ❯ task lint
🧹 Checking go.mod…
🧹 Checking formatting…
🧹 Vetting…
🧹 Linting…
cmd/root.go:44:2: import 'github.com/spf13/viper' is not allowed from list 'no-raw-viper': Import internal/config/viperx instead of github.com/spf13/viper. The wrapper omits WriteConfig, SafeWriteConfig, and BindPFlags by design so transient flag state cannot leak into drconfig.yaml. See docs/development/configuration.md. (depguard)
        "github.com/spf13/viper"
        ^
cmd/root.go:146:18: Error return value of `viper.BindPFlags` is not checked (errcheck)
        viper.BindPFlags(RootCmd.PersistentFlags())
                        ^
2 issues:
* depguard: 1
* errcheck: 1
task: Failed to run task "lint": exit status 1
```

## PR Automation

> [!IMPORTANT]
>

### DIFF
--- a/.golangci.yaml
+++ b/.golangci.yaml
@@ -10,6 +10,7 @@ linters:
     - dupword
     - errname
     - exhaustive
+    - forbidigo
     - loggercheck
     - misspell
     - nestif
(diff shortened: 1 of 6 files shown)


ajalon1:
Not necessary now that #460 is merged

PR #461, opened 2026-04-30
[CFX-5919] Adding --output param to 'dr dotenv setup'

victorborshak:
# RATIONALE

Adding --output param to 'dr dotenv setup'. This allows to write the `.env` file to a custom location.

Usage: `dr dotenv setup --output <path_to_dir>`

## PR Automation

> [!IMPORTANT]
>

---

> [!NOTE]
> **Low Risk**
> Scoped CLI enhancement that changes only the destination path and ensures directories exist before writing; minimal impact beyond file I/O behavior.
> 
> **Overview**
> `dr dotenv setup` now supports `-o, --output <directory>` to write the generated `.env` file outside the repository root, while keeping the default behavior unchanged.
> 
> The `.env` writer now ensures the target directory exists before creating the file, and new tests cover default output, custom output, and auto-creation of nested output directories. Documentation is updated with the new flag and an example.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 07aa58f8f84f8d080995debfebb33cadeed703cf. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/cmd/dotenv/cmd.go
+++ b/cmd/dotenv/cmd.go
@@ -109,12 +109,20 @@ This wizard will help you:
 	PreRunE: auth.EnsureAuthenticatedE,
 
 	RunE: func(cmd *cobra.Command, _ []string) error {
+		outputDir, _ := cmd.Flags().GetString("output")
+
 		repositoryRoot, err := ensureInRepo()
 		if err != nil {
 			return err
 		}
 
-		dotenvFile := filepath.Join(repositoryRoot, ".env")
+		var dotenvFile string
+
+		if outputDir != "" {
+			dotenvFile = filepath.Join(outputDir, ".env")
+		} else {
+			dotenvFile = filepath.Join(repositoryRoot, ".env")
+		}
 
 		// Check if we should skip when .env exists and all required variables are set
 		flagIfNeededSet, _ := cmd.Flags().GetBool("if-needed")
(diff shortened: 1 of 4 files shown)


cmd/dotenv/cmd.go
```diff
@@ -109,12 +109,20 @@ This wizard will help you:
...
 			return err
 		}
 
-		dotenvFile := filepath.Join(repositoryRoot, ".env")
+		var dotenvFile string
+
+		if outputDir != "" {
+			dotenvFile = filepath.Join(outputDir, ".env")
+		} else {
+			dotenvFile = filepath.Join(repositoryRoot, ".env")
```
c-h-russell-walker, line 124: Any chance you think it's a good idea for UX to quietly accept a string where the user has _also_ included `.env` (maybe strip off `.env` if it exists)

So like instead of: `-o /my/dir/`
I send `-o /my/dir/.env`

?

c-h-russell-walker (approved):

cmd/dotenv/cmd.go
victorborshak: Probably not. It should not be writing to any custom filenames, which users can provide this way - only `.env`. Also, I don't think we want to block the possibility to save into a directory named `.env`. So from this point of view, having the argument to be a directory path seems to be the right choise.

c-h-russell-walker: Sounds good!

PR #504, opened 2026-05-14
[CFX-5956] Fix and refactor telemetry flush on os.exit...

cdevent:
# RATIONALE

Telemetry events were silently dropped whenever a command exited via `os.Exit(1)`.
Cobra's `PersistentPostRunE` hook (where `client.Flush` was called) is never invoked
when a code path calls `os.Exit` directly — not even deferred functions run. The same
bypass caused `log.Stop()` to be skipped, leaving log resources uncleaned. Additionally,
commands that printed their own ❌ … error message and then called `os.Exit(1)` could
not propagate a proper exit code through cobra, so the pattern had spread through the
codebase as a workaround.

### Immediate fix — telemetry flush on all exit paths

**exit.go (new file)**
- Added `cmd.Exit(code int)` — flushes the `telemetryClient` package-level variable
  (3 s timeout) then calls `os.Exit`. Nil-safe: if `PersistentPreRunE` never ran (e.g.
  flag parse failure) there are no queued events and it falls straight through to `os.Exit`.

**telemetry.go**
- Added exported `ClientContextKey{}` — the context key used to store and retrieve the
  active `*Client` from a cobra command's context; defined here so all sub-packages can
  access it without importing cmd (which would be circular).
- Added `ExitWithContext(ctx context.Context, code int)` — looks up the client stored
  under `ClientContextKey{}`, flushes it, then calls `os.Exit`. Used by cobra
  sub-commands that must propagate a subprocess exit code.
- **Removed** `globalClient`, `SetGlobal`, and `Exit` — telemetry is now
  **stateless**; it holds no global variables.

**root.go**
- Replaced `telemetry.SetGlobal(client)` with `telemetryClient = client` (package-level
  var shared with exit.go).
- Changed context key from the unexported `telemetryClientKey{}` to the exported
  `telemetry.ClientContextKey{}`, so sub-packages can retrieve the client without
  importing cmd.

**main.go**
- Error path now calls `cmd.Exit(1)` instead of `telemetry.Exit(1)`; removed direct
  telemetry import.
- Added `defer log.Stop()` so log cleanup is unconditional (not only on the happy path).

### Long-term fix — eliminate `os.Exit` from cobra command internals

**command.go**
- `ErrSilent` sentinel: commands that already print their own user-facing error return
  `cli.ErrSilent` with `SilenceErrors: true` on the cobra command, preventing cobra from
  double-printing `Error: …`.

**discovery.go**
- Replaced `ExitWithError(err)` (which called `os.Exit`) with `DiscoveryError(err) error`
  that returns a formatted error value.

**All cmd commands converted from `Run` → `RunE`**, removing all internal `os.Exit` /
`telemetry.Exit` calls: `check`, `login`, `logout`, `start`, `add`, `update`, `dotenv`
(`UpdateCmd`, `ValidateCmd`), `list`, `compose`, `run`.

**`SilenceErrors: true` added** to `component add` and `component update` commands so
that `cli.ErrSilent` returned by their `RunE` is not echoed by cobra.

### Two intentional `telemetry.ExitWithContext` call-sites remain

These are legitimate exceptions where an exact integer exit code from a child process
must be forwarded to the OS and cannot be encoded in a Go error:

- discovery.go — plugin subprocess exit-code passthrough
- cmd.go — `--exit-code` flag passthrough

### Documentation (telemetry.md)
- Added "Process exit and telemetry flush" section documenting `cmd.Exit`,
  `telemetry.ExitWithContext`, state ownership table, and `SilenceErrors` / `ErrSilent`
  rules with examples.

## Result

Telemetry is flushed on all exit paths (success, error return, and the two subprocess
passthroughs). No command inside cobra calls `os.Exit` speculatively; errors propagate
via `return err` so `PersistentPostRunE` and `defer` statements run normally.

`task build` ✅ · `task lint` 0 issues ✅ · `task test` ✅

## PR Automation

> [!IMPORTANT]
>

---

> [!NOTE]
> **Medium Risk**
> Touches process-exit and error-handling paths across many commands; behavior changes could affect exit codes and stderr output, though changes are mostly mechanical and centralized.
> 
> **Overview**
> Ensures telemetry and logging are cleaned up even when commands terminate early: adds a global telemetry client plus `telemetry.Exit(code)` (flush then `os.Exit`), registers it in `cmd/root.go`, and switches `main.go` to exit via `telemetry.Exit(1)` with unconditional `log.Stop()`.
> 
> Standardizes command error handling by introducing `cli.ErrSilent` and updating multiple commands (auth, dotenv, component, task, start) to use `RunE`, return errors instead of calling `os.Exit`, and set `SilenceErrors`/`SilenceUsage` to prevent double-printing. Task discovery error handling is refactored from `ExitWithError` to `DiscoveryError(err)`.
> 
> Keeps intentional hard-exit behavior only where exact subprocess exit codes must be propagated (plugin execution and `dr run --exit-code`).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit d6649ffb4c2b92dd606db99eb456122b12ffe835. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/cmd/auth/check/cmd.go
+++ b/cmd/auth/check/cmd.go
@@ -26,6 +26,7 @@ import (
 	"github.com/datarobot/cli/internal/config"
 	"github.com/datarobot/cli/internal/envbuilder"
 	"github.com/datarobot/cli/internal/repo"
+	"github.com/datarobot/cli/internal/telemetry"
 	"github.com/datarobot/cli/tui"
 	"github.com/spf13/cobra"
 )
@@ -225,7 +226,7 @@ func Run(_ *cobra.Command, _ []string) {
 			return
 		}
 
-		os.Exit(1)
+		telemetry.Exit(1)
 	}
 
 	if checkDotenvCredentials(repoRoot) {
@@ -236,7 +237,7 @@ func Run(_ *cobra.Command, _ []string) {
 		return
 	}
 
-	os.Exit(1)
+	telemetry.Exit(1)
 }
 
 func Cmd() *cobra.Command {
--- a/cmd/auth/check/cmd.go
+++ b/cmd/auth/check/cmd.go
@@ -23,10 +23,10 @@ import (
 	"strings"
 
 	"github.com/datarobot/cli/internal/auth"
+	"github.com/datarobot/cli/internal/cli"
 	"github.com/datarobot/cli/internal/config"
 	"github.com/datarobot/cli/internal/envbuilder"
 	"github.com/datarobot/cli/internal/repo"
-	"github.com/datarobot/cli/internal/telemetry"
 	"github.com/datarobot/cli/tui"
 	"github.com/spf13/cobra"
 )
@@ -217,36 +217,38 @@ func checkDotenvCredentials(repoRoot string) bool {
 	return true
 }
 
-func Run(_ *cobra.Command, _ []string) {
+func RunE(_ *cobra.Command, _ []string) error {
 	// Check .env credentials if in a repo
 	// If not, check the CLI credentials only
 	repoRoot, err := repo.FindRepoRoot()
 	if err != nil {
 		if checkCLICredentials() {
-			return
+			return nil
 		}
 
-		telemetry.Exit(1)
+		return cli.ErrSilent
 	}
 
 	if checkDotenvCredentials(repoRoot) {
-		return
+		return nil
 	}
 
 	if checkCLICredentials() {
-		return
+		return nil
 	}
 
-	telemetry.Exit(1)
+	return cli.ErrSilent
 }
 
 func Cmd() *cobra.Command {
 	return &cobra.Command{
-		Use:   "check",
-		Short: "✅ Check if DataRobot credentials are valid",
+		Use:           "check",
+		Short:         "✅ Check if DataRobot credentials are valid",
+		SilenceErrors: true,
+		SilenceUsage:  true,
 		Long: `Verify that your DataRobot credentials are properly configured and valid.
 
 If you're in a project directory with a '.env' file, this will check those credentials.`,
-		Run: Run,
+		RunE: RunE,
 	}
 }
--- a/cmd/component/add/cmd.go
+++ b/cmd/component/add/cmd.go
@@ -144,10 +144,11 @@ func Cmd() *cobra.Command {
 	names := strings.Join(copier.EnabledShortNames, ", ")
 
 	cmd := &cobra.Command{
-		Use:     fmt.Sprintf("add [%s or component_url]", names),
-		Short:   "➕ Add a component",
-		PreRunE: PreRunE,
-		RunE:    RunE,
+		Use:           fmt.Sprintf("add [%s or component_url]", names),
+		Short:         "➕ Add a component",
+		PreRunE:       PreRunE,
+		RunE:          RunE,
+		SilenceErrors: true,
 	}
 
 	cmd.Flags().StringArrayVarP(&addFlags.DataArgs, "data", "d", []string{}, "Provide answer data in key=value format (can be specified multiple times)")
--- a/cmd/component/add/cmd.go
+++ b/cmd/component/add/cmd.go
@@ -64,7 +64,9 @@ func RunE(_ *cobra.Command, args []string) error {
 		return err
 	}
 
-	compose.Cmd().Run(nil, nil)
+	if err := compose.Cmd().RunE(nil, nil); err != nil {
+		return err
+	}
 
 	// Validate and edit .env if needed
 	if err := dotenv.ValidateAndEditIfNeeded(); err != nil {
--- a/cmd/component/add/cmd.go
+++ b/cmd/component/add/cmd.go
@@ -17,7 +17,6 @@ package add
 import (
 	"errors"
 	"fmt"
-	"os"
 	"strings"
 
 	tea "github.com/charmbracelet/bubbletea"
@@ -57,7 +56,7 @@ func RunE(_ *cobra.Command, args []string) error {
 	cliData, err := shared.ParseDataArgs(addFlags.DataArgs)
 	if err != nil {
 		log.Error(err)
-		os.Exit(1)
+		telemetry.Exit(1)
 
 		return nil
 	}
@@ -134,7 +133,7 @@ func addComponents(repoURLs []string, componentConfig *config.ComponentDefaults,
 		err := copier.ExecAdd(repoURL, mergedData, addFlags)
 		if err != nil {
 			log.Error(err)
-			os.Exit(1)
+			telemetry.Exit(1)
 
 			return nil
 		}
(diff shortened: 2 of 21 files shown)


cdevent:
@cursor review

cdevent:
@cursor review

cmd/component/add/cmd.go
```diff
@@ -56,10 +55,7 @@ func RunE(_ *cobra.Command, args []string) error {
 
 	cliData, err := shared.ParseDataArgs(addFlags.DataArgs)
 	if err != nil {
-		log.Error(err)
-		os.Exit(1)
-
-		return nil
```
taras-pokornyy, line 62: Just wondering, how should that be working previously?

taras-pokornyy (approved):
LGTM!

cmd/auth/check/cmd.go
```diff
@@ -216,36 +217,38 @@ func checkDotenvCredentials(repoRoot string) bool {
 	return true
 }
 
-func Run(_ *cobra.Command, _ []string) {
+func RunE(_ *cobra.Command, _ []string) error {
```
ajalon1, line 220: Switching over to RunE makes more sense across the board.

cmd/auth/check/cmd.go
```diff
@@ -216,36 +217,38 @@ func checkDotenvCredentials(repoRoot string) bool {
...
 	// If not, check the CLI credentials only
 	repoRoot, err := repo.FindRepoRoot()
 	if err != nil {
 		if checkCLICredentials() {
-			return
+			return nil
 		}
 
-		os.Exit(1)
+		return cli.ErrSilent
```
ajalon1, line 229: Works for me.

internal/cli/command.go
```diff
@@ -15,10 +15,19 @@
...
 
 import (
+	"errors"
+
 	"github.com/datarobot/cli/internal/features"
 	"github.com/spf13/cobra"
 )
 
+// ErrSilent is returned by RunE implementations that have already printed their
+// own user-facing error message. Pair it with SilenceErrors: true on the
```
ajalon1, line 25: Docs here are helpful. I think it would be good to add a line somewhere in the telemetry dev docs.

internal/log/logger.go
```diff
@@ -108,7 +108,10 @@ func StartFile() {
 func StopFile() {
 	fileLogger = nil
 
-	fileWriter.Close()
+	if fileWriter != nil {
+		fileWriter.Close()
+		fileWriter = nil
+	}
```
ajalon1, line 114: Huh, I wonder why this wasn't caught before! Let's sneak it in.

internal/task/discovery.go
```diff
@@ -148,26 +148,23 @@ func (d *Discovery) Discover(root string, maxDepth int) (string, error) {
 	return rootTaskfilePath, nil
 }
 
-func ExitWithError(err error) {
+// DiscoveryError formats a discovery error into a user-friendly message string.
+// Commands should call this to get the message, print it themselves if needed,
+// and return cli.ErrSilent (or the returned error directly).
+func DiscoveryError(err error) error {
```
ajalon1: Maybe a better name? `FormatDiscoveryError`?

internal/task/discovery.go
```diff
@@ -148,26 +148,23 @@ func (d *Discovery) Discover(root string, maxDepth int) (string, error) {
...
-		fmt.Fprintln(os.Stderr, tui.BaseTextStyle.Render(err.Error()))
-		os.Exit(1)
-
-		return
+		return fmt.Errorf("%s\n%s",
+			tui.ErrorStyle.Render("Error: Cannot generate 'Taskfile' because an existing 'Taskfile' already has a dotenv directive."),
+			tui.BaseTextStyle.Render(err.Error()))
 	}
 
-	_, _ = fmt.Fprintln(os.Stderr, "Error discovering tasks: ", err)
```
ajalon1, line 168: +1

internal/telemetry/telemetry.go
```diff
@@ -160,6 +161,28 @@ func (c *Client) Flush(timeout time.Duration) {
...
+// delivered even when cobra's PersistentPostRunE hook is bypassed (e.g. when
+// RunE returns an error, or when a command calls os.Exit directly).
+func Exit(code int) {
+	if globalClient != nil {
+		globalClient.Flush(3 * time.Second)
+	}
+
+	os.Exit(code)
+}
+
```
ajalon1: I'm OK with there being a global client for telemetry, but I don't like that it is scoped to `internal/telemetry/telemetry` which previously was completely stateless.

Can we move it somewhere else?

cmd/component/update/cmd.go
ajalon1, line 1: Some comments on this refactor could be useful. Cyclomatic complexity?

ajalon1 (commented):
Looks pretty good overall, but I really don't like the way we use a global telemetry client now.

> No command inside cobra calls os.Exit speculatively; errors propagate via return err so PersistentPostRunE and defer statements run normally.

I'd suggest configuring golangci-lint to make sure that we don't allow further `os.Exit()`. forbidigo is already configured.

internal/telemetry/telemetry.go
ajalon1: Like the root command? Everything goes through there anyway.

ajalon1: Alternate claudy solution:


```suggestion
// globalClient holds the active telemetry client for process-wide Exit calls.
// It is set by SetGlobal during PersistentPreRunE before any command runs.
var globalClient *Client

// SetGlobal stores c as the global telemetry client. Must be called during
// PersistentPreRunE before any os.Exit-based code paths can be reached.
func SetGlobal(c *Client) {
	if c == nil {
	    // failsafe to make sure we don't screw up in dev or testing
		panic("telemetry.SetGlobal: client cannot be nil")
	}
	globalClient = c
}

// Exit flushes any pending telemetry events then calls os.Exit(code).
// Use this instead of os.Exit in command code so that Amplitude events are
// delivered even when cobra's PersistentPostRunE hook is bypassed (e.g. when
// RunE returns an error, or when a command calls os.Exit directly).
func Exit(code int) {
	if globalClient == nil {
	    // failssafe to make sure we don't screw up in dev or testing
		panic("telemetry.Exit called before SetGlobal - PersistentPreRunE must run first")
	}
	globalClient.Flush(3 * time.Second)
	
	os.Exit(code)
}

```

cdevent: that make sense for me too, i've tried to have like more condensed and store at it's own level.. like that's related to telemetry so might be there. but no problem move that on root level.

cdevent:
@ajalon1 need you review here again..

internal/log/logger.go
cdevent: who knows =)

internal/cli/command.go
cdevent: Added some docs in telemetry.md

cdevent:
@ajalon1 can we move forward with that? any objections, pros/cons?

ajalon1 (approved):

PR #520, opened 2026-05-21
[CFX-6255]Add support detection for CLI-supported shells and tests

cdevent:
---

# RATIONALE

When `dr` is installed via Homebrew, the cask post-install hook runs inside Homebrew's Ruby interpreter. `DetectShell()` called `parentProcessName()` first, which ran `ps -p <ppid> -o comm=` and got back `ruby`. Since `ruby` passed through `normalizeShellName()` unchanged and returned with no error, the `$SHELL` fallback was never reached — causing `dr self completion install` to fail immediately with:

```
Error: Unsupported shell: ruby.
```

install.sh was unaffected because it reads `$SHELL` directly.

## CHANGES

**shell.go**
- Added `isSupportedShell(name string) bool` — checks a normalized name against `SupportedShells()` (`bash`, `zsh`, `fish`, `powershell`)
- `DetectShell()` now only returns the parent process name if `isSupportedShell()` passes; non-shell parents (ruby, python, node, etc.) fall through to the `$SHELL` fallback
- Updated comment on the `$SHELL` branch to clarify it is a real environment variable that IS inherited through the full subprocess chain (brew → ruby → `dr`)

**shell_test.go**
- Added `TestIsSupportedShell` — verifies supported shells return true, and non-shells (`ruby`, `sh`, `node`, `""`) return false
- Added a tombstone comment explaining why `ZSH_VERSION`/`BASH_VERSION`/`FISH_VERSION` were considered and rejected (they are shell-only parameters, not environment variables, and are never inherited by subprocesses)

## PR Automation

> [!IMPORTANT]
>

---

> [!NOTE]
> **Medium Risk**
> Changes shell-detection behavior used for completion installation; mis-detection could affect install UX across environments (notably package-manager invocations), but scope is limited and covered by new tests.
> 
> **Overview**
> **Refines shell detection for completion install.** `DetectShell` now normalizes the parent process name and only accepts it if it’s one of the CLI’s supported shells, avoiding false positives when launched by non-shell parents (e.g. package managers).
> 
> Adds `isSupportedShell` plus tests to validate supported-shell filtering, and documents that shell-only version parameters (e.g. `ZSH_VERSION`) are intentionally not used since they aren’t inherited like `$SHELL`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit dc2e72fd0dae8ac673cffca670d83b43ba646b15. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/internal/shell/shell.go
+++ b/internal/shell/shell.go
@@ -81,23 +81,8 @@ func DetectShell() (string, error) {
 		}
 	}
 
-	// Shell-specific version variables are exported by the shell and inherited
-	// by every subprocess, including package-manager installers (brew → ruby).
-	// Check them before $SHELL because they reflect the active session shell
-	// rather than the login shell recorded in /etc/passwd.
-	if os.Getenv("ZSH_VERSION") != "" {
-		return string(Zsh), nil
-	}
-
-	if os.Getenv("BASH_VERSION") != "" {
-		return string(Bash), nil
-	}
-
-	if os.Getenv("FISH_VERSION") != "" {
-		return string(Fish), nil
-	}
-
-	// Try SHELL environment variable last (Unix/macOS login shell path).
+	// $SHELL is a real environment variable that is inherited by all
+	// subprocesses, including package-manager installers (brew → ruby → dr).
 	if shellPath := os.Getenv("SHELL"); shellPath != "" {
 		return normalizeShellName(filepath.Base(shellPath)), nil
 	}
--- a/internal/shell/shell.go
+++ b/internal/shell/shell.go
@@ -59,13 +59,45 @@ func normalizeShellName(name string) string {
 	return name
 }
 
-func DetectShell() (string, error) {
-	// Prefer the parent process name — accurate even after `exec sh` or similar.
-	if name := parentProcessName(); name != "" {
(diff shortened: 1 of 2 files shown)


cdevent:
@cursor review

internal/shell/shell.go
```diff
@@ -59,13 +59,30 @@ func normalizeShellName(name string) string {
...
-	// Prefer the parent process name — accurate even after `exec sh` or similar.
+	// Prefer the parent process name — accurate for normal interactive use.
+	// Only trust it when it resolves to a shell the CLI actually supports;
+	// non-shell parents (e.g. Homebrew's Ruby interpreter) must fall through.
 	if name := parentProcessName(); name != "" {
-		return normalizeShellName(name), nil
+		if normalized := normalizeShellName(name); isSupportedShell(normalized) {
+			return normalized, nil
+		}
 	}
```
ajalon1, line 84: I'd add a log.debug at this point in the code, when we fallback to $SHELL, here just so we can track this in the future.

ajalon1 (approved):
LGTM, but one log.debug would be great

PR #527, opened 2026-05-27
[CFX-6024][CFX-6028] Refactor workload config manifest validation

brunoromano-dr:
# RATIONALE

Workload `.wapi/` state (`config.json`, `manifest.json`) was only checked for JSON syntax on load. Tampered or hand-edited files could carry invalid IDs, unsafe file paths, inconsistent sync metadata, or malformed hashes—leading to confusing sync/checkout failures later (including after network/API work) instead of a clear local corruption error.

This PR adds struct-tag validation via `go-playground/validator/v10`, centralized in `internal/workload/wapi/validate.go`, and runs it at **load time** for on-disk config/manifest and at the **start of `Initialize`** for caller-supplied options. Validation is intentionally **not** applied on `SaveConfig`/`SaveManifest` so the sync engine (trusted writer) cannot fail Phase 6 after a successful remote upload.

Security-focused rules reject path separators and `..` in IDs (used under `.wapi/.checkouts/<versionId>/`) and reuse `fileops.SafeRelPath` for manifest file keys.

### Dependency

- Add direct dependency `github.com/go-playground/validator/v10` (transitive: `universal-translator`, `locales`, `go-urn`, `mimetype`).

### New: `internal/workload/wapi/validate.go`

| Function | Purpose |
|----------|---------|
| `getValidator()` | Lazy singleton; registers custom validators |
| `validateDRID` | Custom tag `dr_id`: non-empty, max 256 chars, no `/`, `\`, or `..` |
| `validateDRNonemptyPtr` | Custom tag `dr_nonempty_ptr`: non-nil `*string` must not point to `""` |
| `validateConfig` | Struct tags + cross-field: `lastSyncedVersionId` requires `catalogId` |
| `validateManifest` | Struct tags + `validateManifestSyncState` + `SafeRelPath` on `files` keys |
| `validateManifestSyncState` | `syncedAt` and `syncedVersionId` must both be set or both absent |
| `validateInitOptions` | Struct tags + same catalog/version cross-field as config |
| `formatStructValidation` / `formatFieldError` | User-facing messages using JSON field names |

### Validator tags

| Tag | Meaning |
|-----|---------|
| **`required`** | The field must be present and non-zero (e.g. non-empty string, non-zero `time.Time`). |
| **`omitempty`** | Skip remaining tag checks when the field is empty (nil pointer, empty string, zero number). |
| **`dr_id`** | Custom: value must be a non-empty identifier (max 256 chars) with no `/`, `\`, or `..` (safe for filesystem paths). |
| **`dr_nonempty_ptr`** | Custom: if a `*string` is non-nil, the pointed-to string must not be `""` (rejects JSON `""` where `null` was intended). |
| **`eq=1`** | Integer field must equal `1` (only supported manifest schema version). |
| **`gte=0`** | Numeric field must be greater than or equal to zero. |
| **`len=64`** | String must be exactly 64 characters long. |
| **`hexadecimal`** | String must contain only lowercase hex digits (`0-9`, `a-f`). |
| **`dive`** | Validate each value inside a map/slice (used so every `files` entry runs `FileMeta` rules). |

### Struct tags and wiring

| Type | File | `validate` tags (summary) | Called from |
|------|------|----------------------------|-------------|
| **`Config`** | `config.go` | `artifactId` required + `dr_id`; optional `*catalogId` / `*lastSyncedVersionId` → `dr_nonempty_ptr`, `dr_id`; `createdAt`, `cliVersion` required | **`LoadConfig`** after `json.Unmarshal` → `*CorruptedError` on failure |
| **`Manifest`** | `manifest.go` | `version` eq=1; `syncedVersionId` optional + `dr_nonempty_ptr` + `dr_id`; `files` `dive` | **`LoadManifest`** after unmarshal + nil `files` normalization → `*CorruptedError` |
| **`FileMeta`** | `manifest.go` | `hash` required, lowercase hex, len=64; `size` gte=0 | Validated via `Manifest.files` `dive` on load |
| **`InitOptions`** | `init.go` | `artifactId` required + `dr_id`; optional `catalogId` / `lastSyncedVersionId` + `dr_id` | **`Initialize`** first line (before any `mkdir`) → plain `error` |

**Not validated on save:** `SaveConfig`, `SaveManifest`, `writeConfig`, `writeManifest` (sync engine / init remain trusted writers).

**`CorruptedError`** (`errors.go`): doc updated to include semantic validation failures.

#### `Config` / `InitOptions` IDs

- **Required:** `artifactId`, `createdAt`, `cliVersion` (config only).
- **`dr_id`:** Rejects empty, `/`, `\`, `..`, length > 256.
- **`dr_nonempty_ptr`:** Rejects `"catalogId": ""` / `"lastSyncedVersionId": ""` as non-nil empty strings (config).
- **Cross-field:** If `lastSyncedVersionId` is set, `catalogId` must be set (config pointers; init strings).

#### `Manifest`

- **`version`:** Must be `1`.
- **`syncedVersionId`:** Same `dr_id` / `dr_nonempty_ptr` rules when present.
- **Sync state:** `syncedAt` ↔ `syncedVersionId` must appear together (fresh init: both null is valid).
- **`files` keys:** Each path checked with `fileops.SafeRelPath` (no `..`, absolute paths, backslashes).
- **`FileMeta`:** SHA-256 hex exactly 64 chars; `size >= 0`.

#### `validate_test.go` (unit tests on validators)

| Test | What it checks |
|------|----------------|
| `TestValidateConfig_ValidPointers` | Full config + JSON round-trip passes |
| `TestValidateDRNonemptyPtr_OnPointer` | `dr_nonempty_ptr` on `*string` in isolation |
| `TestIsValidDRID` | `isValidDRID` accept/reject cases |
| `TestValidateInitOptions` / `valid` | Artifact-only; +catalog; +catalog+version |
| `TestValidateInitOptions` / `errors` | Missing artifact; version without catalog; `dr_id` on each field |
| `TestValidateConfig` / `valid nil optionals` | Nil catalog/version pointers OK |
| `TestValidateConfig` / `errors` | Missing/empty fields, zero `createdAt`, `dr_id`, version without catalog |
| `TestValidateManifest` / `valid empty base` | Version 1, empty `files` |
| `TestValidateManifest` / `errors` | Wrong version, sync mismatch, path escape, hash length/hex, negative size |

#### Integration tests (file I/O + `CorruptedError`)

| File | New / updated cases |
|------|---------------------|
| `config_test.go` | `TestConfig_LoadInvalid`: missing/empty `artifactId`, empty optional strings, zero `createdAt`, missing `cliVersion`, version without catalog |
| `manifest_test.go` | `TestManifest_LoadInvalid`: version 2, `../` path, short/non-hex hash, negative size, sync pointer mismatch; round-trip hashes updated to 64-char hex |
| `init_test.go` | `TestInitialize_InvalidOptions`: empty artifact, version without catalog — asserts **no `.wapi/`** created |
| `helpers_test.go` | `testHash()` helper for 64-char manifest fixtures |

> [!IMPORTANT]
>


ajalon1:
Good choice for validation frameworks.

go.mod
```diff
@@ -14,6 +14,7 @@ require (
 	github.com/codeclysm/extract/v4 v4.0.0
 	github.com/denisbrodbeck/machineid v1.0.1
 	github.com/gitsight/go-vcsurl v1.0.1
+	github.com/go-playground/validator/v10 v10.30.2
```
ajalon1, line 17: Well, since you're the first one in -- you get CFX-6024 as well.

internal/workload/wapi/manifest.go
```diff
@@ -23,11 +23,10 @@ import (
 )
 
 // FileMeta is the per-file entry in the BASE manifest. Hash is SHA-256 hex
-// (64 chars) as produced by the sync engine; this package does not validate
-// its shape.
+// (64 chars) as produced by the sync engine.
 type FileMeta struct {
-	Hash string `json:"hash"`
-	Size int64  `json:"size"`
+	Hash string `json:"hash" validate:"required,hexadecimal,len=64"`
```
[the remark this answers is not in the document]
brunoromano-dr: Solved in commit b8b6d3f400d557ac7c667db0b3cd598c37c793e4

internal/workload/wapi/init.go
```diff
@@ -46,6 +46,10 @@ type InitOptions struct {
 // mkdir, the incomplete .wapi/ tree is left in place for the user to inspect
 // or remove manually rather than attempting rollback.
 func Initialize(projectDir string, opts InitOptions) error {
+	if err := validateInitOptions(opts); err != nil {
+		return err
+	}
+
 	if err := os.MkdirAll(projectDir, 0o755); err != nil {
```
ajalon1, line 56: Minor nit:  you could update the docstring to mention that basic validation happens before we get to mkdir. Or you couldn't, I don't feel strongly either way.

internal/workload/wapi/validate.go
ajalon1, line 1: I would strongly suggest documenting each of these custom validators for future devs. These also look like good candidates for a common validate module, so if you want to add a TODO and a JIRA issue for that, I'd appreciate it. But at minimum, docs.

internal/workload/wapi/manifest_test.go
```diff
@@ -91,10 +92,10 @@ func TestManifest_PreservesPathsVerbatim(t *testing.T) {
...
 
 	paths := map[string]FileMeta{
-		"forward/slash.py":   {Hash: "aaa", Size: 1},
-		"deep/nested/x.py":   {Hash: "bbb", Size: 2},
-		"unicode/café.py":    {Hash: "ccc", Size: 3},
-		"with spaces/f.json": {Hash: "ddd", Size: 4},
+		"forward/slash.py":   {Hash: testHash('a'), Size: 1},
+		"deep/nested/x.py":   {Hash: testHash('b'), Size: 2},
+		"unicode/café.py":    {Hash: testHash('c'), Size: 3},
+		"with spaces/f.json": {Hash: testHash('d'), Size: 4},
```
ajalon1, line 98: I like this -- intent of the fixture is very clear now.

internal/workload/wapi/validate.go
```diff
@@ -0,0 +1,253 @@
...
+			return fmt.Errorf("files[%q]: %w", path, err)
+		}
+	}
+
+	return nil
+}
+
+func validateManifestSyncState(m Manifest) error {
+	hasVersion := m.SyncedVersionID != nil && *m.SyncedVersionID != ""
+	hasSyncedAt := m.SyncedAt != nil
```
ajalon1: Claude tells me that there is an uncovered edge case here, when `m.SyncedAt.IsZero()`, or "0001-01-01T00:00:00Z").

Seems like a reasonable check. This could be another custom validator, or I think the following will suffice. But doesn't seem urgent to me. :D


```suggestion
func validateManifestSyncState(m Manifest) error {
	hasVersion := m.SyncedVersionID != nil && *m.SyncedVersionID != ""
	hasSyncedAt := m.SyncedAt != nil && !m.SyncedAt.IsZero()
```

ajalon1 (approved):
LGTM, thanks @brunoromano-dr !

go.mod
brunoromano-dr: Sounds good to me

internal/workload/wapi/init.go
brunoromano-dr: Solved in commit c8172e72c654f7d0d75531380752c7a7a0c4b049

go.mod
brunoromano-dr: Finalized with commit f3a176e57b642af7c65dab13432a713e08a7f102

internal/workload/wapi/validate.go
brunoromano-dr: We have now (commit ba00ad6347d9b001726966fe54efa8f3d44247c5) `validator.go` module with simple validation methods, decoupled from validating Config, InitOptions and Manifest in `validate.go`. You can find validator docs in `validator.go`. Let me know if I need to create a separate .md for validator doc

ajalon1: works for me

PR #538, opened 2026-05-29
Revert "Revert "Correctly use local template for tasks if available during `dr run`""

carsongee:
Reverts datarobot-oss/cli#537

## Changes since original merge

* Added preference to use on-disk Taskfile over Taskfile.gen.yaml
* Added fixture to test it
* Added integration test to validate it will not loop using fixture
* Added environment injection to escape loop

---

> [!NOTE]
> **Low Risk**
> Refactors taskfile generation/discovery only; no auth, data, or deployment path changes beyond matching local template behavior for `dr run`.
> 
> **Overview**
> Task discovery is **consolidated** into `task.NewDiscovery` in `internal/task`, replacing duplicate logic in the compose command (`createDiscovery` / `validateTemplatePath`).
> 
> **`dr run`** now uses that shared path instead of only `NewTaskDiscovery`, so generated `Taskfile.gen.yaml` can use a **project-root `.Taskfile.template`** when no `--template` is passed. Auto-detection of that file moved from compose-time setup into **`Discover()`**, keyed off the project root.
> 
> Compose still passes an explicit template via the flag when set; empty template defers to runtime detection during discovery.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit ebfbb1a33c3604f77518a08f602966035db64075. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


victorborshak (approved):

victorborshak:
@carsongee This looks good to me but please note that smoke tests are failing.

ajalon1 (commented):
Follow-up review of this merged change (retroactive pass). Two more issues outside this diff's line ranges, noted here for visibility since GitHub won't anchor inline comments to untouched code:

1. **`cmd/task/list/cmd.go` (~line 294)** — still calls the old `task.NewTaskDiscovery("Taskfile.gen.yaml")` instead of the new discovery/`PreferRootTaskfile` path added here for `dr run`. In a template project with a `.Taskfile.template` (or a committed root Taskfile), `dr run <task>` now uses the project-specific Taskfile, but `dr task list` regenerates `Taskfile.gen.yaml` from the embedded default template — overwriting the file `dr run` just produced and showing a different task list than what will actually execute next.

2. **`cmd/task/run/cmd.go` (~line 240, `completeTaskNames`)** — shell completion also still uses the old `task.NewTaskDiscovery("Taskfile.gen.yaml")` path rather than the new discovery/`PreferRootTaskfile` logic added in this PR. A recipe with a custom `.Taskfile.template` defines extra tasks that tab-completion for `dr run <TAB>` won't offer (or offers tasks that don't exist in the file that actually runs), and it overwrites `Taskfile.gen.yaml` with the embedded-default version as a side effect of completion.

Filing the recursion bug below (inline comment) as a separate CFX ticket for follow-up.

cmd/task/run/cmd.go
```diff
@@ -69,6 +71,32 @@ func splitTaskArgs(args []string) (taskNames []string, taskArgs []string) {
 	return taskNames, taskArgs
 }
 
+func taskfileForRun(dir string) (string, bool, error) {
```
ajalon1, line 74: **Correctness — infinite recursion.** This recursion guard only activates when a pre-existing root `Taskfile.yaml`/`.yml` is found (the `os.Getenv(taskRunFromRootEnv) != ""` branch is set by the *caller*, not here). When the Taskfile is freshly generated from an auto-detected `.Taskfile.template`, `taskRunFromRootEnv` is never set on this path, so a task that itself invokes `dr task run <task>` recurses forever.

Reproduced live: a template with `.datarobot/answers` and a `.Taskfile.template` (no committed Taskfile.yaml/yml) whose `lint` task runs `dr task run lint`. Running `dr task run lint` spawns an unbounded chain of `task`/`dr` child processes (>15 generations observed in 10s before a forced timeout), because every nested invocation regenerates the same `Taskfile.gen.yaml` from the same template and never sets `DATAROBOT_CLI_TASK_RUN_FROM_ROOT`. The sample fixture's `start` task avoids this only by accident (it calls `dr task compose` first, which materializes a real `Taskfile.yaml` so the *next* recursion hits the `PreferRootTaskfile` branch); `lint`/`install`/`test` in the same template have no such side effect and loop indefinitely.

Filing as a CFX ticket for follow-up since this already shipped.

internal/task/discovery.go
```diff
@@ -107,47 +109,101 @@ func NewTaskDiscovery(rootTaskfileName string) *Discovery {
...
 
-	// Check if any discovered Taskfiles already have a dotenv directive
-	if err := d.checkForDotenvConflicts(root, includes); err != nil {
-		return "", err
+	return NewComposeDiscovery(taskfileName, absPath), nil
+}
+
+// IsTemplateDir reports whether dir is a DataRobot template directory
+// (i.e. it contains a ".datarobot" folder).
+func IsTemplateDir(dir string) bool {
```
ajalon1, line 141: `task.IsTemplateDir` duplicates `repo.IsTemplateDir` with weaker, diverging semantics (any `.datarobot` dir vs. requiring `.datarobot/answers` or a non-trivial `.datarobot/cli`) — even though `repo.IsTemplateDir`'s doc comment says to use it "instead of ad-hoc os.Stat(\".datarobot\") checks." A directory with a bare/empty `.datarobot` folder (e.g. only `.datarobot/cli/state.yaml`) is treated as a template by `dr task run`/`dr task compose` but not by `dr dotenv`/`FindRepoRoot`, giving inconsistent answers depending on which command runs. Worth consolidating on `repo.IsTemplateDir`.

internal/task/runner.go
```diff
@@ -175,7 +176,11 @@ func (r *Runner) Run(tasks []string, opts RunOpts) error {
 
 	cmd.Stdout = r.opts.Stdout
 	cmd.Stderr = r.opts.Stderr
+
 	cmd.Stdin = r.opts.Stdin
```
ajalon1, line 180: Nit: `cmd.Stdin = r.opts.Stdin` is immediately followed by `if len(opts.Env) > 0 {` with no blank line between them — violates the project's whitespace convention ("Separate statement types" per CLAUDE.md, enforced by `wsl_v5`). `task lint` should flag this.

carsongee: Weird, this isn't even in the real code, and it wouldn't have passed linters 🤔

PR #568, opened 2026-06-12
[RAPTOR-17792] feat(workload): workload endpoint/logs commands

wojtekwdr:
# RATIONALE

Adds `dr workload endpoint` and `dr workload logs` so users can grab a workload's URL and read or tail its container logs from the CLI. Also drops `--wait` from `start`/`stop`/`status`: in-CLI blocking only earns its keep on the genuinely long-running build commands, and the lifecycle target-wait it needed kept generating edge cases (residual-state false-settles, settled states that can never reach the target). `dr workload status` is back to a single fetch; poll it yourself if you need to block.

## CHANGES

**`dr workload endpoint <id>`** (new) - prints the bare endpoint URL on stdout so it composes in scripts; errors if the workload has no endpoint.

**`dr workload logs <id>`** (new) - shows container logs newest-last like `kubectl logs --tail`, with `--limit`, `--level` (validated against the server's set), and `--output-format json`. `--follow`/`-f` tails new lines (Ctrl-C to stop) using a server-side time cursor, so bursts aren't dropped and an idle follow stays quiet; transient errors retry up to a cap.

**`start` / `stop` / `status`** - `--wait`/`--poll-interval`/`--poll-timeout` removed. `start`/`stop` print the ack and point at `dr workload status`; `status` is a single fetch (an `errored` status prints and exits zero). `--wait` stays on `build trigger`/`build get`, the only genuinely long-running commands.

## TESTING

`task lint` clean; `task test` (race) green; verified the flag surfaces on the built `./dist/dr` (`start`/`stop`/`status` carry only `--output-format`; `build trigger` keeps `--wait`; `logs` keeps `--follow`).


internal/workload/logs.go
```diff
@@ -0,0 +1,473 @@
...
+// fetch retrieves the next poll's entries: the newest-limit window when
+// seeding or in window mode, else everything newer than the cursor. hadSince
+// reports which mode was used.
+func (f *logFollower) fetch() (entries []WorkloadLogEntry, hadSince bool, err error) {
+	since := ""
+	maxEntries := f.limit
+
+	if f.seeded && f.cursorUsable && !f.cursor.IsZero() {
+		since = f.cursor.Add(-followLagAllowance).UTC().Format(time.RFC3339Nano)
+		maxEntries = 0
```
brunoromano-dr, line 326: Why do we want to default page_size (by setting maxEntries to 0) if these conditions are met? Or, opposite question, why do we want to limit page size to maxEntries if these conditions are not met?

internal/workload/logs.go
```diff
@@ -0,0 +1,473 @@
...
+
+	// Newest first from the server; chronological for display.
+	slices.Reverse(entries)
+
+	fresh := f.dedup.filterUnseen(entries)
+
+	// A full window with zero overlap means >limit lines arrived since the
+	// last poll and the excess is unfetchable.
+	if f.seeded && !hadSince && len(entries) == f.limit && len(fresh) == len(entries) {
+		f.onWarn(fmt.Sprintf("possible gap: more than %d new lines arrived since the last poll and some may have been skipped (re-run with a larger --limit)", f.limit))
```
brunoromano-dr, line 376: For the future: if we find a full window came by and we may have lost some, couldn't we refetch since the same point in time with multiple smaller windows?

internal/workload/logs.go
```diff
@@ -0,0 +1,473 @@
...
+		if _, ok := d.cur[key]; ok {
+			continue
+		}
+
+		if _, ok := d.prev[key]; ok {
+			continue
+		}
+
+		if len(d.cur) >= d.genCap {
+			d.prev = d.cur
```
brunoromano-dr, line 463: We are rotating prev and cur entries: how do we prevent having a stream of 2*d.genCap triggering this full rotation twice and erase all history?

internal/workload/logs.go
```diff
@@ -0,0 +1,473 @@
...
+			all = all[:maxEntries]
+
+			break
+		}
+
+		if resp.Next == "" {
+			break
+		}
+
+		if err := drapi.AssertNextOnSameHost(resp.Next); err != nil {
```
brunoromano-dr, line 205: Should we have a unit test to prevent malformed or off-host next links?

internal/workload/logs_test.go
```diff
@@ -0,0 +1,659 @@
...
+	require.Error(t, err)
+
+	var httpErr *drapi.HTTPError
+
+	require.ErrorAs(t, err, &httpErr)
+	assert.Equal(t, http.StatusNotFound, httpErr.StatusCode)
+	assert.Equal(t, 1, calls)
+}
+
+func TestFollowWorkloadLogs_GivesUpAfterSustainedTransientErrors(t *testing.T) {
```
brunoromano-dr, line 545: Do we have a test for transient error and subsequent recovery? Can we support this or actually not?

internal/workload/logs_test.go
```diff
@@ -0,0 +1,659 @@
...
+
+	srv := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		calls++
+
+		switch {
+		case calls == 1:
+			fmt.Fprint(w, logsPage("", logEntryDoc("INFO", "a")))
+		case r.URL.Query().Get("startTime") != "":
+			// An older server rejects the filter outright.
+			w.WriteHeader(http.StatusUnprocessableEntity)
```
brunoromano-dr: Found the following with Cursor:
"What’s worth adding is a small, explicit 400 case [...] the test today only uses 422 in the mock; it never checks that 400 is classified as filter rejection rather than a terminal 4xx. A typo in isFilterRejectedError (e.g. dropping 400) would fail silently until a real server returns 400."

brunoromano-dr (commented):
Had a few questions about some test cases or potential bug, but looks good to me. Great work as usual

internal/workload/logs.go
wojtekwdr: Deffered for now. Capping cursor mode at limit would silently drop the older end of the new window whenever more than limit lines arrived between polls, which is exactly what the cursor is meant to bridge.

internal/workload/logs.go
wojtekwdr: Logging as a backlog item, for now the warning points the user at --limit

internal/workload/logs.go
wojtekwdr: added test

internal/workload/logs_test.go
wojtekwdr: added tests

internal/workload/logs_test.go
wojtekwdr: added tests

internal/workload/logs.go
wojtekwdr: deferred, dedup is the backstop and the cursor’s 10s overlap is what actually catches repeats so the lost generation only matters when the cursor is broken AND >10k lines arrive in one poll, which we already warn about

brunoromano-dr (approved):
lgtm

damjandr (approved):

PR #577, opened 2026-06-16
[CFX-5789] [CFX-5790]  `ToolRegistry` & Surface environment-aware failure messaging

taras-pokornyy:
# RATIONALE
When `dr deps install` runs a tool's install command and it fails, the previous error was a bare exit-code line with no guidance on what to try next. This change wires up the existing `ToolRegistry` data to produce actionable, environment-aware tips when an install command fails — _"You have pyenv — try: pip install uv"_ — and falls back to the canonical install command (e.g. curl | sh) when no manager is detected. Permission-denied failures (exit code 126 or matching stderr text) get an OS-specific "re-run with sudo / as Administrator" message instead.

## PR Automation

> [!IMPORTANT]
>

---

> [!NOTE]
> **Low Risk**
> UX and messaging only around failed installs; success path unchanged. Suggestions are advisory shell commands, not auto-executed.
> 
> **Overview**
> When `dr deps install` fails, users now get structured output on the writer (not only a minimal error): what was tried, optional tips, retry command, and docs URL. **`InstallPrerequisites`** tees install output into a buffer so failures can be classified and messaged.
> 
> A new **`ToolRegistry`** maps tools (python, uv, node, pulumi, task, git) to ordered install strategies (pyenv/brew/asdf/winget/etc.) plus fallbacks. **`DetectEnvironment`** probes PATH and NVM layout; **`selectInstallStrategy`** picks the first matching manager, skipping the manager that just failed. **`buildInstallFailureMsg`** / **`buildInstallTip`** add OS-specific guidance for permission errors (exit 126 / stderr heuristics) vs alternate install commands (e.g. “You have pyenv — try: pip install uv”). Returned errors are shortened; detail lives in stdout.
> 
> Broad unit coverage was added for registry, strategy selection, and failure messaging.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 6c10ba742b0cd20c3237b22b2ce800697aa1b665. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/internal/dependencies/installer.go
+++ b/internal/dependencies/installer.go
@@ -139,16 +139,16 @@ func buildInstallFailureMsg(prerequisite tools.Prerequisite, exitCode int, permD
 		fmt.Fprintf(&sb, "✗ %s install failed (exit code %d)\n", toolName, exitCode)
 	}
 
-	fmt.Fprintf(&sb, "  Tried: %s\n", installCmd)
+	fmt.Fprintf(&sb, TAB+"Tried: %s\n", installCmd)
 
 	if tip := buildInstallTip(prerequisite, permDenied, env, goos); tip != "" {
 		fmt.Fprintf(&sb, "%s\n", tip)
 	}
 
-	fmt.Fprintf(&sb, "  Raw command if you want to retry: %s\n", installCmd)
+	fmt.Fprintf(&sb, TAB+"Raw command if you want to retry: %s\n", installCmd)
 
 	if prerequisite.URL != "" {
-		fmt.Fprintf(&sb, "  Refer to %s for manual installation instructions.\n", prerequisite.URL)
+		fmt.Fprintf(&sb, TAB+"Refer to %s for manual installation instructions.\n", prerequisite.URL)
 	}
 
 	return sb.String()
@@ -160,11 +160,11 @@ func buildInstallTip(prerequisite tools.Prerequisite, permDenied bool, env map[s
 	if permDenied {
 		switch goos {
 		case "windows":
-			return "  Tip: This action requires Administrator privileges. Please restart your terminal/tool as Administrator."
+			return TAB + "Tip: This action requires Administrator privileges. Please restart your terminal/tool as Administrator."
 		case "darwin", "linux":
-			return "  Tip: This action requires root privileges. Please re-run this tool using 'sudo'."
+			return TAB + "Tip: This action requires root privileges. Please re-run this tool using 'sudo'."
 		default:
-			return "  Tip: Administrative or root privileges are required to perform this action."
+			return TAB + "Tip: Administrative or root privileges are required to perform this action."
 		}
 	}
 
--- a/internal/dependencies/installer.go
+++ b/internal/dependencies/installer.go
@@ -15,10 +15,13 @@
 package dependencies
 
 import (
+	"bytes"
 	"errors"
 	"fmt"
 	"io"
 	"os/exec"
+	"runtime"
+	"strings"
 
 	"github.com/datarobot/cli/internal/log"
 	"github.com/datarobot/cli/internal/repo"
@@ -45,7 +48,9 @@ func InstallPrerequisites(w io.Writer, prerequisites []tools.Prerequisite) ([]st
 
 		fmt.Fprintf(w, "📦 Installing %s...\n", prerequisite.Name)
 
-		exitCode, err := ExecuteShLine(installCmd, w)
+		var cmdBuf bytes.Buffer
+
+		exitCode, err := ExecuteShLine(installCmd, io.MultiWriter(w, &cmdBuf))
 		if err != nil {
 			log.Debug("deps: install failed to start", "name", prerequisite.Name, "err", err)
 
@@ -55,7 +60,13 @@ func InstallPrerequisites(w io.Writer, prerequisites []tools.Prerequisite) ([]st
 		if exitCode != 0 {
 			log.Debug("deps: install exited non-zero", "name", prerequisite.Name, "exit_code", exitCode)
 
-			return installed, fmt.Errorf("install failed for %q (exit code %d)\n  Please run manually: %s\n  Or check %s", prerequisite.Name, exitCode, installCmd, prerequisite.URL)
+			env := DetectEnvironment()
+			permDenied := isPermissionDenied(exitCode, cmdBuf.String())
+			msg := buildInstallFailureMsg(prerequisite, exitCode, permDenied, env, runtime.GOOS)
+
+			fmt.Fprint(w, msg)
+
+			return installed, fmt.Errorf("install failed for %q (exit code %d)", prerequisite.Name, exitCode)
 		}
 
 		log.Debug("deps: tool installed", "name", prerequisite.Name)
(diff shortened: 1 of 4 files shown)


taras-pokornyy:
@cursor review

internal/dependencies/registry.go
```diff
@@ -0,0 +1,362 @@
...
+	"os/exec"
+	"path/filepath"
+	"runtime"
+	"strings"
+)
+
+// Strategy is implemented by ManagerStrategy and FallbackStrategy.
+// getStrategyTip returns the user-facing tip line for an install failure, or ""
+// when no actionable suggestion is available.
+// withVersion returns a copy of the strategy with {version} and {version_mm}
```
c-h-russell-walker, line 32: Sorry for the dumb question but what's the `_mm` suffix stand for?

c-h-russell-walker: Ah, just saw `majorMinorVersion()` and it clicked ✅

internal/dependencies/registry.go
```diff
@@ -0,0 +1,362 @@
...
+		tipMsg = "\n\t" + strings.Join(ms.Commands, "\n\t")
+	}
+
+	return fmt.Sprintf("  Tip: You have %s — try: %s", ms.Manager, tipMsg)
+}
+
+func (fs FallbackStrategy) getStrategyTip(goos string) string {
+	cmds := fs.Commands
+
+	if goos == "windows" && len(fs.CommandsWindows) > 0 {
```
c-h-russell-walker, line 122: nit here:

Do we have reusable values in this repo. so we're not using magic strings for things like `"windows"` and likely `"darwin"` and `"linux"` ?

internal/dependencies/registry.go
```diff
@@ -0,0 +1,362 @@
...
+	switch len(cmds) {
+	case 0:
+		if fs.URL != "" {
+			return "  See: " + fs.URL
+		}
+
+		return ""
+
+	case 1:
+		return "  Try: " + cmds[0]
```
c-h-russell-walker: So wasn't sure where to put this.

Curious what you and @ajalon1 and others think.

I like your usage of `\t` elsewhere and wonder if that would be better to use than the many places you're using double spaces??

```suggestion
		return "\tTry: " + cmds[0]
```

To me it's:
- less likely to accidentally have one or three spaces at some point as a UI bug
- easier to read in the code as opposed to empty spaces - maybe that's just me though

c-h-russell-walker (approved):
Looks great - made some nit comments. And thanks for the demo today during our meeting.

internal/dependencies/registry.go
taras-pokornyy: That stands for Major + Minor version without patch/build version
Example:
 - version : 3.9.6
 - version_mm : 3.9
 
 Reason not all managers handle patch/build version . For example `brew` has 
 python@3.10
 python@3.11
 python@3.12
 python@3.13
 python@3.14
 python@3.9 (deprecated)

internal/dependencies/registry.go
```diff
@@ -0,0 +1,362 @@
...
+	present := func(name string) bool {
+		_, err := lookPath(name)
+
+		return err == nil
+	}
+
+	nvmDir := getenv("NVM_DIR")
+
+	if nvmDir == "" {
+		home := getenv("HOME")
```
ajalon1: The correct way to do this would be `home, err := os.UserHomeDir()`

internal/dependencies/registry.go
ajalon1: "windows", "darwin", and "linux" are all potential values of `runtime.GOOS`, which is a build-time constant.

https://pkg.go.dev/internal/goos#GOOS
https://cs.opensource.google/go/go/+/refs/tags/go1.26.4:src/internal/goos/zgoos_linux.go;l=7

**I am not recommending this because I'm not convinced it's worth it here** but one way to approach this would be to conditionally compile the fallbackstrategy code using go build. (You can extend packages based on runtime.GOOS and runtime.GOARCH.) A working example of this can be found in `internal/workload` where Wojciech has implemented OS-specific `diskspace` implementations.
- https://github.com/datarobot-oss/cli/blob/main/internal/workload/sync/diskspace_unix.go
- https://github.com/datarobot-oss/cli/blob/main/internal/workload/sync/diskspace_windows.go

What MIGHT be worth doing is something like this, where we enumerate the OS we actually support.

```
     type OS string

     const (
         OSDarwin  OS = "darwin"
         OSLinux   OS = "linux"
         OSWindows OS = "windows"
     )

     func IsWindows(os OS) bool { return os == OSWindows }
     func IsUnixLike(os OS) bool { return os == OSDarwin || os == OSLinux }
```

internal/dependencies/installer.go
```diff
@@ -81,6 +92,102 @@ func InstallPrerequisites(w io.Writer, prerequisites []tools.Prerequisite) ([]st
...
+
+	installCmd, _ := prerequisite.PlatformInstallCommand()
+	failedMgr := extractFailedManager(installCmd)
+
+	strategy := selectInstallStrategy(toolKey, failedMgr, env)
+	if strategy == nil {
+		return ""
+	}
+
+	return strategy.withVersion(prerequisite.MinimumVersion).getStrategyTip(goos)
```
ajalon1, line 188: What happens if prerequisite.MinimumVersion == ""? Will we get a string like `"nvm install {version}"` back?

internal/dependencies/registry.go
ajalon1: This is all just informational though, since I know @taras-pokornyy you want a bigger discussion on how we can scale this registry.

internal/dependencies/installer.go
```diff
@@ -81,6 +92,102 @@ func InstallPrerequisites(w io.Writer, prerequisites []tools.Prerequisite) ([]st
 	return installed, nil
 }
 
+// isPermissionDenied inspects exit codes and stderr text across OS types.
+func isPermissionDenied(exitCode int, stderr string) bool {
```
ajalon1, line 96: Nice.

internal/dependencies/installer.go
```diff
@@ -81,6 +92,101 @@ func InstallPrerequisites(w io.Writer, prerequisites []tools.Prerequisite) ([]st
...
+// extractFailedManager heuristically identifies the package/version manager
+// referenced in cmd (e.g. "brew" in "brew install uv"). Returns "" if none found.
+func extractFailedManager(cmd string) string {
+	for _, m := range knownManagers {
+		if strings.Contains(cmd, m) {
+			return m
+		}
+	}
+
+	return ""
```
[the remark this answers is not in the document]
ajalon1: I don't know how much crazier we want to get here. I think this is fine.

internal/dependencies/registry.go
ajalon1: I think it'd be good to stick to a single convention, but that becomes a renderer / UI issue and I think we can do this better with a more structured formatting approach than just `\t`. `text/tabwriter` is a helpful package if we want to go that route.

ajalon1 (approved):
This LGTM as is. The registry could use some more work, but that is out of scope.

The one thing I would suggest addressing is what happens with `strategy.WithVersion("")`. You can either do that here or in a followup.

internal/dependencies/registry.go
taras-pokornyy: Good catch ! Thank you 🤝 
Fixed in [ff5cca2](https://github.com/datarobot-oss/cli/pull/577/commits/ff5cca28227f843aea69daedfc92001fe5a06872)

internal/dependencies/registry.go
taras-pokornyy: That's a good point. 
First level indention was two spaces. Next level indention was "\t" (It is much longer). 

Created a const TAB which is two spaces and used that for indentions. 
First level indention is just  TAB
Next level indention is TAB+TAB

Commit : [dc1017f](https://github.com/datarobot-oss/cli/pull/577/commits/dc1017f8f438a5e3bd471254ee1860349b0d3ea5)

internal/dependencies/installer.go
taras-pokornyy: That's very good point !  🎉 
Added DefaultVersion to the Strategies.,which is used once MinimumVersion is empty. 

Please note that not all managers accept version! 

Commit : [e1fec45](https://github.com/datarobot-oss/cli/pull/577/commits/e1fec45745b6cb979c38560fd658576435450baa)

internal/dependencies/registry.go
taras-pokornyy: I think it is a good idea to have constants for GOOS instead of magic strings but not in the scope of this PR 🤷‍♂️

taras-pokornyy:
Hey @c-h-russell-walker , @ajalon1 !
Thank you for reviews and comments. I really appreciate that 😀 
I believe I have fixed all the issues except GOOS magic strings.

Going to merge this guy tonight. 
Please let me know in case of anything 😉

internal/dependencies/installer_test.go
ajalon1, line 1: nice, good test

ajalon1 (approved):
Still LGTM pending smoke test results

c-h-russell-walker (approved):

PR #583, opened 2026-06-19
[CFX-6587] Phase 2 Workflows restructure — De-duplication & cost 

cdevent:
# RATIONALE

Follows Phase 0 (CodeQL/govulncheck baseline) and Phase 1 (security hardening) on the same branch. This phase removes redundant CI work and consolidates build paths — no functional behavior changes, no test coverage removed.

## CHANGES

**R1 — Collapse `lint` job to a single `task lint`** (`checks.yaml`)

`golangci/golangci-lint-action` and a standalone `goreleaser-action check` step were both running before `task lint`, which already calls `golangci-lint run` and `goreleaser check` internally. Removed the two duplicate steps; `task lint` is the single source of truth (matches local dev per `AGENTS.md`).

**R2 — Replace 3-OS PR build matrix with one goreleaser cross-compile** (`checks.yaml`)

The `build` job was spinning up native ubuntu + macOS + windows runners just to compile-check on every Go change. Replaced with a single `ubuntu-latest` job running `goreleaser build --snapshot --clean`, which cross-compiles all platforms in one runner at ~1/3 the cost. The `dr-ubuntu` artifact (consumed by `completion-tests`) is uploaded from `dist/linux_linux_amd64_v1/dr` — downstream job is unaffected. Real macOS/Windows execution is still covered by smoke tests.

**R3 — Single Trivy scan feeding both SARIF and table** (`security-scan.yaml`)

Was running two full filesystem scans back-to-back (one for the SARIF Security-tab upload, one for the human-readable table). Now: one scan to JSON via `trivy-action`, then `trivy convert --format sarif` for the upload and `trivy convert --format table` for the log display.

**R4 — One Windows build path** (`.build-matrix.yaml`)

Removed `windows-latest` from the default `os-matrix` (now `["ubuntu-latest", "macos-latest"]`). Windows binaries always go through `.build-windows.yaml` which cross-compiles via GoReleaser on ubuntu. Added a comment documenting this invariant.

**Doc** (`docs/development/ci-workflow-optimization-plan.md`): added Phase 2 deliverables section and ticked off R1–R4 in the quick-win checklist.

## NOTES

- `Build Binaries` check name changes from three per-OS matrix entries (`Build Binaries / (ubuntu-latest)` etc.) to a single `Build Binaries`. Not a required check today, so no branch-protection update needed.
- `.build-matrix.yaml` remains in place (no callers broken) with the narrowed default for any future use.

## PR Automation

> [!IMPORTANT]
>

---

> [!NOTE]
> **Low Risk**
> Workflow-only changes with no application logic; the main regression surface is a wrong GoReleaser `dist/` path or a broken `dr-ubuntu` artifact for completion tests.
> 
> **Overview**
> **CI cost and duplication** are reduced by making local `task` workflows the single path for lint/build checks and by scanning once where Trivy ran twice.
> 
> The **lint** job no longer runs `golangci-lint-action` and a separate GoReleaser `check` before `task lint` (those tools already run inside `task lint`). It adds a **golangci-lint binary/analysis cache** keyed off `Taskfile.yaml`’s version and Go sources.
> 
> **Build Binaries** stops calling `.build-matrix.yaml` on three native runners and instead uses one `ubuntu-latest` job with `goreleaser build --snapshot --clean`, a quick `./dist/linux_linux_amd64_v1/dr --help` smoke, and the same **`dr-ubuntu`** artifact path for `completion-tests`.
> 
> **Trivy** runs one filesystem scan to JSON, then `trivy convert` for SARIF upload and table logs instead of two full scans.
> 
> **.build-matrix.yaml**’s default `os-matrix` drops `windows-latest` (documented as GoReleaser cross-compile via `.build-windows.yaml`).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit eb289be0c04b000e83fca2346e2416f3aadb412c. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/.github/workflows/.installation-tests-matrix.yaml
+++ b/.github/workflows/.installation-tests-matrix.yaml
@@ -13,6 +13,13 @@ on:
         required: false
         type: string
         default: ''
+    secrets:
+      DR_API_TOKEN:
+        required: true
+
+# S5: deny all by default; jobs opt-in to what they need.
+permissions:
+  contents: read
 
 # S5: deny all by default; jobs opt-in to what they need.
 permissions:
--- a/.github/workflows/.installation-tests-matrix.yaml
+++ b/.github/workflows/.installation-tests-matrix.yaml
@@ -13,9 +13,6 @@ on:
         required: false
         type: string
         default: ''
-    secrets:
-      DR_API_TOKEN:
-        required: true
 
 # S5: deny all by default; jobs opt-in to what they need.
 permissions:
--- a/.github/workflows/.installation-tests-matrix.yaml
+++ b/.github/workflows/.installation-tests-matrix.yaml
@@ -18,10 +18,6 @@ on:
 permissions:
   contents: read
 
-# S5: deny all by default; jobs opt-in to what they need.
-permissions:
-  contents: read
-
 jobs:
   installation-tests:
     strategy:
--- a/.github/workflows/checks.yaml
+++ b/.github/workflows/checks.yaml
@@ -64,6 +64,9 @@ jobs:
     if: needs.detect-changes.outputs.go_code == 'true'
     runs-on: ubuntu-latest
     name: Lint Code
+    permissions:
+      contents: read
+      actions: write   # actions/cache (save)
 
     steps:
       - name: Checkout Code
@@ -91,7 +94,7 @@ jobs:
           path: |
             tmp/bin/golangci-lint
             ~/.cache/golangci-lint
-          key: golangci-lint-${{ runner.os }}-${{ steps.golangci-version.outputs.version }}-${{ hashFiles('**/*.go', 'go.sum', '.golangci.yml') }}
+          key: golangci-lint-${{ runner.os }}-${{ steps.golangci-version.outputs.version }}-${{ hashFiles('**/*.go', 'go.sum', '.golangci.yaml') }}
           restore-keys: |
             golangci-lint-${{ runner.os }}-${{ steps.golangci-version.outputs.version }}-
 
--- a/.github/workflows/checks.yaml
+++ b/.github/workflows/checks.yaml
@@ -80,6 +80,21 @@ jobs:
           version: 3.x
           repo-token: ${{ secrets.GITHUB_TOKEN }}
 
+      - name: Get golangci-lint version
+        id: golangci-version
+        run: |
+          echo "version=$(grep 'GOLANGCI_LINT_VERSION' Taskfile.yaml | grep -oP 'v[\d.]+')" >> "$GITHUB_OUTPUT"
+
+      - name: Cache golangci-lint binary and analysis
+        uses: actions/cache@0057852bfaa89a56745cba8c7296529d2fc39830 # v4
+        with:
+          path: |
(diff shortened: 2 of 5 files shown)


cdevent:
@cursor review

.github/workflows/.installation-tests-matrix.yaml
```diff
@@ -18,6 +18,10 @@ on:
 permissions:
   contents: read
 
+# S5: deny all by default; jobs opt-in to what they need.
+permissions:
+  contents: read
+
```
ajalon1, line 20: Duplicated

.github/workflows/checks.yaml
```diff
@@ -74,19 +74,6 @@ jobs:
         with:
           go-version: '1.26.4'
 
-      - name: golangci-lint
-        uses: golangci/golangci-lint-action@9fae48acfc02a90574d7c304a1758ef9895495fa # v7
-        with:
-          version: v2.11.4
-          args: --timeout=5m
```
ajalon1, line 81: Removing this action and using `task lint` in CI seems fine to me, but I did see that golangci-lint-action caches the analysis cache. It seems like we should benchmark this change first, then add caching as needed.

I wonder if there's a way to have `task lint` call golangci-lint directly in local and delegate to the golangci-lint action in CI.

.github/workflows/security-scan.yaml
```diff
@@ -26,28 +26,27 @@ jobs:
...
-          format: 'sarif'
-          output: 'trivy-results.sarif'
+          format: 'json'
+          output: 'trivy-results.json'
           severity: 'CRITICAL,HIGH,MEDIUM'
 
+      - name: Convert results to SARIF
+        # trivy is in PATH after the action step above.
+        run: trivy convert --format sarif --output trivy-results.sarif trivy-results.json
+
```
ajalon1, line 44: Claude tells me this needs an `if: always()` since the upload step has `if: always()`.

ajalon1 (commented):
some small nits

.github/workflows/.installation-tests-matrix.yaml
cdevent: yeah that's after rebase

cdevent:
@cursor review

.github/workflows/security-scan.yaml
cdevent: yeah , good catch...

.github/workflows/checks.yaml
cdevent: Looked with a fresh brain today and yes that's something that should be fixed, and i think we need return to the action instead....

cdevent: ok, one more why we removed: 
Run all checks step runs run lint what have inside go releaser and golang lint

ajalon1: OK, fine with removing the action, just should look into the caching. That can be a follow up, honestly, it doesn't need to happen here.

ajalon1 (approved):

PR #604, opened 2026-06-24
[CMPT-6415]: propagate pipelines-api latest changes

sunny2get:
# RATIONALE

Propagates the public API changes from `pipelines-api` and `pipelines-api` PR #154. This PR covers two waves of changes on top of the image schema work:

**Wave 1 — API gap sync (spec items missed in earlier PRs):**
- Schedule endpoints use a non-versioned URL path (`/api/v2/pipelines/{id}/schedules`, not `.../versions/{v}/schedules`); `--version` removed from all 5 schedule subcommands
- `PipelineScheduleCreateRequest` now requires `cronExpression`, `pipelineVersionId`, `imageId`, and `imageVersion` → `--image` and `--image-version` added as required flags to `schedule create`
- `imageId` is required (not optional) on run dispatch → `--image` marked required on `run create`
- `dr pipeline run task {list,get,logs,result}` — new subcommand group backed by `TaskExecution` client functions for `/api/v2/pipelines/{pid}/dispatches/{did}/tasks/…`
- `dr pipeline list --search` for name-substring filtering
- `RunStatusPreparing = "PREPARING"` constant

**Wave 2 — pipelines-api PR #154 (optional file on PATCH, name/description updates):**
- `PATCH /pipelines/{id}` no longer requires a file; `name` and `description` can be updated independently
- `--name` and `--description` flags added to `dr pipeline update`
- `CreateResponse` carries optional `Description` field

### `internal/pipeline/schedule.go`
- Fix URL: `scheduleBase(pipelineID)` → `/api/v2/pipelines/{id}/schedules` (non-versioned)
- Remove `version int` from all 5 function signatures
- `ScheduleCreateRequest`: add `CronExpression`, `PipelineVersionID`, `ImageID`, `ImageVersion`; `Schedule` struct: add `ImageID`, `ImageVersion`

### `cmd/pipeline/schedule/create`
- Add required `--image` and `--image-version` flags; remove `--version` from list/get/update/delete

### `internal/pipeline/run.go`
- Add `RunStatusPreparing = "PREPARING"`

### `cmd/pipeline/run/create`
- `--image` changed from optional to required

### `internal/pipeline/run_task.go` (new)
- `TaskExecution`, `TaskExecutionLogs`, `TaskExecutionDurableLog`, `TaskExecutionResult` structs
- `ListTaskExecutions`, `GetTaskExecution`, `GetTaskLogs`, `GetTaskDurableLog`, `GetTaskResult`

### `cmd/pipeline/run/task/` (new — 4 subcommands)
- `list`: table of task executions for a run
- `get`: detail view of a single task by sequential ID
- `logs`: live pod logs via `GetTaskLogs`; durable S3 logs via `GetTaskDurableLog` when `--stream stdout|stderr` is set
- `result`: show result URL, content type, expiry, and value preview

### `internal/pipeline/pipeline.go`
- `ListPipelines` signature: add `search string` parameter
- `CreateResponse`: add `Description *string`
- `UpdatePipeline` signature: add `name`, `description`; `filePath` is now optional
- `buildMultipartBody`: only write the `file` part when `filePath != ""`

### `cmd/pipeline/list`
- Add `--search` flag

### `cmd/pipeline/update`
- File is now optional; add `--name` and `--description`; validate at least one of file/name/description/image

### `internal/pipeline/pipeline_output.go`
- `printCreateResponseHuman`: show Description line when present

---

> [!NOTE]
> **Medium Risk**
> Schedule URL fix is a correctness change (prior code targeted a non-existent versioned endpoint). Run task commands are additive. PR #154 changes make file optional without breaking existing callers.

### DIFF
--- a/Taskfile.yaml
+++ b/Taskfile.yaml
@@ -160,22 +160,6 @@ tasks:
     cmds:
       - ./smoke_test_scripts/run_pre_release_smoke_test.sh {{if .DR_API_TOKEN}}{{.DR_API_TOKEN}}{{else}}$DR_API_TOKEN{{end}}
 
-  demo-pipelines:
-    desc: "End-to-end demo of every dr pipelines command. Defaults to interactive (keypress per step); pass DELAY=N or `-- --delay=N` for auto-advance. Set SKIP_SCHEDULES=true or SKIP_ENVIRONMENTS=true to skip steps that require k8s/Covalent."
-    deps: [build]
-    vars:
-      DEMO_DIR: '{{default "/Users/sunnypal.sharma/Desktop/tests/pipelines-api" .DEMO_DIR}}'
-      DELAY_FLAG: '{{if .DELAY}}--delay={{.DELAY}}{{end}}'
-    env:
-      DATAROBOT_CLI_FEATURE_PIPELINE: "true"
-      # DATAROBOT_CLI_ENDPOINT: "http://localhost:8100/api/v2"
-      # DATAROBOT_CLI_TOKEN: "local"
-      # DATAROBOT_CLI_SKIP_AUTH: "true"
-      DEMO_SKIP_SCHEDULES: '{{if eq .SKIP_SCHEDULES "false"}}false{{else}}true{{end}}'
-      DEMO_SKIP_IMAGES: '{{if eq .SKIP_IMAGES "false"}}false{{else}}true{{end}}'
-    cmds:
-      - DR_BIN="$PWD/dist/dr" DEMO_DIR="{{.DEMO_DIR}}" bash "{{.DEMO_DIR}}/demo.sh" {{.DELAY_FLAG}} {{.CLI_ARGS}} 2>&1 | tee output.txt
-
   smoke-test-self-update:
     desc: "Run self-update smoke tests"
     cmds:
@@ -199,8 +183,8 @@ tasks:
     cmds:
       - echo "📚 Starting documentation server…"
       - uv sync
-      - echo "🌐 Open http://localhost:8001 in your browser"
-      - uv run mkdocs serve --dev-addr localhost:8001
+      - echo "🌐 Open http://localhost:8000 in your browser"
+      - uv run mkdocs serve
 
   copyright:
     silent: true
--- a/Taskfile.yaml
+++ b/Taskfile.yaml
@@ -160,6 +160,22 @@ tasks:
     cmds:
       - ./smoke_test_scripts/run_pre_release_smoke_test.sh {{if .DR_API_TOKEN}}{{.DR_API_TOKEN}}{{else}}$DR_API_TOKEN{{end}}
 
+  demo-pipelines:
+    desc: "End-to-end demo of every dr pipelines command. Defaults to interactive (keypress per step); pass DELAY=N or `-- --delay=N` for auto-advance. Set SKIP_SCHEDULES=true or SKIP_ENVIRONMENTS=true to skip steps that require k8s/Covalent."
+    deps: [build]
+    vars:
+      DEMO_DIR: '{{default "/Users/sunnypal.sharma/Desktop/tests/pipelines-api" .DEMO_DIR}}'
+      DELAY_FLAG: '{{if .DELAY}}--delay={{.DELAY}}{{end}}'
+    env:
+      DATAROBOT_CLI_FEATURE_PIPELINE: "true"
+      # DATAROBOT_CLI_ENDPOINT: "http://localhost:8100/api/v2"
+      # DATAROBOT_CLI_TOKEN: "local"
+      # DATAROBOT_CLI_SKIP_AUTH: "true"
+      DEMO_SKIP_SCHEDULES: '{{if eq .SKIP_SCHEDULES "false"}}false{{else}}true{{end}}'
+      DEMO_SKIP_IMAGES: '{{if eq .SKIP_IMAGES "false"}}false{{else}}true{{end}}'
+    cmds:
+      - DR_BIN="$PWD/dist/dr" DEMO_DIR="{{.DEMO_DIR}}" bash "{{.DEMO_DIR}}/demo.sh" {{.DELAY_FLAG}} {{.CLI_ARGS}} 2>&1 | tee output.txt
+
   smoke-test-self-update:
     desc: "Run self-update smoke tests"
     cmds:
@@ -183,8 +199,8 @@ tasks:
     cmds:
       - echo "📚 Starting documentation server…"
       - uv sync
-      - echo "🌐 Open http://localhost:8000 in your browser"
-      - uv run mkdocs serve
+      - echo "🌐 Open http://localhost:8001 in your browser"
+      - uv run mkdocs serve --dev-addr localhost:8001
 
   copyright:
     silent: true
--- a/cmd/pipeline/create/cmd.go
+++ b/cmd/pipeline/create/cmd.go
@@ -32,6 +32,7 @@ func Cmd() *cobra.Command {
 		description  string
 		name         string
 		mode         string
+		imageID      string
 		outputFormat outputformat.OutputFormat
 		fromFile     string
 	)
@@ -52,6 +53,7 @@ Example:
   dr pipeline create ./my_pipeline.py
   dr pipeline create --from-file=./my_pipeline.py
   dr pipeline create ./my_pipeline.py --name "My Pipeline" --description "First draft" --mode draft
+  dr pipeline create ./my_pipeline.py --image <image-id>
   dr pipeline create --from-file=./my_pipeline.py --output-format json`,
 		Args:         cobra.MaximumNArgs(1),
 		SilenceUsage: true,
@@ -74,7 +76,7 @@ Example:
 				return err
 			}
 
-			result, err := pipeline.CreatePipeline(filePath, description, name, mode)
+			result, err := pipeline.CreatePipeline(filePath, description, name, mode, imageID)
 			if err != nil {
 				return err
 			}
@@ -88,6 +90,7 @@ Example:
 	cmd.Flags().StringVar(&description, "description", "", "Optional description for the pipeline")
 	cmd.Flags().StringVar(&name, "name", "", "Optional human-readable display name; defaults to the title-cased @pipeline function name")
 	cmd.Flags().StringVar(&mode, "mode", "", "Pipeline mode: draft or locked")
+	cmd.Flags().StringVar(&imageID, "image", "", "Execution image ID to associate with this pipeline")
 	cmd.Flags().StringVar(&fromFile, "from-file", "", "Path to the Python file to upload, e.g. --from-file=./my_pipeline.py (alternative to the positional argument)")
 
 	telemetry.TrackWith(cmd, func(_ *cobra.Command, _ []string) map[string]any {
--- a/cmd/pipeline/image/cmd.go
+++ b/cmd/pipeline/image/cmd.go
@@ -17,6 +17,7 @@ package image
 import (
 	"github.com/datarobot/cli/cmd/pipeline/image/create"
 	"github.com/datarobot/cli/cmd/pipeline/image/del"
+	"github.com/datarobot/cli/cmd/pipeline/image/get"
 	"github.com/datarobot/cli/cmd/pipeline/image/list"
 	"github.com/datarobot/cli/cmd/pipeline/image/update"
 	"github.com/datarobot/cli/cmd/pipeline/image/version"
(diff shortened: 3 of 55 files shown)


internal/pipeline/image_output.go
```diff
@@ -187,11 +201,23 @@ func PrintImageHuman(img Image) {
...
+	for _, ver := range versions {
+		baseImageStr := emptyValuePlaceholder
+		if ver.Definition.BaseImage != nil && *ver.Definition.BaseImage != "" {
+			baseImageStr = *ver.Definition.BaseImage
+		}
+
+		condaStr := emptyValuePlaceholder
+		if ver.Definition.Conda != nil && len(ver.Definition.Conda.Deps) > 0 {
+			condaStr = joinPackages(ver.Definition.Conda.Deps)
+		}
```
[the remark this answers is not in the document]
sunny2get: fixed in latest commit

internal/pipeline/image_output.go
```diff
@@ -276,10 +328,14 @@ func joinPackages(packages []string) string {
 	return joined[:maxLen-3] + "..."
 }
```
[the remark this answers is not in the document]
sunny2get: fixed

cmd/pipeline/image/cmd.go
```diff
@@ -25,18 +25,18 @@ import (
...
 		Long: `Manage pipeline execution images.
 
-Images are named, immutable-versioned bags of pip packages that
-pipelines can be built against. Each ` + "`update`" + ` adds packages by
-creating a new version; older versions can be deleted individually with
-` + "`image version delete`" + `.`,
+Images are named, immutable-versioned execution environments backed by
+pip packages that pipelines can be built against. Each ` + "`update`" + ` creates
+a new version with a complete replacement definition; older versions can be
+deleted individually with ` + "`image version delete`" + `.`,
```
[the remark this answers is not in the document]
sunny2get: fixed

cmd/pipeline/create/cmd.go
```diff
@@ -88,6 +90,7 @@ Example:
 	cmd.Flags().StringVar(&description, "description", "", "Optional description for the pipeline")
 	cmd.Flags().StringVar(&name, "name", "", "Optional human-readable display name; defaults to the title-cased @pipeline function name")
 	cmd.Flags().StringVar(&mode, "mode", "", "Pipeline mode: draft or locked")
+	cmd.Flags().StringVar(&imageID, "image", "", "Execution image ID to associate with this pipeline")
```
ajalon1, line 93: I assume users will know how to find this execution image ID.

cmd/pipeline/image/create/cmd.go
```diff
@@ -38,12 +42,15 @@ func Cmd() *cobra.Command {
 		Long: `Create a new pipeline execution image.
 
 A new image is registered with an initial version (v1) containing the
-supplied pip packages. The image may be referenced by pipelines once
+supplied package definition. The image may be referenced by pipelines once
 its first version reaches the READY state.
 
+At least one of --package (pip) or --conda must be provided.
```
ajalon1, line 48: 🐍 !

cmd/pipeline/image/create/cmd.go
```diff
@@ -54,12 +61,18 @@ Example:
...
-				return err
+			pip := pipeline.NormalizePackageList(rawPackages)
+			conda := pipeline.BuildCondaValue(rawConda, rawCondaChans)
+
+			if len(pip) == 0 && conda == nil {
+				return errors.New("at least one of --package (pip) or --conda is required")
+			}
+
+			if conda != nil && len(conda.Deps) == 0 {
+				return errors.New("--conda-channel requires at least one --conda package")
```
ajalon1, line 72: Is the inverse true? --conda requires --conda-channel? It doesn't make sense to me, I can see using the default conda channel for packages, but just want to be sure.

cmd/pipeline/source/cmd.go
```diff
@@ -0,0 +1,116 @@
...
+	var httpErr *drapi.HTTPError
+
+	if errors.As(err, &httpErr) && httpErr.StatusCode == http.StatusNotFound {
+		fmt.Println(tui.DimStyle.Render("No source available for pipeline: " + pipelineID))
+
+		return nil
+	}
+
+	return err
+}
```
ajalon1: when output format is JSON, I think you should either raise an error or use PrintJSONEnvelope() and add an error message within the envelope object.

ajalon1 (commented):
Looks pretty reasonable. Some general questions given the pipelines API has changed again. Are you concerned with users dealing w. backwards incompatible changes? I assume not given there's still a feature gate.

Also, I made some enhancements to `--output-format json` in #582 and I think you should consider using them, specifically PrintJSONEnvelope().

cmd/pipeline/create/cmd.go
sunny2get: yep, its one of the pre-requisites to run the pipeline. We will make sure the docs reflect that.

cmd/pipeline/image/get/cmd.go
```diff
@@ -0,0 +1,78 @@
...
+				if errors.As(err, &httpErr) && httpErr.StatusCode == http.StatusNotFound {
+					fmt.Println(tui.DimStyle.Render("No image found: " + imageID))
+
+					return nil
+				}
+
+				return err
+			}
+
+			return pipeline.RenderImage(outputFormat, *img)
```
ajalon1, line 66: It looks like this one doesn't have `outputFormat = outputformat.GetFormat(cmd)`

cmd/pipeline/update/cmd.go
```diff
@@ -25,27 +26,31 @@ import (
...
+(name/description/image) without re-uploading the source file.
 
-The path to the Python file can be supplied either as a positional argument
-or via the --from-file=<path> flag. Exactly one of the two must be provided.
+When a .py file is provided (positional or --from-file), a new version is
+appended. The pipeline name encoded in the uploaded file must match the
+existing pipeline name. Locked pipelines cannot be updated.
 
-By default, output is human-readable. Use --output-format json for machine-parseable output.
+At least one of --file, --name, --description, or --image must be supplied.
```
ajalon1: Are you sure on "At least one of --file, --name, --description, or --image must be supplied." Specifically, `--file`.

Taskfile.yaml
```diff
@@ -160,6 +160,22 @@ tasks:
...
+      DELAY_FLAG: '{{if .DELAY}}--delay={{.DELAY}}{{end}}'
+    env:
+      DATAROBOT_CLI_FEATURE_PIPELINE: "true"
+      # DATAROBOT_CLI_ENDPOINT: "http://localhost:8100/api/v2"
+      # DATAROBOT_CLI_TOKEN: "local"
+      # DATAROBOT_CLI_SKIP_AUTH: "true"
+      DEMO_SKIP_SCHEDULES: '{{if eq .SKIP_SCHEDULES "false"}}false{{else}}true{{end}}'
+      DEMO_SKIP_IMAGES: '{{if eq .SKIP_IMAGES "false"}}false{{else}}true{{end}}'
+    cmds:
+      - DR_BIN="$PWD/dist/dr" DEMO_DIR="{{.DEMO_DIR}}" bash "{{.DEMO_DIR}}/demo.sh" {{.DELAY_FLAG}} {{.CLI_ARGS}} 2>&1 | tee output.txt
```
ajalon1: I assume this is a temporary feature while you're testing? If it is, then should mark it as such and when it'll be removed. If it isn't, I'd suggest moving this to a new `/Taskfile.pipelines.yaml` codeowned by your team, and then using

```
includes:
  pipelines: ./Taskfile.pipelines.yaml
```

Taskfile.yaml
```diff
@@ -160,6 +160,22 @@ tasks:
     cmds:
       - ./smoke_test_scripts/run_pre_release_smoke_test.sh {{if .DR_API_TOKEN}}{{.DR_API_TOKEN}}{{else}}$DR_API_TOKEN{{end}}
 
+  demo-pipelines:
+    desc: "End-to-end demo of every dr pipelines command. Defaults to interactive (keypress per step); pass DELAY=N or `-- --delay=N` for auto-advance. Set SKIP_SCHEDULES=true or SKIP_ENVIRONMENTS=true to skip steps that require k8s/Covalent."
+    deps: [build]
+    vars:
+      DEMO_DIR: '{{default "/Users/sunnypal.sharma/Desktop/tests/pipelines-api" .DEMO_DIR}}'
```
ajalon1: This should not live in the codebase. 😅

Taskfile.yaml
```diff
@@ -183,8 +199,8 @@ tasks:
     cmds:
       - echo "📚 Starting documentation server…"
       - uv sync
-      - echo "🌐 Open http://localhost:8000 in your browser"
-      - uv run mkdocs serve
+      - echo "🌐 Open http://localhost:8001 in your browser"
+      - uv run mkdocs serve --dev-addr localhost:8001
```
ajalon1: Why is this 8000? Can you revert this?

internal/pipeline/image.go
```diff
@@ -149,16 +237,81 @@ func ListImages(offset, limit int) ([]ImageSummary, error) {
...
+		return nil, err
+	}
+
+	// Fetch the current image to resolve its canonical name — required by
+	// the update request body even though the server overrides it with the
+	// stored name anyway.
+	current, err := GetImage(imageID)
+	if err != nil {
+		return nil, err
+	}
```
ajalon1, line 309: What would this error look like in practice? I think we could at least log.Warn that one was not found.

internal/pipeline/run_task.go
```diff
@@ -0,0 +1,189 @@
...
+}
+
+// ListTaskExecutions returns all task execution records for a run.
+func ListTaskExecutions(pipelineID, runID string) ([]TaskExecution, error) {
+	endpoint, err := taskBase(pipelineID, runID)
+	if err != nil {
+		return nil, err
+	}
+
+	var tasks []TaskExecution
```
ajalon1, line 81: Shouldn't this be a DataPage?

internal/pipeline/run_task.go
```diff
@@ -0,0 +1,189 @@
...
+
+	err = doJSON(http.MethodGet, endpoint, nil, "task executions", &tasks)
```
ajalon1: What about drapi.GetJSON?

ajalon1 (changes requested):
This LGTM on the assumption that we don't care about older versions of the Pipelines API (on older DR builds). :)

Requesting changes for:
- outputFormat (which is a bug)
- docs on pipeline update
- Taskfile

I see some places where you might want to add additional logging (log.Debug, log.Warn) to help with addressing user issues in the future. Up to you.

Everything else is informational and not a blocker.

Taskfile.yaml
sunny2get: yeah, apologies - this needs to be removed

internal/pipeline/image.go
sunny2get: updated

cmd/pipeline/image/get/cmd.go
sunny2get: updated

sunny2get:
@ajalon1 - please take a look when you get another chance as I have gone thru your comments and addressed as below:

 - outputFormat bug — added outputFormat = outputformat.GetFormat(cmd) at the top of RunE in
  image/get/cmd.go so the output format flag is respected correctly.
  - docs on pipeline update — fixed the description, this endpoint has been changed to accomodate to allow just updating name OR description OR update the pipeline source code
  - Taskfile — removed the demo-pipelines task and reverted docs-serve back to port 8000.

  Also took your suggestion on logging — added a log.Warn in UpdateImage when a 404 is returned from
  GetImage, so users get a visible signal when trying to update a non-existent image.

cmd/pipeline/update/cmd.go
sunny2get: yep, API surface changed to make all of them optional by default as per UI folks request.

internal/pipeline/run_task.go
sunny2get: No — the task list endpoint (GET /dispatches/{id}/tasks) returns a plain JSON array like so:
image

internal/pipeline/run_task.go
sunny2get: Good catch — updated

internal/pipeline/run_task.go
ajalon1: Got it. Well, that's an anti-pattern we don't want in the Public API (bare JSON arrays), and I'd suggest the team update that before pipelines goes GA, but that's beyond the scope of this PR :)

ajalon1 (approved):

PR #605, opened 2026-06-24
[CFX-6634] fix(update): prevent circular dr symlink

ajalon1:
# RATIONALE

In DataRobot Codespaces, running `dr self update` twice left `dr` a self-referential symlink (`dr → dr`), breaking the CLI with "too many levels of symbolic links". Codespaces install `dr` as a directory on `PATH` (`dr/dr`) rather than a flat file; `install.sh`'s `ln -sf dr datarobot` then followed the `datarobot → dr` directory symlink on the second run and wrote the new link *inside* the directory, clobbering the real binary. This is critical (CFX-6634) since the CLI becomes unusable, and reproduces only in Codespaces.

## CHANGES

- `install.sh`: add `ensure_datarobot_alias()` using `rm -f` + `ln -sfn` so `ln` never dereferences a symlinked directory.
- `cmd/self/update`: pass `INSTALL_DIR` resolved from the running binary so updates land where `dr` actually lives.
- Add `resolveInstallDir` unit tests and an `install.sh` alias regression smoke test.

## NOTES

- Fix reaches users as soon as it merges to `main` (`dr self update` fetches `install.sh` from `main`); the Go change ships in the next build.
- Already-broken Codespaces recover by re-running the installer directly.

## TESTING

- `task lint` (0 issues), `go test -race ./cmd/self/...`, install alias + install integration smoke tests all pass.

## PR Automation

> [!IMPORTANT]
> **Forked PR:** the `run-smoke-tests` label won't work — a maintainer needs to `/approve-smoke-tests` or `/skip-smoke-tests` to clear the required Smoke Tests check.

---

> [!NOTE]
> **Medium Risk**
> Touches self-update and install scripting paths that can brick the CLI if wrong, but changes are narrow and covered by unit and smoke tests.
> 
> **Overview**
> Fixes a **Codespaces-only** failure where a second `dr self update` could replace the real binary with a self-referential `dr → dr` symlink (“too many levels of symbolic links”).
> 
> **`install.sh`** centralizes alias creation in `ensure_datarobot_alias()`: remove any existing `datarobot` entry, then `ln -sfn` so `ln` does not follow a `datarobot → dr` directory symlink and write a link inside `dr/dr`. The script can be sourced for tests via `DR_INSTALL_SH_NO_MAIN`.
> 
> **`dr self update`** sets `INSTALL_DIR` from the running binary’s directory (symlinks resolved) when the user has not set `INSTALL_DIR`, so the fetched installer updates in place instead of defaulting to `~/.local/bin`.
> 
> Adds Go tests for `resolveInstallDir` and a shell smoke test for the alias behavior in Codespace vs flat layouts.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 24ebb9088200798237542f9ef546a831a486ff35. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -316,14 +316,9 @@ download_and_install() {
 }
 
 # Create or refresh the 'datarobot' alias symlink (datarobot -> dr).
-#
-# Uses `ln -sfn` (no-dereference; supported by both GNU and BSD/macOS ln) and
-# removes any existing entry first. This is critical: if `dr` is installed as a
-# directory on PATH (as in DataRobot Codespaces, where the binary lives at
-# dr/dr) then a prior run leaves `datarobot` as a symlink to that directory.
-# A plain `ln -sf dr datarobot` would follow that symlink and create the new
-# link *inside* the directory, clobbering the real binary at dr/dr with a
-# self-referential `dr -> dr` symlink ("too many levels of symbolic links").
+# Removes the existing entry first to avoid ln following a symlink into a
+# directory (e.g. Codespaces installs dr as dr/dr) and creating a
+# self-referential link inside it.
 ensure_datarobot_alias() {
     rm -f "$INSTALL_DIR/datarobot"
     ln -sfn "$BINARY_NAME" "$INSTALL_DIR/datarobot"
--- a/install.sh
+++ b/install.sh
@@ -199,7 +199,7 @@ check_existing_installation() {
             # Ensure datarobot alias symlink exists
             if [ ! -L "$INSTALL_DIR/datarobot" ]; then
                 step "Creating missing 'datarobot' alias..."
-                ln -sf "$BINARY_NAME" "$INSTALL_DIR/datarobot"
+                ensure_datarobot_alias
             fi
 
             if ! echo ":$PATH:" | grep -q ":$INSTALL_DIR:"; then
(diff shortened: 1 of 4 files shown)


ajalon1:
Putting this out here, but it is untested. If someone wants to attempt to test this, they can tag a new dev release from this branch, let the publish workflow complete, and then do a `dr self update -f <version>` twice in a Codespace to verify.

install.sh
```diff
@@ -312,7 +312,21 @@ download_and_install() {
...
+# Uses `ln -sfn` (no-dereference; supported by both GNU and BSD/macOS ln) and
+# removes any existing entry first. This is critical: if `dr` is installed as a
+# directory on PATH (as in DataRobot Codespaces, where the binary lives at
+# dr/dr) then a prior run leaves `datarobot` as a symlink to that directory.
+# A plain `ln -sf dr datarobot` would follow that symlink and create the new
+# link *inside* the directory, clobbering the real binary at dr/dr with a
+# self-referential `dr -> dr` symlink ("too many levels of symbolic links").
+ensure_datarobot_alias() {
+    rm -f "$INSTALL_DIR/datarobot"
+    ln -sfn "$BINARY_NAME" "$INSTALL_DIR/datarobot"
```
ajalon1, line 324: ln -sf -> ln -sfn is the biggest change here.

c-h-russell-walker: Had to use `man ln`

```
 -n    Same as -h, for compatibility with other ln implementations.
```

```
-h    If the target_file or target_dir is a symbolic link, do not follow it.  This is most useful with the -f option, to replace a symlink which may point to a directory.
```

c-h-russell-walker (approved):
Very nice - looks great 👍

victorborshak (approved):

PR #607, opened 2026-06-26
[CFX-6647] test(completion,windows): windows smoke tests for exercising `dr self completion install`

ajalon1:
# RATIONALE

On a fresh Windows machine, `dr self completion install` writes a PowerShell profile that won't load under the default `Restricted` execution policy. This smoke test proves the bug by exercising `dr self completion install powershell --yes` under both `Restricted` and `RemoteSigned` policies.

The Restricted-policy warning assertion is marked as **xfail** (expected failure) — it will pass once Phase 2 (CFX-6647-phase2, the Go fix in #702) is merged. CI stays green.

Related: CFX-6647

## CHANGES

Adds Windows smoke test coverage for `dr self completion install` with execution policy handling:

- `Test-DRCompletionInstallWithExecutionPolicy` function with parametrized policy, warning expectation, and xfail support
- Tests under `Restricted` policy: expects `Write-XFailMsg` (warning not yet emitted; will pass after Phase 2)
- Tests under `RemoteSigned` policy: expects silent completion (no warning)
- Asserts profile is created with completion blocks
- Asserts the installer never mutates the execution policy
- `Write-XFailMsg` helper logs expected failures in yellow without exiting
- `$global:XFailCount` tracks expected failures for a summary at the end
- `exit 0` even when xfail assertions fire

### Taskfile fix: PSModulePath clearing

GitHub Actions `windows-latest` runners use `pwsh` (PowerShell 7) as the default shell. When `task` launches `powershell.exe` (5.1) via the Taskfile, it inherits pwsh 7's `PSModulePath`, which contains incompatible PowerShell 7 module paths. This causes `Microsoft.PowerShell.Security` to fail to load and `Get-ExecutionPolicy` to be unavailable.

Fix: `PSModulePath= powershell.exe ...` (POSIX env-prefix assignment, since Taskfile uses mvdan/sh) clears the env var so 5.1 reconstructs its own default module path.

Ref: https://github.com/PowerShell/PowerShell/issues/18530#issuecomment-1325691850
Ref: https://github.com/actions/runner-images/issues/13221

### Execution policy scope: Process instead of CurrentUser

`Set-ExecutionPolicy -Scope CurrentUser` fails on GHA Windows runners (Group Policy locks down registry-based policy changes). Switched to `-Scope Process`, which only affects the current PowerShell session and doesn't touch the registry. `Get-ExecutionPolicy` (effective, no `-Scope`) correctly reflects the Process-scope override.

The `$fixCommand` string literal still says `-Scope CurrentUser` — that's the user-facing advice the Go installer should print (users should set it at CurrentUser scope, not Process scope).

### Safeguards

- `Get-ExecutionPolicy` guard: try/catch to skip gracefully if the security module can't load
- `Set-ExecutionPolicy` guard: try/catch to skip gracefully if the runner restricts policy changes
- `--force` flag on installer: ensures real install each time instead of short-circuiting on existing profile
- Profile cleanup before/after each test for isolation
- EAP relaxation around installer call (consistent with the `--debug` test pattern)

## TESTING

- Windows smoke test CI: `task smoke-test-windows` exercises the full flow
- CI run verified: Restricted test xfails correctly, RemoteSigned test passes all assertions (warning behavior, profile creation, policy preservation)

```
==================== Testing dr self completion install (Restricted execution policy) ====================
[XFAIL] Assertion xfail [Restricted]: installer did not warn user with the execution policy fix command (expected: Phase 2 not yet merged)

==================== Testing dr self completion install (permissive execution policy) ====================
[OK] Assertion passed [RemoteSigned]: warning behavior correct
[OK] Assertion passed: profile contains completion block
[OK] Assertion passed [RemoteSigned]: execution policy remains 'RemoteSigned'

==================== XFAIL SUMMARY ====================
1 expected failure(s) (xfail) -- these will pass after Phase 2 (CFX-6647 Go fix) is merged.
[OK] Smoke tests for Windows completed successfully.
```

## NOTES

- **Follow-up needed**: Add pwsh 7 (PowerShell Core) smoke test coverage. Currently only `powershell.exe` (5.1) is tested. Real users may have either version, and profile paths, module availability, and execution policy handling differ between them.
- **Phase 2 dependency**: The XFAIL for the Restricted warning assertion will become a hard failure once #702 is merged. At that point, remove the `$isXFail` branch and the XFAIL summary logic.
- **PSModulePath fix is in the Taskfile** (not the script) because that's where `powershell.exe` is launched from the pwsh parent. The fix applies to anyone running `task smoke-test-windows` from a pwsh 7 terminal, not just CI.

## RELATED

- #702 — Phase 2 Go fix: warn Windows users when execution policy blocks PowerShell profile
- #689 — Add windows-latest to the unit-test CI matrix (no conflict; different Taskfile tasks)

## PR Automation

> [!IMPORTANT]
>

### DIFF
--- a/smoke_test_scripts/windows/run_smoke_test.ps1
+++ b/smoke_test_scripts/windows/run_smoke_test.ps1
@@ -150,6 +150,54 @@ if (Test-Path $completion_file) {
 }
 Write-End
 
+# Test completion install under a fresh-machine execution policy.
+# This intentionally exposes the bug: dr self completion install writes a
+# profile that PowerShell refuses to load when the default Restricted policy
+# is in effect. The CLI does not yet fix the policy, so the assertion below
+# expects the policy to remain Restricted after install.
+Write-Delimiter "Testing dr self completion install (Restricted execution policy)"
+
+$originalPolicy = Get-ExecutionPolicy -Scope CurrentUser
+
+# Simulate a fresh Windows machine where the default execution policy is Restricted.
+Set-ExecutionPolicy Restricted -Scope CurrentUser -Force
+
+# Run the completion installer. This is the exact user path that is broken.
+dr self completion install powershell --yes
+if ($LASTEXITCODE -ne 0) {
+    Set-ExecutionPolicy $originalPolicy -Scope CurrentUser -Force -ErrorAction SilentlyContinue
+    Write-ErrorMsg "dr self completion install powershell --yes failed"
+}
+
+# Verify the completion profile was written.
+$profilePath = "$env:USERPROFILE\Documents\WindowsPowerShell\Microsoft.PowerShell_profile.ps1"
+if (-not (Test-Path $profilePath)) {
+    $profilePath = "$env:USERPROFILE\Documents\PowerShell\Microsoft.PowerShell_profile.ps1"
+}
+if (-not (Test-Path $profilePath)) {
+    Set-ExecutionPolicy $originalPolicy -Scope CurrentUser -Force -ErrorAction SilentlyContinue
+    Write-ErrorMsg "Assertion failed: PowerShell profile was not created"
+}
+$profileContent = Get-Content $profilePath -Raw
+if ($profileContent -notmatch "dr completion powershell") {
+    Set-ExecutionPolicy $originalPolicy -Scope CurrentUser -Force -ErrorAction SilentlyContinue
+    Write-ErrorMsg "Assertion failed: profile does not contain completion block"
+}
+Write-SuccessMsg "Assertion passed: profile written and contains completion block"
+
+# Verify the bug: the installer does NOT change the execution policy, so a
+# fresh machine remains Restricted and the profile will not load next session.
+$policy = Get-ExecutionPolicy -Scope CurrentUser
+if ($policy -ne "Restricted") {
+    Set-ExecutionPolicy $originalPolicy -Scope CurrentUser -Force -ErrorAction SilentlyContinue
+    Write-ErrorMsg "Assertion failed: expected execution policy to remain Restricted (proving the bug), but it is '$policy'"
+}
+Write-SuccessMsg "Assertion passed: execution policy remains '$policy' after install, confirming the bug"
+
+# Restore the original policy so the rest of the smoke test is not affected.
+Set-ExecutionPolicy $originalPolicy -Scope CurrentUser -Force -ErrorAction SilentlyContinue
+Write-End
+
 # Test dr run command
 Write-Delimiter "Testing dr run command"
 dr run
--- a/smoke_test_scripts/windows/run_smoke_test.ps1
+++ b/smoke_test_scripts/windows/run_smoke_test.ps1
@@ -25,6 +25,7 @@ function Write-InfoMsg {
     Write-Host $Message
 }
 
+# Prints a section header banner centered around the given message.
 function Write-Delimiter {
     param([string]$Message)
     Write-Host ""
@@ -35,6 +36,7 @@ function Write-Delimiter {
     Write-Host ("=" * 20)
 }
 
+# Prints a closing END banner to delimit test sections.
 function Write-End {
     Write-Host ("=" * 20) -NoNewline
     Write-Host " END " -NoNewline
@@ -51,6 +53,7 @@ function Test-URLAccessible {
     }
 }
 
+# Installs PowerShell completions under a given execution policy and asserts warning behavior, profile contents, and policy preservation.
 function Test-DRCompletionInstallWithExecutionPolicy {
     param(
         [string]$TestName,
@@ -59,9 +62,11 @@ function Test-DRCompletionInstallWithExecutionPolicy {
         [bool]$ExpectWarning
     )
 
+    # Save the current policy so it can be restored after the test, then apply the policy under test.
     $originalPolicy = Get-ExecutionPolicy -Scope CurrentUser
     Set-ExecutionPolicy $Policy -Scope CurrentUser -Force
 
+    # Run the installer and fail fast if it exits non-zero.
     $installOutput = (dr self completion install powershell --yes 2>&1 | Out-String)
     $installExitCode = $LASTEXITCODE
     if ($installExitCode -ne 0) {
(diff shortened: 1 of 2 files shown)


ajalon1:
Additional work in progress: fixing the `Get-ExecutionPolicy` cmdlet crash on the Windows runner and converting the execution-policy warning assertion to a pending marker. I want to actually implement a fix, but in a followup PR.

ajalon1:
You can see from the [Harness run](https://github.com/datarobot-oss/cli/actions/runs/28209098484):

```
==================== Testing shell detection ====================
[INFO] Running dr --debug self version to verify shell detection...
[OK] Assertion passed: Shell detection correctly identified PowerShell.
==================== END ====================

==================== Testing completion generation ====================
[OK] Assertion passed: We have expected completion_powershell.ps1 file with Register-ArgumentCompleter.
==================== END ====================

==================== Testing dr self completion install (Restricted execution policy) ====================
Get-ExecutionPolicy : The 'Get-ExecutionPolicy' command was found in the module 'Microsoft.PowerShell.Security', but 
the module could not be loaded. For more information, run 'Import-Module Microsoft.PowerShell.Security'.
At D:\a\cli\cli\smoke_test_scripts\windows\run_smoke_test.ps1:66 char:23
+     $originalPolicy = Get-ExecutionPolicy -Scope CurrentUser
+                       ~~~~~~~~~~~~~~~~~~~
    + CategoryInfo          : ObjectNotFound: (Get-ExecutionPolicy:String) [], ParentContainsErrorRecordException
    + FullyQualifiedErrorId : CouldNotAutoloadMatchingModule
 
task: Failed to run task "smoke-test-windows": exit status 1
Error: exit status 1
```

This should pass in #702

smoke_test_scripts/windows/run_smoke_test.ps1
```diff
@@ -51,6 +62,91 @@ function Test-URLAccessible {
...
+        Set-ExecutionPolicy $originalPolicy -Scope CurrentUser -Force -ErrorAction SilentlyContinue
+        Write-ErrorMsg "dr self completion install powershell --yes failed with exit code $installExitCode"
+    }
+
+    # Assert the installer warned (or did not warn) about the execution policy fix command.
+    $fixCommand = "Set-ExecutionPolicy RemoteSigned -Scope CurrentUser"
+    $hasWarning = $installOutput -match $fixCommand
+
+    if ($ExpectWarning -and -not $hasWarning) {
+        Set-ExecutionPolicy $originalPolicy -Scope CurrentUser -Force -ErrorAction SilentlyContinue
```
chasdr: 👋 chas's claude again — think your agent tangled the xfail cleanup.

line 91 restores $originalPolicy inside the xfail branch, but Write-XFailMsg doesn't exit, so control falls through to the "execution policy remains Restricted" check at 139. by then the policy is back to the restored value (Undefined on a fresh box), so it mismatches $ExpectedPolicy="Restricted" and hits Write-ErrorMsg → exit 1.

so the Restricted xfail doesn't actually soften — it hard-fails the whole smoke run unless the machine's CurrentUser policy already happens to be Restricted. want a `return` right after Write-XFailMsg (restore then return) so it skips the remaining assertions? that also drops the stray "[OK] warning behavior correct" that currently prints after the [XFAIL].

ajalon1: And here I am trying to bootstrap an AI dark factory. Thanks I'll revisit.

ajalon1:
pushing some tests now, keeping in draft until this works
ETA: never let something else do a git rebase for you

ajalon1:
```
==================== Testing completion generation ====================
[OK] Assertion passed: We have expected completion_powershell.ps1 file with Register-ArgumentCompleter.
==================== END ====================

==================== Testing dr self completion install (Restricted execution policy) ====================
[INFO] Skipping [Restricted]: Get-ExecutionPolicy cmdlet not available on this system.
==================== END ====================

==================== Testing dr self completion install (permissive execution policy) ====================
[INFO] Skipping [RemoteSigned]: Get-ExecutionPolicy cmdlet not available on this system.
==================== END ====================
```

Looks like an issue with powershell 7. PSModulePath issues. See:

https://github.com/actions/runner-images/issues/13221
https://github.com/PowerShell/PowerShell/issues/18530#issuecomment-1325691850

I can see this happening locally too, if a Windows dev has PS v7 installed (reference https://learn.microsoft.com/en-us/powershell/scripting/whats-new/migrating-from-windows-powershell-51-to-powershell-7?view=powershell-7.6).

Testing a fix in Taskfile, fe3220d

smoke_test_scripts/windows/run_smoke_test.ps1
ajalon1: e0dc224

ajalon1:
Are you kidding me

```
==================== Testing completion generation ====================
[OK] Assertion passed: We have expected completion_powershell.ps1 file with Register-ArgumentCompleter.
==================== END ====================

==================== Testing dr self completion install (Restricted execution policy) ====================
[INFO] Skipping [Restricted]: unable to set execution policy to Restricted.
==================== END ====================

==================== Testing dr self completion install (permissive execution policy) ====================
[INFO] Skipping [RemoteSigned]: unable to set execution policy to RemoteSigned.
==================== END ====================
```

ajalon1:
🧨 

```
==================== Testing completion generation ====================
[OK] Assertion passed: We have expected completion_powershell.ps1 file with Register-ArgumentCompleter.
==================== END ====================

==================== Testing dr self completion install (Restricted execution policy) ====================
[XFAIL] Assertion xfail [Restricted]: installer did not warn user with the execution policy fix command (expected: Phase 2 not yet merged)
==================== END ====================

==================== Testing dr self completion install (permissive execution policy) ====================
[OK] Assertion passed [RemoteSigned]: warning behavior correct
[OK] Assertion passed: profile contains completion block
[OK] Assertion passed [RemoteSigned]: execution policy remains 'RemoteSigned'
==================== END ====================
```

chasdr (approved):
👋 chas's claude again, and the xfail fall-through is sorted now, nice.

one thought on the Restricted case: it returns at the xfail before the profile-creation and policy asserts, so today it's a pure placeholder until Phase 2. but Restricted only blocks *loading* the profile, not writing it, so the installer still drops the profile there.

worth asserting profile-was-created under Restricted too, so Phase 1 gets real coverage from that case instead of just the RemoteSigned one? or are you saving all of it for #702?

ajalon1:
> or are you saving all of it for https://github.com/datarobot-oss/cli/pull/702?

definitely just saving it for #702. You already know I have that PR up and I'd rather push additional work over there. This one has sat around too long.

PR #609, opened 2026-06-26
[CFX-6668] DR CLI: Private CA support

asuslov:
## Summary

- Add `--skip-certificate-check` / `-k` persistent flag to disable TLS verification (insecure, for testing)
- Add `--ca-cert <path>` persistent flag to supply a PEM CA bundle; persisted to `drconfig.yaml`
- Add `--export-windows-certs` persistent flag (Windows only) to export the system certificate store to `%APPDATA%\DataRobot\ca-bundle.pem`
- New `internal/tls` package replaces `http.DefaultTransport` so all HTTP clients — including auth token validation — honour the custom TLS configuration

## Motivation

Customers with private certificate authorities hit an auth loop: DR CLI's token validation fails TLS → re-triggers login on every invocation. This gives them a path to either skip verification or supply their CA bundle.

## Test plan

- [ ] `dr --skip-certificate-check <subcommand>` against a self-signed cluster — no TLS error
- [ ] `dr --ca-cert /path/to/ca.pem <subcommand>` — custom CA trusted
- [ ] `ca-cert` value persists to `drconfig.yaml` after `dr auth login --ca-cert ...`
- [ ] `dr --export-windows-certs` on Windows — exports to `%APPDATA%\DataRobot\ca-bundle.pem`
- [ ] `dr --export-windows-certs` on macOS/Linux — returns clear "only supported on Windows" error

---

> [!NOTE]
> **High Risk**
> Mutates global `http.DefaultTransport` for every command, including auth, and exposes an explicit insecure skip-verify path; misconfiguration could weaken TLS for all CLI HTTP traffic.
> 
> **Overview**
> Adds **private CA / TLS controls** so outbound HTTPS (including auth token checks) can work behind corporate CAs or for testing.
> 
> New persistent flags: `--skip-certificate-check` / `-k`, `--ca-cert` (PEM bundle path, viper-bound and **persistable** in `drconfig.yaml`), and `--export-windows-certs` (Windows-only: dumps Root/CA stores to `%APPDATA%\DataRobot\ca-bundle.pem` and uses that path for the run). `setupTLS` runs in root `PersistentPreRunE` **after** config load so saved `ca-cert` applies globally.
> 
> New **`internal/tls`** clones `http.DefaultTransport` and either sets `InsecureSkipVerify`, loads a custom `RootCAs` pool from the PEM file, or leaves defaults unchanged. Non-Windows builds get a stub that errors on `--export-windows-certs`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit f96725472c52e696db230e267dec240a542ade15. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


cmd/root.go
```diff
@@ -188,6 +195,9 @@ func init() {
 	RootCmd.PersistentFlags().Duration("plugin-update-check-interval", internalPlugin.DefaultUpdateCheckInterval, "cooldown between plugin update checks (0s disables)")
 	RootCmd.PersistentFlags().Bool("skip-plugin-update-check", false, "skip plugin update checks before running plugins")
 	RootCmd.PersistentFlags().Bool("disable-telemetry", false, "disable anonymous usage telemetry")
+	RootCmd.PersistentFlags().BoolP("skip-certificate-check", "k", false, "skip TLS certificate verification (insecure)")
```
ajalon1: why `-k`?

internal/tls/tls_windows.go
ajalon1, line 1: Rename to tls_windows.go which is a Go convention.

internal/tls/tls_other.go
ajalon1, line 1: Rename to tls_other.go, following internal/telemetry pattern.

internal/tls/tls_other_test.go
ajalon1, line 1: tls_windows_test.go

internal/tls/windows.go
```diff
@@ -0,0 +1,63 @@
...
+		`$out = @()`,
+		`foreach ($store in 'Root','CA') {`,
+		`  foreach ($loc in 'LocalMachine','CurrentUser') {`,
+		`    try {`,
+		`      Get-ChildItem -Path "Cert:\$loc\$store" -ErrorAction SilentlyContinue | ForEach-Object {`,
+		`        $out += '-----BEGIN CERTIFICATE-----'`,
+		`        $out += [Convert]::ToBase64String($_.RawData, 'InsertLineBreaks')`,
+		`        $out += '-----END CERTIFICATE-----'`,
+		`      }`,
+		`    } catch {}`,
```
ajalon1: I don't like the error handling in the powershell script. :P

internal/telemetry/userid.go
```diff
@@ -92,7 +92,6 @@ func GetAccountInfo(_ context.Context) (*AccountInfo, error) {
 
 	var info AccountInfo
 
-	//nolint:contextcheck // GetJSON does not yet accept context; ctx is reserved for future use
 	if err := drapi.GetJSON(url, "", &info); err != nil {
```
ajalon1: Why remove?

cmd/root.go
```diff
@@ -188,6 +195,9 @@ func init() {
...
+	RootCmd.PersistentFlags().String("ca-cert", "", "path to a PEM-encoded CA certificate bundle")
+	RootCmd.PersistentFlags().Bool("export-windows-certs", false, "export Windows certificate store to the DataRobot CA bundle (Windows only)")
```
ajalon1: I can see the rationale for not making skip-certificate-check a flag persisted in drconfig.yaml or env var, but what about export-windows-cert?

cmd/root.go
```diff
@@ -291,6 +302,38 @@ func setUnknownArgGuards(root *cobra.Command) {
...
+		CACertPath: caCert,
+	})
+}
+
+// windowsCACertPath returns the well-known path where DR tooling writes the
+// exported Windows certificate store bundle.
+func windowsCACertPath() string {
+	appData := os.Getenv("APPDATA")
+
+	return filepath.Join(appData, "DataRobot", "ca-bundle.pem")
```
ajalon1: if APPDATA is nil like I'd expect it to be in linux or darwin, what does this become? is weird.

cmd/root.go
```diff
@@ -291,6 +302,38 @@ func setUnknownArgGuards(root *cobra.Command) {
...
+}
+
```
ajalon1: Would you consider moving these into cmd/tls.go? Just want to keep cmd/root clean.

ajalon1 (dismissed):
go a little bit further with the conditional builds :)

cmd/root.go
ajalon1: like we have `config.GetConfigDir()`, maybe that's a good place.

cmd/root.go
asuslov: The curl -k (or curl --insecure) command tells curl to bypass SSL/TLS certificate verification.

cmd/root.go
asuslov: hm. I thought of it as one time run to export explicitly

internal/telemetry/userid.go
asuslov: The directive on its own line is considered "unused" by nolintlint — contextcheck reports the issue on the if line itself, so the suppressor has to be on the same line. That's exactly why task delint removed it as stale in the first place. The inline  form is the correct placement 🤷

internal/tls/tls_windows.go
asuslov: done

internal/tls/tls_other.go
asuslov: done

internal/tls/tls_other_test.go
asuslov: `tls_other_test.go` right?

internal/tls/windows.go
asuslov: removed try/catch

cmd/root.go
asuslov: AppData moved to windows only

cmd/root.go
asuslov: done

internal/telemetry/userid.go
ajalon1: ahh, task delint. OK!

internal/tls/tls_other_test.go
ajalon1: oh, yeah

cmd/plugin/discovery.go
```diff
@@ -221,3 +243,22 @@ func performPluginUpdate(result *internalPlugin.UpdateCheckResult) {
...
+		case "--ca-cert":
+			if i+1 < len(args) && !strings.HasPrefix(args[i+1], "-") {
+				caCert = args[i+1]
+				i++
+			}
+		}
+	}
+
+	return
+}
```
ajalon1, line 278: Given a CLI plugin called `tlstest`, this misses a few valid inputs:
```
# Test
./dist/dr tlstest --ca-cert /nonexistent.pem       # error = parsed (works)
./dist/dr tlstest --ca-cert=/nonexistent.pem        # no error = NOT parsed (bug)
./dist/dr tlstest --ca-cert -nonexistent.pem        # no error = rejected (bug)
```


```suggestion
// scanTLSArgs extracts TLS flags from raw args passed to plugin commands.
// DisableFlagParsing: true means Cobra never processes persistent root flags,
// so -k/--skip-certificate-check and --ca-cert arrive here unprocessed.
// Uses pflag to correctly handle --flag=value syntax, -- terminators, and
// dash-leading paths. Unknown flags are whitelisted so plugin-specific args
// pass through without error.
func scanTLSArgs(args []string) (skipVerify bool, caCert string) {
    // NOTE This code and behavior is likely to change
	fs := pflag.NewFlagSet("tls-args", pflag.ContinueOnError)
	fs.ParseErrorsWhitelist = pflag.ParseErrorsWhitelist{UnknownFlags: true}

	fs.BoolP("skip-certificate-check", "k", false, "")
	fs.String("ca-cert", "", "")

	_ = fs.Parse(args)

	skipVerify, _ = fs.GetBool("skip-certificate-check")
	caCert, _ = fs.GetString("ca-cert")

	return
}
```

ajalon1 (commented):
You don't have to reinvent the wheel, spf13/pflag exists :)

This looks fine to me for now, I just want to be cautious about this as an overall pattern.

cmd/plugin/discovery.go
ajalon1: Also add docs for how this would work, just so we can have something to point Dell to it

ajalon1:
Going to take over this PR and clean it up, then get it merged in.

ajalon1 (approved):
LGTM @asuslov will merge after smoke tests pass.

I added a full feature gate here and fixed a handful of edge cases. Will ask Joao to test this.

PR #615, opened 2026-06-29
[CFX-6700] fix(plugin): let plugin processes respond to SIGINT/SIGKILL

ajalon1:
# RATIONALE

Plugin execution used a custom signal-forwarding goroutine to propagate signals to the child process. That approach did not integrate well with the `exec.Command` context lifecycle, so Ctrl-C or a timeout were not handled consistently.

I'm not sure the basis of the pattern for the original goroutine, because, but (as of [Go 1.20](https://go.dev/blog/go1.20#standard-library-additions)) the os.exec package allows us to define the behavior of an exec.Cmd when the associated context is cancelled or the associated process ends.

Reference: https://github.com/golang/go/issues/50436

**What that means is that we don't have to propagate signals ourselves, we can rely on the stdlib to do it for us.**

### Plugin execution now uses `exec.CommandContext`

- **Pass context through plugin execution** (`internal/plugin/exec.go`, `cmd/plugin/discovery.go`)
  - `ExecutePlugin` now accepts a `context.Context`.
  - `cmd/plugin/discovery.go` passes the Cobra command context (`pluginCmd.Context()`) into `ExecutePlugin`.
  - Authentication uses the same context, so cancelling during an auth prompt also cancels the plugin subprocess.

- **Remove custom signal goroutine** (`internal/plugin/exec.go`)
  - Deleted the manual `signal.Notify` goroutine that forwarded `SIGINT`/`SIGTERM` to `cmd.Process`.
  - Replaced it with `exec.CommandContext` and a `Cancel` hook that sends `SIGTERM`.
  - Added `cmd.WaitDelay = 5 * time.Second` so Go escalates to `SIGKILL` if the plugin ignores `SIGTERM`.
  - Windows `.ps1` plugins receive the same behavior via `powershell.exe`.

- **Modernize error handling** (`internal/plugin/exec.go`)
  - Replaced `errors.As(err, &exitErr)` with `errors.AsType[*exec.ExitError](err)` to avoid shadowing the stdlib helper. (This was changed in go 1.26)

### Tests exercise signal passthrough

- **Update existing tests for new signature** (`internal/plugin/exec_test.go`)
  - All `ExecutePlugin` call sites now pass `context.Background()`.

- **Add signal-handling tests** (`internal/plugin/exec_test.go`)
  - `TestExecutePluginContextCancellation` verifies that cancelling the context sends `SIGTERM` and the plugin can trap it.
  - `TestExecutePluginContextCancellationKillsUnresponsive` verifies that an unresponsive plugin is killed after the grace period.
  - Both tests skip on Windows because `SIGTERM` handling is Unix-specific.

#### BEFORE

Ctrl-C works for core but not plugin commands.

https://github.com/user-attachments/assets/07fcac0d-06af-4054-93f4-f77c77c526ff

#### AFTER

Ctrl-C works for both core and plugin commands.

https://github.com/user-attachments/assets/308f4a82-f917-41b8-8030-017a00369acf

## TODO

- use CancelCauses
- fix auth HTTP server not getting closed properly

## PR Automation

> [!IMPORTANT]
> - A maintainer uses `/skip-smoke-tests` to bypass the check without running smoke tests
>

---

> [!NOTE]
> **Medium Risk**
> Changes interrupt and teardown behavior for every plugin subprocess; the stdlib pattern is well-defined but mis-timed kills or auth cancellation could affect long-running plugins.
> 
> **Overview**
> **Plugin subprocess shutdown is wired through `context` instead of a custom signal goroutine**, so Ctrl-C and other cancellations behave like core CLI commands.
> 
> `ExecutePlugin` takes a `context.Context`; plugin discovery passes `pluginCmd.Context()`, and the same context drives auth when a plugin requires it. Subprocesses are started with `exec.CommandContext`, with `Cancel` sending **SIGTERM** and **`WaitDelay` of 5 seconds** before escalation to SIGKILL (including Windows `.ps1` via PowerShell). The old `signal.Notify` forwarder is removed. Exit errors are handled with `errors.AsType[*exec.ExitError]`.
> 
> Tests pass `context.Background()` everywhere and add Unix-only cases for graceful SIGTERM handling and killing plugins that ignore SIGTERM.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 9e28eff85c91f31c465318e854267471ad396bd8. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


c-h-russell-walker (approved):
Very nice - looks good to me.

PR #625, opened 2026-07-01
[APP-6487] Added universal arg handling and plugin arg passing

carsongee:
# RATIONALE
We're having an issue where we aren't passing universal arguments into plugins. We want to follow the established pattern of consuming them if they happen before the plugin is called, and then converting them into environment variables for plugins to decide if they want to consume them. Specifically for now, that is just two items --debug and --disable-telemetry. We'll want to make these a struct and then pass environment variables to the plugin with the prefix DATAROBOT_CLI_ so that when --debug is turned on we pass DATAROBOT_CLI_DEBUG=1. For the future, if that argument has a value like --cacert <path> we'd pass DATAROBOT_CLI_CACERT=<path> to the plugin. We need to clearly document this. Both for the plugin authors to consume, and for users of the CLI that they would need to pass dr --debug <plugin> <plugin args> to set debug for both the core functionality and the plugin (if it supports that flag).

### Goal
dr --debug <plugin> <args> and dr --disable-telemetry <plugin> <args> must:

Be consumed by core (enable core debug / disable telemetry).
NOT be passed to the plugin as literal args.
Be forwarded to the plugin subprocess as env vars with the DATAROBOT_CLI_ prefix: DATAROBOT_CLI_DEBUG=1, DATAROBOT_CLI_DISABLE_TELEMETRY=1.
Struct-driven and extensible: a future value flag --cacert <path> emits DATAROBOT_CLI_CACERT=<path>.

Hard invariant (non-negotiable)
Core stays BLIND to any args AFTER the plugin name (kubectl/helm model). Only flags BEFORE the plugin name are consumed by core; everything after passes to the plugin verbatim and is never parsed/validated by core.
### Decisions

Boolean universal flags emit value 1 (per the DATAROBOT_CLI_DEBUG=1 example). Struct is extensible so future value flags emit their actual value.
Only flags before the plugin name are consumed by core; args after the plugin name are never seen by core (kubectl/helm model). Regression-tested.
Mechanism: TraverseChildren: true on root — chosen over os.Args splitting.
Scope: only --debug and --disable-telemetry for now.

---

> [!NOTE]
> **Medium Risk**
> Changes global CLI flag parsing and plugin invocation contract; behavior is regression-tested but mis-placed flags (`dr myplugin --debug`) remain easy for users to get wrong despite docs.
> 
> **Overview**
> **Universal flags** (`--debug`, `--disable-telemetry`, and `--verbose`) can now affect both the core CLI and plugin subprocesses when placed **before** the plugin name (e.g. `dr --debug myplugin …`). Core consumes them for its own behavior and injects `DATAROBOT_CLI_DEBUG=1` (and peers) into the plugin environment via annotation-driven discovery—not as literal argv to the plugin.
> 
> Root **`TraverseChildren: true`** is added so Cobra parses persistent root flags left-to-right before descending into plugin commands; plugins keep **`DisableFlagParsing`**, so anything after the plugin name still passes through verbatim and core does not parse it (kubectl/helm model). Registration is centralized in **`bindUniversal`**, which binds viper and marks flags with `plugin-universal` for `universalFlagEnv` in `internal/plugin`.
> 
> Shared **`config.EnvPrefix`** and **`UniversalAnnotationKey`** replace scattered `DATAROBOT_CLI_` literals. Docs cover user/plugin author expectations; tests guard `TraverseChildren` and env forwarding. Dotenv model tests use the suite **`FinalModel`** helper only.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 3124498d60ef49662b693efe1f9a73776b16a149. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


internal/plugin/exec.go
```diff
@@ -136,6 +136,11 @@ func buildPluginCommand(ctx context.Context, executable string, args []string, r
 func buildPluginEnv(pluginPath string, requireAuth bool) []string {
 	env := os.Environ()
 
+	// Forward universal root flags (e.g. --debug, --disable-telemetry) as
+	// DATAROBOT_CLI_* env vars so plugins can optionally honour them.
+	// These override any inherited env vars of the same name.
+	env = append(env, universalFlagEnv()...)
```
ajalon1: I think this makes sense to add DATAROBOT_CLI_* env vars before any of the hardcoded ones later in this function.

carsongee: Isn't that what its doing already? It's the first one in the set, or are you suggesting I move it up in the call stack?

ajalon1: I was just agreeing with this approach! No changes needed.

cmd/dotenv/model_test.go
ajalon1, line 1: nice bit of cleanup

internal/plugin/universalenv.go
```diff
@@ -0,0 +1,75 @@
...
+package plugin
+
+import (
+	"github.com/datarobot/cli/internal/config/viperx"
+)
+
+// universalFlag describes a root-level CLI flag that is forwarded to plugins
+// as a DATAROBOT_CLI_<EnvSuffix> environment variable.
+//
+// Bool flags emit "<prefix>=1" when true and are omitted when false.
```
ajalon1: So basically it's a set / unset mechanism.

internal/plugin/universalenv.go
```diff
@@ -0,0 +1,75 @@
...
+
+// universalFlags is the canonical list of root CLI flags forwarded to plugins.
+// Only flags placed BEFORE the plugin name on the command line are consumed by
+// core; flags after the plugin name are never parsed by core (kubectl/helm model).
+var universalFlags = []universalFlag{
+	{ViperKey: "debug", EnvSuffix: "DEBUG", IsBool: true},
+	{ViperKey: "disable-telemetry", EnvSuffix: "DISABLE_TELEMETRY", IsBool: true},
+}
+
+const universalEnvPrefix = "DATAROBOT_CLI_"
```
ajalon1: We use this for everything, I think it's time to move this into a shared const in `internal/config`. There are at least three references.

We can DRY it later, though.

internal/plugin/universalenv.go
```diff
@@ -0,0 +1,75 @@
...
+	IsBool bool
+}
+
+// universalFlags is the canonical list of root CLI flags forwarded to plugins.
+// Only flags placed BEFORE the plugin name on the command line are consumed by
+// core; flags after the plugin name are never parsed by core (kubectl/helm model).
+var universalFlags = []universalFlag{
+	{ViperKey: "debug", EnvSuffix: "DEBUG", IsBool: true},
+	{ViperKey: "disable-telemetry", EnvSuffix: "DISABLE_TELEMETRY", IsBool: true},
+}
```
ajalon1: I think this mapping is overkill; you should only have to define the ViperKey.

ajalon1 (commented):
I Understand the whole idea, I just think there's too much extra cruft.

I don't really like the idea of having to update two files here, registering a flag in cmd/root.go and then marking it universal in universalenv.go.

What about using flag annotations? We already use those for telemetry and for feature gates, I think it could easily replace the universalFlag set.

Idea from Sonnet:

```go
// internal/plugin/universalenv.go
const universalEnvAnnotation = "dr:universal-env"

func tagUniversal(fs *pflag.FlagSet, name string) {
  suffix := strings.ToUpper(strings.ReplaceAll(name, "-", "_"))
  fs.Lookup(name).SetAnnotation(universalEnvAnnotation, []string{suffix})
}

func tagUniversalBool(fs *pflag.FlagSet, name string, value bool, usage string) *bool {
  p := fs.Bool(name, value, usage)
  tagUniversal(fs, name)
  return p
}
```

```go
// cmd/root.go
// in init(), replacing the two plain fs.Bool(...) calls
tagUniversalBool(RootCmd.PersistentFlags(), "debug", false, "debug output")
tagUniversalBool(RootCmd.PersistentFlags(), "disable-telemetry", false, "disable anonymous usage telemetry")
```

Then `internal/plugin/universalenv.go` walks the RootCmd.PersistentFlags() set looking for that annotation.

ETA, an example flag annotation would be, on `debug`, `dr:universal-env=DEBUG` which removes the need for specifically defining the env suffix. Of course, if we need to be able to listen to other env vars, that's a different thing, but that might be out of scope. It might be up to the plugin code to decide whether to handle DATAROBOT_CLI_* or some other env var.

internal/plugin/universalenv.go
ajalon1: you could set EnvSuffix and IsBool as additional flag annotations.

Actually, `flag.Value.Type()` could replace IsBool.

cmd/dotenv/model_test.go
carsongee: Thanks! Some test randomization broke with a flake so I had to fix it

cmd/root.go
```diff
@@ -170,6 +177,21 @@ func ExecuteContext(ctx context.Context) error {
...
+// registration point — adding a new universal flag only requires one call here.
+func bindUniversal(name string) {
+	flag := RootCmd.PersistentFlags().Lookup(name)
+	_ = viperx.BindPFlag(name, flag)
+
+	if flag.Annotations == nil {
+		flag.Annotations = map[string][]string{}
+	}
+
+	flag.Annotations[config.UniversalAnnotationKey] = []string{strings.ToUpper(strings.ReplaceAll(name, "-", "_"))}
```
carsongee, line 192: re:
>It might be up to the plugin code to decide whether to handle DATAROBOT_CLI_* or some other env var.

For future generations, my expectation/purpose for this is so that they could write `func bindUniversalAs(name string, suffix string)` and nothing would need to change anywhere else since we'll pull the variable suffix out of the annotation

ajalon1:
@bugbot review this

cmd/root.go
```diff
@@ -209,16 +231,19 @@ func init() {
 
 	outputformat.AddPersistentFlag(RootCmd.Command, &rootOutputFormat)
 
-	// Make some of these flags available via Viper
+	// Universal flags: bound to viper AND forwarded to plugin subprocesses as DATAROBOT_CLI_* env vars.
+	// To add a new universal flag, call bindUniversal here next to its registration above.
+	bindUniversal("debug")
+	bindUniversal("disable-telemetry")
+	bindUniversal("verbose")
```
ajalon1, line 238: nice

docs/commands/plugins.md
```diff
@@ -250,3 +250,32 @@ When no plugins are found, the command displays a message and the discovery loca
...
+```
+
+When a global flag is placed before the plugin name, the CLI forwards it to the
+plugin subprocess as a `DATAROBOT_CLI_*` environment variable in addition to
+consenting it internally:
+
+| CLI flag | Forwarded as env var | Value |
+|---|---|---|
+| `--debug` | `DATAROBOT_CLI_DEBUG` | `1` |
+| `--disable-telemetry` | `DATAROBOT_CLI_DISABLE_TELEMETRY` | `1` |
```
ajalon1, line 277: Noticed verbose isn't here. Feel free to add it, or just clarify this is a nonexhaustive list. (I'm getting kind of annoyed with these generated tables in our docs...)

internal/config/constants.go
```diff
@@ -19,4 +19,13 @@ const (
 	DataRobotAPIKey = "token"
 
 	APIConsumerTrackingEnabled = "api-consumer-tracking-enabled"
+
+	// EnvPrefix is the canonical prefix for all DATAROBOT_CLI_* environment
+	// variables. Use this constant instead of hard-coding the string literal.
+	EnvPrefix = "DATAROBOT_CLI_"
```
ajalon1, line 25: Thanks for tackling this

ajalon1 (approved):
:shipit:

docs/commands/plugins.md
carsongee: I feel ya. What about the other way, a scheduled agent job to update the docs once a week?

PR #627, opened 2026-07-06
[CFX-5345] Parallel plugin discovery

taras-pokornyy:
# RATIONALE

Some user reported that they were unable to run ‘dr assist' after ‘dr plugin install assist' was successfully run. However, this does NOT seem to be reproducible for all users though. 

**i did an install based on the [docs,](https://docs.datarobot.com/en/docs/api/agent-assist-cli/installation.html#install-and-run-agent-assist)**

- dr plugin install assist which succeed, then
- dr plugin list which shows that assist is installed, but when i attempt
- dr assist --help  but this returns an Error -> Error: unknown command "assist" for "dr"

**if i just run the sh script identify in dr plugin list, then it seems to run as expected**

## Summary

- Replace the all-or-nothing goroutine+select timeout in `RegisterPluginCommands` with
  context-based partial discovery. Previously, if discovery didn't finish within the
  deadline, **all** results were discarded — including managed plugins that had already
  completed. Now the deadline context is threaded through the entire stack so whatever
  finished before the cutoff is returned.
- Parallelize manifest fetching across PATH directories. Previously each PATH directory
  was scanned serially; now all run concurrently via `discoverPathDirsParallel`, with
  results merged in deterministic directory order (first directory wins on name conflicts).
- Consolidate the two discovery entry points (`discoverPlugins` + goroutine wrapper in
  `RegisterPluginCommands`) into a single `DiscoverPluginsWithContext(ctx)`. `GetPlugins`
  calls it with `context.Background()` (no deadline, cached); `RegisterPluginCommands`
  calls it with a deadline context.

## CHANGES

**`internal/plugin/discover.go`**
- Remove `discoverPlugins()` — replaced by `DiscoverPluginsWithContext(ctx context.Context)`
- Add `discoverPathDirsParallel` — fans out one goroutine per PATH directory using
  `sync.WaitGroup.Go`; each goroutine gets its own `maps.Clone(baseSeen)` to avoid map
  races; results merged in input order
- Add `getManifestsParallel` — fans out one goroutine per executable within a directory;
  goroutines check `ctx.Done()` before invoking the subprocess; indexed result slice
  preserves alphabetical order for deterministic dedup
- Thread `ctx` through `discoverInDir` and `getManifest`; `getManifest` now uses the
  passed ctx as the parent of the per-binary timeout so the outer deadline can cut it short

**`cmd/plugin/discovery.go`**
- Remove goroutine + `time.After` select block
- Create `context.WithTimeout` from the configured deadline and pass it to
  `DiscoverPluginsWithContext`

**`internal/plugin/discover_test.go`**
- Update all existing `discoverInDir` and `getManifest` call sites with `context.Background()`
- Add `TestDiscoverInDirCancelledContext` — already-cancelled ctx returns no plugins
- Add `PathDirsTestSuite` covering `discoverPathDirsParallel`:
  - Empty dirs
  - Multi-dir collection
  - Cross-dir dedup (first dir wins)
  - `baseSeen` filtering
  - `baseSeen` immutability (caller's map is never modified)
  - Cancelled context
  - Partial results — fast plugin is returned while slow plugin (`exec sleep 30`) is
    skipped after the deadline expires

## Behavior

| Scenario | Before | After |
|---|---|---|
| Timeout fires before any plugin responds | Returns nothing | Returns managed + local plugins (file I/O, no subprocess) |
| Timeout fires mid-PATH scan | Returns nothing | Returns plugins from PATH dirs that finished in time |
| No timeout | Serial PATH scan | Parallel PATH scan, same results |
| Duplicate name across PATH dirs | First dir wins (serial) | First dir wins (parallel, merged in order) |

## PR Automation

> [!IMPORTANT]
>

---

> [!NOTE]
> **Medium Risk**
> Changes CLI startup command registration and subprocess-heavy discovery concurrency; wrong timeout/dedup behavior could hide plugins or register duplicates, but behavior is heavily tested and scoped to discovery.
> 
> **Overview**
> Replaces the **all-or-nothing** plugin discovery timeout in `RegisterPluginCommands` with a single **`DiscoverPluginsWithContext`** deadline passed through discovery. When the cutoff hits, **managed/local plugins still register** (they finish before PATH work), and PATH scanning returns **whatever completed in time** instead of dropping everything.
> 
> PATH discovery is now **parallel**: one goroutine per PATH directory (`discoverPathDirsParallel`) and per `dr-*` binary within a dir (`getManifestsParallel`), with **per-goroutine `seen` clones** and merge in **directory / lexicographic order** so “first wins” dedup matches the old serial behavior. **`getManifest`** nests its per-binary timeout under the outer context so the global deadline can cancel slow subprocesses.
> 
> Tests add **`PathDirsTestSuite`** and cancellation/partial-result cases; existing calls pass **`context.Background()`**.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit c1d93b0767d616de9816db9287388624b82f208e. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


internal/plugin/discover.go
```diff
@@ -43,14 +45,17 @@ var registry = &DiscoveredPluginsRegistry{}
...
 func GetPlugins() ([]DiscoveredPlugin, error) {
 	registry.once.Do(func() {
-		registry.plugins, registry.err = discoverPlugins()
+		registry.plugins = DiscoverPluginsWithContext(context.Background())
 	})
 
 	return registry.plugins, registry.err
 }
 
-// TODO: Consider parallel manifest fetching using errgroup for better performance with many PATH directories
```
ajalon1, line 52: Was errgroup actually considered for this refactor? Was it a red herring?

internal/plugin/discover.go
```diff
@@ -68,28 +73,82 @@ func discoverPlugins() ([]DiscoveredPlugin, error) {
...
 
-	for _, dir := range filepath.SplitList(pathEnv) {
-		dirPlugins, errs := discoverInDir(dir, seen)
-		plugins = append(plugins, dirPlugins...)
+	plugins = append(plugins, discoverPathDirsParallel(ctx, filepath.SplitList(os.Getenv("PATH")), baseSeen)...)
 
-		for _, err := range errs {
-			log.Debug("Plugin discovery error", "dir", dir, "error", err)
-		}
+	log.Debug("Plugin discovery complete", "count", len(plugins))
```
ajalon1, line 97:
```suggestion
    if ctx.Err() != nil {
        log.Warn("Plugin discovery deadline exceeded...", ...) // or some similar warning
    }

	log.Debug("Plugin discovery complete", "count", len(plugins))
```

It's not clear to me right now if we even surface the fact taht plugin discovery failed due to timeout. Also, I think this a good place to point out that the user can deal with this by changing the value of `--plugin-discovery-timeout`

ajalon1 (approved):
LGTM, but I'd suggest getting @chasdr review since he authored most of this. :D

internal/plugin/discover.go
taras-pokornyy: Hey @ajalon1 !
Thank you for review and comments 🤝 
Almost 😅 
WaitGroup was used instead of errgroup 
https://medium.com/@pengcheng1222/golang-when-do-you-need-to-use-errgroup-rather-than-waitgroup-ad5e0779a2c3

I think it suites better in this case

chasdr (approved):
LGTM 👍 

minor observation: registry.err is never assigned now that DiscoverPluginsWithContext
doesn't return an error, so GetPlugins always returns (plugins, nil). could drop the error from the signature in a later cleanup.

internal/plugin/discover.go
taras-pokornyy: Added log.Warn in commit [4f60c3a](https://github.com/datarobot-oss/cli/pull/627/commits/4f60c3ae99844bb20bc48041c97d66a405a33de4)

taras-pokornyy:
> LGTM 👍
> 
> minor observation: registry.err is never assigned now that DiscoverPluginsWithContext doesn't return an error, so GetPlugins always returns (plugins, nil). could drop the error from the signature in a later cleanup.


Hey @chasdr ! 
Thank you for review and comment 🤝 
Really a good catch 🚀 

Looks like `discoverPlugins` was always returning `plugins, nil` even before current PR. 
Have changed the signature in commit [c9cae0a](https://github.com/datarobot-oss/cli/pull/627/commits/c9cae0aa6cc1f5a9a0c7859e019c5ecbd5dc3baa)

PR #628, opened 2026-07-06
[CFX-6839] Add plugin discovery fallback & make plugin exec respect --skip-auth

victorborshak:
# RATIONALE

- Make plugin discovery support `XDG_CONFIG_DIRS` var as a list of plugin locations
- make plugin exec flow respect `--skip-auth` falg

## PR Automation

> [!IMPORTANT]
>

---

> [!NOTE]
> **Medium Risk**
> Skipping auth on plugins that declare authentication can run without credentials or API env vars; discovery changes only affect which managed dirs are scanned.
> 
> **Overview**
> **Managed plugin discovery** now walks every directory returned by `ManagedPluginsDirs()` instead of a single managed path. When `XDG_CONFIG_HOME` is set, discovery checks that XDG plugins folder first and also **`~/.config/datarobot/plugins`** so installs from before XDG was set still show up; duplicate paths are deduped.
> 
> **Plugin execution** skips `auth.EnsureAuthenticated` when the global **`skip-auth`** viper flag is set, while still running the plugin subprocess (including auth-related env when the manifest requires it). Tests cover dual-directory discovery, path ordering/dedup, and skip-auth on/off behavior.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit f6bf0268e488e5a7a7f5448d0ba8ccfd5f9a7d8b. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


internal/plugin/discover.go
```diff
@@ -55,14 +55,17 @@ func discoverPlugins() ([]DiscoveredPlugin, error) {
...
-		plugins = append(plugins, managedPlugins...)
+		for _, managedDir := range managedDirs {
+			managedPlugins, errs := discoverManagedPlugins(managedDir, seen)
+			plugins = append(plugins, managedPlugins...)
 
-		for _, err := range errs {
-			log.Debug("Plugin discovery error in managed dir", "dir", managedDir, "error", err)
+			for _, err := range errs {
+				log.Debug("Plugin discovery error in managed dir", "dir", managedDir, "error", err)
+			}
```
ajalon1, line 68: This is an interesting decision, why? Are we actually running into scenarios where XDG_CONFIG_HOME is set to something else but folks are putting DR into `~/.config`?

ajalon1 (commented):
The skip-auth change makes sense to me, but I want to know more about plugin discovery fallback.

What scenarios does it cover? ~~Can we document EXACTLY what the flow is here?~~ 

I also wonder if we should start relying on an external package like https://github.com/adrg/xdg for config path discovery.

Users should also be able to understand why this might fail with `--debug`, though I think that is covered.

victorborshak:
> The skip-auth change makes sense to me, but I want to know more about plugin discovery fallback.
> 
> What scenarios does it cover? ~Can we document EXACTLY what the flow is here?~

In codespaces, there's no easy way to have pre-installed plugins because currently they are installed in the persistent volume (pointed to using the xdg var). So, in order to pre-install plugins using Dockerfiles, we have to install them in ~/.config and be able to discover them there.

On the other hand, if user installs a new version of the plugin in the codespace, we must install in the xdg location to have plugin persistence between codespace sessions.

> 
> I also wonder if we should start relying on an external package like https://github.com/adrg/xdg for config path discovery.

Yes, looks promising. But I don't want to do it in this PR to speed up things.

> 
> Users should also be able to understand why this might fail with `--debug`, though I think that is covered.

ajalon1:
The bugbot feedback is appropriate, so thanks for handling it.

---

> A cleaner primitive already exists for exactly this use case: XDG_CONFIG_DIRS (plural). The XDG Base Directory spec defines it as a preference-ordered list of additional base dirs to search for config, specifically for "system/image pre-seeded, read-only defaults, overridable by the user's own config home." That's precisely the codespaces scenario: bake plugins into a path exposed via XDG_CONFIG_DIRS at build time (not under the volume mount), and the CLI searches XDG_CONFIG_HOME first (writable, persistent, wins on conflict) then falls through XDG_CONFIG_DIRS entries. This is explicit, spec-compliant, generalizes to any future pre-seeding/air-gapped scenario, and doesn't silently reactivate for a random dev who sets XDG_CONFIG_HOME for unrelated reasons and happens to have old cruft in ~/.config.

> Where adrg/xdg fits: it already implements this correctly (xdg.ConfigDirs, xdg.SearchConfigFile, etc.), handles edge cases (empty env vars, Windows fallback semantics, dedup) that the PR's hand-rolled os.Getenv("XDG_CONFIG_HOME") + homedir-join logic doesn't bother with. It wouldn't magically solve the design question by itself (you still have to decide to route through ConfigDirs), but it removes the need to hand-roll and re-verify this logic in internal/config/config.go and the new paths.go, and it's a mature, widely-used library (not currently a dependency here). I'd consider it a solid implementation detail improvement, not a substitute for the design fix.

---

Could the Dockerfile point XDG_CONFIG_HOME at a different, non-mounted build-time path just for the install step (e.g. /opt/datarobot-xdg), then have the CLI (or an entrypoint script) treat that as an explicit extra search path?

---

I think we should look at XDG_CONFIG_DIRS and adrg/xdg, but given the time pressure, if you can't get that working then this is fine, since we always rewrite just ManagedPluginsDir(). Refactoring is cheap.

So approved -- if not addressed now, then we should address in a follow up very soon.

victorborshak:
> The bugbot feedback is appropriate, so thanks for handling it.
> 
> > A cleaner primitive already exists for exactly this use case: XDG_CONFIG_DIRS (plural). The XDG Base Directory spec defines it as a preference-ordered list of additional base dirs to search for config, specifically for "system/image pre-seeded, read-only defaults, overridable by the user's own config home." That's precisely the codespaces scenario: bake plugins into a path exposed via XDG_CONFIG_DIRS at build time (not under the volume mount), and the CLI searches XDG_CONFIG_HOME first (writable, persistent, wins on conflict) then falls through XDG_CONFIG_DIRS entries. This is explicit, spec-compliant, generalizes to any future pre-seeding/air-gapped scenario, and doesn't silently reactivate for a random dev who sets XDG_CONFIG_HOME for unrelated reasons and happens to have old cruft in ~/.config.
> 
> > Where adrg/xdg fits: it already implements this correctly (xdg.ConfigDirs, xdg.SearchConfigFile, etc.), handles edge cases (empty env vars, Windows fallback semantics, dedup) that the PR's hand-rolled os.Getenv("XDG_CONFIG_HOME") + homedir-join logic doesn't bother with. It wouldn't magically solve the design question by itself (you still have to decide to route through ConfigDirs), but it removes the need to hand-roll and re-verify this logic in internal/config/config.go and the new paths.go, and it's a mature, widely-used library (not currently a dependency here). I'd consider it a solid implementation detail improvement, not a substitute for the design fix.
> 
> Could the Dockerfile point XDG_CONFIG_HOME at a different, non-mounted build-time path just for the install step (e.g. /opt/datarobot-xdg), then have the CLI (or an entrypoint script) treat that as an explicit extra search path?
> 
> I think we should look at XDG_CONFIG_DIRS and adrg/xdg, but given the time pressure, if you can't get that working then this is fine, since we always rewrite just ManagedPluginsDir(). Refactoring is cheap.
> 
> So approved -- if not addressed now, then we should address in a follow up very soon.

@ajalon1  Yes, supporting XDG_CONFIG_DIRS instead of the fallback path sounds like a better idea. I updated the PR accordingly. Not using adrg/xdg just yet, but can do in the follow up PR if you don't mind.

ajalon1 (approved):
SGTM, approved on smoke tests passing

For further context, what I was trying to remember for our earlier discussions was how Jupyter specifically handles this. I figured their prior art would be reasonable to look at given how much Codespaces is built on that idea.

They have JUPYTER_CONFIG_PATH which is essentially XDG_CONFIG_DIRS. Sonnet tells me the history for jupyter_core/paths.py suggests that they ran into this same problem, but wrote it themselves. :)

PR #635, opened 2026-07-07
[CFX-6846] Fix flaky smoke-test apt-get update failure (MS repo)

ajalon1:
## Summary

The nightly smoke test occasionally fails at the "Install smoke-test dependencies" step because `packages.microsoft.com` (azure-cli / microsoft-prod apt sources, pre-seeded by GitHub's `ubuntu-latest` runner image) served a corrupted/unsigned `InRelease` file, which fails `apt-get update` outright.

See [run 28896705873](https://github.com/datarobot-oss/cli/actions/runs/28896705873/job/85722945755).
See [run 28900829395](https://github.com/datarobot-oss/cli/actions/runs/28900829395/job/85736940035).

We don't install anything from those Microsoft repos (for eaxample, azure-cli), so this removes the two pre-seeded source list files before `apt-get update` in both places that install `expect`/`bash-completion`: [.github/actions/install-deps/action.yaml](.github/actions/install-deps/action.yaml) and the duplicated inline step in [.github/workflows/pr-checks.yaml](.github/workflows/pr-checks.yaml).

Confirmed this is a transient issue: I checked the last 15 `nightly-smoke` runs — this was the only failure, bracketed by successful runs 1h13m before and 47m after with no code changes in between.

If a future job genuinely needs a Microsoft apt repo, it should add that repo + signing key explicitly for that job — independent of these two unused pre-baked files. Alternately, we can remove these changes.

## Test plan
- [x] `task lint` passes
- [ ] CI: confirm `pr-checks.yaml`'s "Install dependencies" step succeeds
- [ ] Manually trigger `nightly-smoke` / `smoke-on-demand` after merge and confirm the "Install smoke-test dependencies" step no longer depends on Microsoft's mirror

🤖 Generated with [Claude Code](https://claude.com/claude-code)

---

> [!NOTE]
> **Low Risk**
> CI-only apt setup change; no application code, auth, or data paths affected.
> 
> **Overview**
> Addresses sporadic CI failures where **`apt-get update`** dies on GitHub **`ubuntu-latest`** runners because pre-seeded **`packages.microsoft.com`** sources (`azure-cli.list`, `microsoft-prod.list`) sometimes return a bad **`InRelease`**.
> 
> Before **`apt-get update`**, both Linux dependency install paths now **`sudo rm -f`** those two list files (with comments that the repo does not use those repos and any future MS package should be added explicitly per job): the **`install-deps`** composite action and the inline **Install dependencies** step in **`pr-checks.yaml`** completion tests.
> 
[...]

PR #637, opened 2026-07-07
[CFX-6846] Document composite-action @main-pinning testing gap

ajalon1:
## RATIONALE

While fixing the flaky smoke-test apt-get issue (#635, #636), we found that composite actions under `.github/actions/` are referenced by `_smoke.yaml`/`_pre-release-smoke.yaml` via a pinned `@main` ref rather than the caller's ref. This means a PR that edits a composite action's `action.yaml` doesn't exercise that change in its own CI — it silently tests against `main`'s existing version, so a broken fix can merge without ever being validated.

## CHANGES

Adds a new informational bugbot rule, [.cursor/bugbot-ci.md](.cursor/bugbot-ci.md), and wires it into [.cursor/BUGBOT.md](.cursor/BUGBOT.md) under a new "CI & Workflow Rules" section, so reviewers flag this as a callout (not a correctness bug) on PRs touching `.github/actions/*/action.yaml`.

Adds the same guidance to [AGENTS.md](AGENTS.md) so contributors know upfront that they'll need to verify out-of-band (temporary ref override, or a manual `workflow_dispatch` post-merge).

### DIFF
--- a/.cursor/BUGBOT.md
+++ b/.cursor/BUGBOT.md
@@ -30,3 +30,9 @@ Ensure consistency and maintainability.
 
 - **bugbot-testing.md** — Race detector, error paths, test seams, mocking
 - **bugbot-cmd.md** — Table rendering, file organization, output consistency
+
+## CI & Workflow Rules
+
+Informational callouts about GitHub Actions workflows and composite actions.
+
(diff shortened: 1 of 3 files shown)


cdevent (approved):
Thank you! good point to document this

PR #668, opened 2026-07-14
[CFX-6839] Use adrg/xdg for XDG directories resolution

victorborshak:
# RATIONALE

Using [adrg/xdg](https://github.com/adrg/xdg) package instead of manual resolution for XDG directories.

## PR Automation

> [!IMPORTANT]
>

---

> [!NOTE]
> **Medium Risk**
> Changes where config, plugins, dotenv backups, and telemetry caches are stored on disk; behavior is intentionally preserved for unset XDG vars but differs if callers relied on implicit OS XDG defaults when env vars were empty.
> 
> **Overview**
> **Centralizes XDG base-directory resolution** on [`github.com/adrg/xdg`](https://github.com/adrg/xdg) instead of ad-hoc `os.Getenv` / home-dir joins scattered across the CLI.
> 
> `internal/config` now exposes **`GetStateDir`**, **`GetConfigDirs`**, and refactored **`GetConfigDir`** via `resolveConfigHome`: explicit `XDG_*` vars use `xdg`’s parsed values; when unset, config/state still fall back to **`~/.config`** and **`~/.local/state`** (not OS-specific XDG defaults). **`XDG_CONFIG_DIRS`** is **opt-in only**—`GetConfigDirs` returns nil unless the env var is set, and plugin search paths use that helper instead of manual list splitting.
> 
> **`dr dotenv`** backup state under `getStateDir` now calls **`config.GetStateDir`** (replacing inline `XDG_STATE_HOME` logic).
> 
> Tests across dotenv, plugins, telemetry, and config switch to **`testutil.SetXDGEnv`**, which sets the env var and calls **`xdg.Reload()`** so `adrg/xdg`’s cache matches per-test overrides.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 89f6a2f143c9f4e9b26b1ab1041e1c438212f482. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/internal/testutil/home.go
+++ b/internal/testutil/home.go
@@ -14,7 +14,11 @@
 
 package testutil
 
-import "testing"
+import (
+	"testing"
+
+	"github.com/adrg/xdg"
+)
 
 // SetTestHomeDir sets the home directory for tests to work cross-platform.
 // Both HOME (Unix) and USERPROFILE (Windows) are set so os.UserHomeDir() works everywhere.
@@ -23,5 +27,16 @@ func SetTestHomeDir(t *testing.T, dir string) {
 	t.Helper()
 	t.Setenv("HOME", dir)
 	t.Setenv("USERPROFILE", dir)
-	t.Setenv("XDG_CONFIG_HOME", "")
+	SetXDGEnv(t, "XDG_CONFIG_HOME", "")
+}
+
+// SetXDGEnv sets an XDG_* environment variable for the duration of the test
+// and reloads github.com/adrg/xdg's cached base directories so the change
+// takes effect immediately. Without this, xdg.ConfigHome/xdg.StateHome/
+// xdg.ConfigDirs would keep reflecting whatever the environment looked like
+// at process start (or the last Reload), ignoring the env var set here.
+func SetXDGEnv(t *testing.T, key, value string) {
+	t.Helper()
+	t.Setenv(key, value)
+	xdg.Reload()
 }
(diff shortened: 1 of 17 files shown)


internal/testutil/home.go
```diff
@@ -23,5 +27,16 @@ func SetTestHomeDir(t *testing.T, dir string) {
...
+
+// SetXDGEnv sets an XDG_* environment variable for the duration of the test
+// and reloads github.com/adrg/xdg's cached base directories so the change
+// takes effect immediately. Without this, xdg.ConfigHome/xdg.StateHome/
+// xdg.ConfigDirs would keep reflecting whatever the environment looked like
+// at process start (or the last Reload), ignoring the env var set here.
+func SetXDGEnv(t *testing.T, key, value string) {
+	t.Helper()
+	t.Setenv(key, value)
+	xdg.Reload()
```
chasdr, line 41: small thing on SetXDGEnv: t.Setenv restores the env var on cleanup but nothing re-runs xdg.Reload() after, so the cache keeps the test's value once the test ends. harmless today since CI runs with XDG unset, but if XDG_CONFIG_HOME is set in the ambient env a later test that reads config without SetXDGEnv would get the stale dir. worth a t.Cleanup(func(){ xdg.Reload() }) registered before t.Setenv so it runs after the env restore?

chasdr (approved):
lgtm. build/vet/lint clean and the config/plugin/telemetry/dotenv tests pass locally. one non-blocking nit on the test helper below.

PR #669, opened 2026-07-14
[CFX-6931] Handle DR prerquisites

taras-pokornyy:
# RATIONALE

`Prerequisite.Command` entries that invoke `dr` itself (e.g. a plugin declaring a dependency on another installed `dr` plugin via `dr plugin version <name>`, or the CLI's own self-version check) need different handling than a regular system-binary prerequisite: `dr` is always on `PATH` (it's the running binary), so `exec.LookPath("dr")` can't tell you whether a *named plugin* is actually installed, and the previous code had dead/broken branches trying to special-case this inside `isInstalled`/`isVersionInstalled` (gated on the wrong field, `tool.Key` instead of the command shape).

This change centralizes all `dr`-command handling into one place, checked *before* the generic `isInstalled` path, and removes several long-unused, already-commented-out helpers (`CheckPrerequisite`, `CheckTool`, `GetMissingTools`) that had no remaining callers.

## CHANGES

- `CheckPrerequisiteList` now parses `tool.Command` up front and branches on `command == "dr"` before falling into the generic `isInstalled`/`isVersionInstalled` path.
- Added `checkDrPrerequisite`, which handles both `dr`-command shapes:
  - `dr plugin version <name>` — dependency on another installed `dr` plugin. Runs the command once (`exec.Command(...).Output()`) and checks both "is it installed" and "does it satisfy `MinimumVersion`" from the same invocation.
  - Any other `dr ...` command (e.g. the CLI's own self-version entry) — falls back to `SufficientSelfVersion`.
- `isInstalled` and `isVersionInstalled` reverted to plain, generic implementations with no `dr`-specific branching.
- Removed dead code: `CheckPrerequisite`, `CheckTool`, `GetMissingTools`.

## PR Automation

> [!IMPORTANT]
>

---

> [!NOTE]
> **Medium Risk**
> Changes how dependency checks run for plugin-declared prerequisites (subprocess `dr` invocations) on the quickstart/deps path; behavior is improved over broken prior logic and is heavily tested, but mistakes could block or mis-report installs.
> 
> **Overview**
> **Prerequisite checks now treat `dr`-invoking commands separately** before the normal PATH/version path, so plugin dependencies and CLI self-version are validated correctly instead of assuming `dr` on PATH means every plugin is installed.
> 
> `CheckPrerequisiteList` parses each command and calls new **`checkDrPrerequisite`** when the binary is `dr`: for **`dr plugin version <name>`** it runs that command once and classifies missing vs wrong version from the output; for other `dr` commands (e.g. **`dr self version`**) it uses **`SufficientSelfVersion`**. **`isInstalled`** / **`isVersionInstalled`** no longer special-case `dr` (including the broken `tool.Key == "dr"` branch). Unused helpers **`CheckPrerequisite`**, **`CheckTool`**, and **`GetMissingTools`** are removed.
> 
> Tests add fake-`dr` coverage for plugin installed/missing/wrong version and self-version, plus **`sufficientVersion`** cases for SNAPSHOT strings and multi-line plugin output with a header line.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 78cb1317991feef62174ad004adb6608b288985e. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/internal/tools/prerequisites.go
+++ b/internal/tools/prerequisites.go
@@ -100,18 +100,6 @@ var RequiredTools = []Prerequisite{
 	{Name: "pulumi", Command: "pulumi version", URL: "https://www.pulumi.com/docs/get-started/download-install/", MinimumVersion: "3.245.0", Install: pulumiInstallCmd},
 }
 
-func CheckPrerequisite(name string) error {
-	for _, tool := range RequiredTools {
-		if tool.Name == name {
-			if !isInstalled(tool.Command) {
-				return fmt.Errorf("%s is not installed.", name)
-			}
-		}
-	}
-
-	return nil
-}
-
 // CheckResult holds the outcome of a CheckPrerequisites call.
 type CheckResult struct {
 	MissingTools         []Prerequisite
@@ -153,7 +141,11 @@ func CheckPrerequisiteList(prereqs []Prerequisite) CheckResult {
 	var result CheckResult
 
 	for _, tool := range prereqs {
-		if !isInstalled(tool.Command) {
+		command, args := commandArgs(tool.Command)
+
+		if command == "dr" {
+			checkDrPrerequisite(&result, tool, command, args)
+		} else if !isInstalled(tool.Command) {
 			log.Debug("deps: tool missing", "name", tool.Name)
 
 			result.MissingTools = append(result.MissingTools, tool)
@@ -169,6 +161,49 @@ func CheckPrerequisiteList(prereqs []Prerequisite) CheckResult {
 	}
 
 	return result
+}
+
+// checkDrPrerequisite handles a Prerequisite whose command invokes the dr CLI itself:
+// either "dr plugin version <name>" (a dependency on another installed dr plugin) or
+// the CLI's own self-version entry (e.g. "dr self version").
+func checkDrPrerequisite(result *CheckResult, tool Prerequisite, command string, args []string) {
+	if len(args) > 2 && args[0] == "plugin" && args[1] == "version" {
+		versionOutput, err := exec.Command(command, args...).Output()
+		if err != nil {
+			log.Debug("deps: plugin dependency missing", "name", tool.Name)
+
+			result.MissingTools = append(result.MissingTools, tool)
+			result.MissingMsgs = append(result.MissingMsgs, fmt.Sprintf("%s %s (%s)", tool.Name, tool.MinimumVersion, tool.URL))
+
+			return
+		}
+
+		if versionInstalled, ok := sufficientVersion(string(versionOutput), tool.MinimumVersion); !ok {
+			log.Debug("deps: plugin dependency wrong version", "name", tool.Name, "msg", versionInstalled)
+
+			result.WrongVersionTools = append(result.WrongVersionTools, tool)
+			result.WrongVersionMsgs = append(result.WrongVersionMsgs, fmt.Sprintf("%s (minimal: v%s, installed: %s)\n%s\n",
+				tool.Name, tool.MinimumVersion, versionInstalled, tool.URL))
+
+			return
+		}
+
+		log.Debug("deps: plugin dependency ok", "name", tool.Name)
+
+		return
+	}
+
+	if !SufficientSelfVersion(tool.MinimumVersion) {
+		log.Debug("deps: tool wrong version", "name", tool.Name)
+
+		result.WrongVersionTools = append(result.WrongVersionTools, tool)
+		result.WrongVersionMsgs = append(result.WrongVersionMsgs, fmt.Sprintf("%s (minimal: v%s, installed: %s)\n%s\n",
+			tool.Name, tool.MinimumVersion, version.Version, tool.URL))
+
+		return
+	}
+
+	log.Debug("deps: tool ok", "name", tool.Name)
 }
 
 // PrerequisitesMsg formats the message to display to the user about missing/wrong-version prerequisites.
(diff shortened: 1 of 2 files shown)


internal/tools/prerequisites_test.go
```diff
@@ -70,6 +70,13 @@ func TestSufficientVersionTrue(t *testing.T) {
 		{installed: "v0.2.55-beta.0-0-gabcd1234", minimal: "0.2.55"},
 		// Fallback format for fresh clones
 		{installed: "v0.0.0-dev.42.gabcd1234", minimal: "0.0.0"},
+		// SNAPSHOT format emitted by Java / Maven-style plugins
+		{installed: "1.3.1-SNAPSHOT-b3cdbb6", minimal: "1.3.0"},
+		{installed: "1.3.1-SNAPSHOT-b3cdbb6", minimal: "1.3.1"},
+		// Plugin-style output: CLI wrapper prints a header line before the version.
+		// sufficientVersion must skip the non-numeric header and parse the second line.
+		{installed: "🔌 Running plugin: xp\n1.3.1-SNAPSHOT-b3cdbb6", minimal: "1.3.0"},
+		{installed: "🔌 Running plugin: xp\n1.3.1-SNAPSHOT-b3cdbb6", minimal: "1.3.1"},
```
ajalon1, line 79: thanks for test updates, I'm happy to see real version numbers that are possible :)

internal/tools/prerequisites_test.go
```diff
@@ -487,3 +498,133 @@ func TestCheckPrerequisiteList_NeverSetsValidationViolations(t *testing.T) {
...
+	require.Len(t, result.WrongVersionTools, 1)
+	assert.Equal(t, "myplugin", result.WrongVersionTools[0].Name)
+	require.Len(t, result.WrongVersionMsgs, 1)
+	assert.Contains(t, result.WrongVersionMsgs[0], "myplugin")
+}
+
+// The dr-self-version branch (checkDrPrerequisite falling through to
+// SufficientSelfVersion) never shells out — it only compares the version.Version
+// package var — so these tests don't need a fake "dr" binary on PATH at all.
+
```
ajalon1:
```suggestion
```

internal/tools/prerequisites.go
```diff
@@ -153,7 +141,11 @@ func CheckPrerequisiteList(prereqs []Prerequisite) CheckResult {
 	var result CheckResult
 
 	for _, tool := range prereqs {
-		if !isInstalled(tool.Command) {
+		command, args := commandArgs(tool.Command)
+
+		if command == "dr" {
```
ajalon1: We already have a constant for this in `internal/version/version.CliName`.

I wonder if we should also handle the possibility that a plugin defines a dependency on `datarobot` and not `dr`, in which case we could compare the command name versus CliName and CliAliases.

internal/tools/prerequisites.go
```diff
@@ -224,15 +255,6 @@ func isVersionInstalled(tool Prerequisite) (string, bool) {
 		return "", true
 	}
 
-	if tool.Key == "dr" {
```
ajalon1, line 227: Now I am glad that we've narrowed the use of this constant down to one place.

internal/tools/prerequisites.go
```diff
@@ -171,6 +163,49 @@ func CheckPrerequisiteList(prereqs []Prerequisite) CheckResult {
...
 }
 
+// checkDrPrerequisite handles a Prerequisite whose command invokes the dr CLI itself:
+// either "dr plugin version <name>" (a dependency on another installed dr plugin) or
+// the CLI's own self-version entry (e.g. "dr self version").
+func checkDrPrerequisite(result *CheckResult, tool Prerequisite, command string, args []string) {
+	if len(args) > 2 && args[0] == "plugin" && args[1] == "version" {
+		versionOutput, err := exec.Command(command, args...).Output()
+		if err != nil {
+			log.Debug("deps: plugin dependency missing", "name", tool.Name)
```
ajalon1: might be worth adding "err" as an extra.

ajalon1 (approved):
LGTM, but I'd suggest `command == "dr"` be refactored to check against version.CLIName and version.CLIAliases.

Thanks for cleaning this up, it had gotten out of hand. Thanks also for logging updates.

internal/tools/prerequisites.go
taras-pokornyy: Yeah! 
It was bothering me too ...

internal/tools/prerequisites.go
taras-pokornyy: Done in commit : [feab825](https://github.com/datarobot-oss/cli/pull/669/commits/feab82591cd0873a1390a79800412b1f3b77bf87)

internal/tools/prerequisites.go
taras-pokornyy: Hey, @ajalon1 
It is really a great catch! 
Thank you for pointing at that 🤝 

Fixed in commit : [feab825](https://github.com/datarobot-oss/cli/pull/669/commits/feab82591cd0873a1390a79800412b1f3b77bf87)

internal/tools/prerequisites_test.go
taras-pokornyy: 😅 
removed i commit : [feab825](https://github.com/datarobot-oss/cli/pull/669/commits/feab82591cd0873a1390a79800412b1f3b77bf87)

PR #684, opened 2026-07-20
[CFX-6907] Fix frozen wizard spinners and unify on a shared tui.Loading component

cdevent:
# RATIONALE

Every TUI flow (`RunWithSpinner`, `setup.Model`, `clone.Model`, `promptModel`, `AddModel`, `UpdateModel`) hand-rolled its own `spinner.Model` field with duplicated Init/Update/View boilerplate and inconsistent glyphs (`Dot` vs `Points`).

That duplication was hiding a real bug: the "Fetching available LLMs from DataRobot..." spinner looked frozen after its first frame whenever the dotenv wizard ran nested inside `dr templates setup` / `dr start` (it worked fine standalone via `dr dotenv setup`).

**Root cause:** `setup.Model.Update()`'s top-level switch intercepted every `spinner.TickMsg` by type alone and returned before the per-screen dispatch that forwards messages to the active child screen (`m.clone`, `m.dotenv` → nested `promptModel`). Since `bubbles/spinner` stamps ticks with the emitting spinner's own ID and no-ops on mismatch, the child's own tick was rejected by the parent's spinner and the function returned immediately — the child's spinner never got rescheduled and froze after one frame.

Once ticks started flowing correctly to the clone screen too, a second issue surfaced: the clone screen showed **two** spinners at once (its own inline one plus the parent wizard's status bar, both now animating). Fixed by dropping the redundant one — the parent's status bar is the single source of truth for that loading state.

## CHANGES

- Fixes `setup.Model.Update()` so `spinner.TickMsg` (and every other message) always reaches the per-screen dispatch instead of being swallowed by the parent's own spinner handling.
- Removes the duplicate spinner render on the clone screen.
- Extracts `tui.Loading`, a single embeddable spinner sub-model, and migrates all hand-rolled spinners onto it (standardizing on the `Dot` glyph, including `RunWithSpinner` which previously used `Points`).
- Adds a regression test proving a `spinner.TickMsg` sent while the wizard is on the clone screen reaches the child's `Update` (verified it fails against the pre-fix code and passes against the fix).
- Adds `tui/loading_test.go` for the new shared component.

https://github.com/user-attachments/assets/3e90145a-32ff-4c1c-97bf-b971202413e3

## TESTING

- `task lint` — clean.
- `task test` — full suite, race detector, clean; also re-ran with `-shuffle=on` repeatedly to confirm no flakiness.
- Manually reasoned through every screen with both a parent status bar and a child spinner (`hostScreen`, `loginScreen`, `listScreen`, `dotenvScreen`, `cloneScreen`) to confirm the clone screen was the only case of double-rendering.

Co-Authored-By: Claude Sonnet 5 <noreply@anthropic.com>

---

> [!NOTE]
> **Low Risk**
> Changes are confined to CLI TUI loading UX and message routing; no auth, API, or data-path changes, with regression tests for the main bug.
> 
> **Overview**
> Fixes **frozen loading spinners** when the dotenv wizard (e.g. “Fetching available LLMs…”) or clone flow runs inside `dr templates setup` / `dr start`. The setup wizard no longer handles `spinner.TickMsg` in a top-level switch that returned before child screens; it updates its own spinner and **forwards every message** to the active screen, batching spinner and child commands.
> 
> Introduces **`tui.Loading`** as the standard embeddable spinner (`Dot` + `InfoStyle`) and migrates component add/update, dotenv prompts, template clone/setup, and `RunWithSpinner` off duplicated `spinner.Model` setup. Status bars that need the raw bubble use `m.spinner.Spinner`. During clone, the **inline duplicate spinner** is removed so only the wizard status bar animates.
> 
> Adds **`tui/loading_test.go`** and a **setup regression test** that spinner ticks reach `m.clone.Update` while cloning.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit cecfa770027077c73f77e927fabbafad69711e53. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/cmd/component/shared/addModel.go
+++ b/cmd/component/shared/addModel.go
@@ -47,7 +47,7 @@ const (
 type AddModel struct {
 	screen   addScreens
 	list     list.Model
-	spinner  spinner.Model
+	spinner  tui.Loading
 	width    int
 	height   int
 	errorMsg string
@@ -55,13 +55,9 @@ type AddModel struct {
 }
 
 func NewAddModel() AddModel {
-	s := spinner.New()
-	s.Spinner = spinner.Dot
-	s.Style = tui.InfoStyle
-
 	return AddModel{
 		screen:  addLoadingScreen,
-		spinner: s,
+		spinner: tui.NewLoading(),
 	}
 }
 
@@ -156,7 +152,7 @@ func (am AddModel) loadComponents() tea.Cmd {
 }
 
 func (am AddModel) Init() tea.Cmd {
-	return tea.Batch(am.loadComponents(), am.spinner.Tick, tea.WindowSize())
+	return tea.Batch(am.loadComponents(), am.spinner.Init(), tea.WindowSize())
 }
 
 func (am AddModel) Update(msg tea.Msg) (tea.Model, tea.Cmd) { //nolint: cyclop
--- a/cmd/component/shared/updateModel.go
+++ b/cmd/component/shared/updateModel.go
@@ -109,7 +109,7 @@ type UpdateModel struct {
 	viewport    viewport.Model
 	help        help.Model
 	keys        detailKeyMap
-	spinner     spinner.Model
+	spinner     tui.Loading
 	updating    bool
 	ready       bool
 	ExitMessage string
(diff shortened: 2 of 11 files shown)


cdevent:
@akshoop Please take a look a demo in preview, how do you think, is it works or should we jut switch/search to any other kind of the spinners?  🙏

chasdr (approved):
Code changes look good, the animation in the description looks good too, but is subjective i guess :)

PR #688, opened 2026-07-21
[CFX-6922] cancelReader on Windows

taras-pokornyy:
# RATIONALE

`internal/misc/reader.ReadString()` (used by every `y/N` prompt and free-text prompt in the CLI, e.g. `dr auth set-url`) had a Windows-specific escape hatch: `if runtime.GOOS == "windows" { return bufio.NewReader(os.Stdin).ReadString('\n') }`.
That bypassed `cancelreader` entirely on Windows, which meant:

- Ctrl+C never canceled a pending prompt on Windows.
- Console echo/backspace handling was inconsistent, since Windows raw  console mode disables `ENABLE_ECHO_INPUT` and nothing was replacing it.
- Arrow keys / Home / End sent raw VT100 escape bytes (e.g. `ESC [ C`) that got appended straight into the answer, corrupting typed input (a URL like `abcdefghi\x1b[C\x1b[C` failing `url.Parse`).

Removing that bypass and routing Windows through `cancelreader` properly surfaced a second, unrelated problem: `cancelreader`'s Windows implementation special-cases stdin — if the reader you hand it has the same fd as the real `os.Stdin`, it silently discards your reader and opens the real console handle (`CONIN$`) instead. Our tests simulate input by reassigning the `os.Stdin` global to a pipe, which on Windows made `cancelreader` ignore that pipe and block on `WaitForMultipleObjects` forever, waiting for real keystrokes that never come on a headless CI runner — surfacing as a 10-minute test timeout panic in `cmd/plugin/install` and, separately, in `internal/auth` (`TestEnsureAuthenticated_NoURL`, which drives the interactive "choose your DataRobot URL" prompt with no stdin stubbed at all).

This PR is the accumulated fix for both the original garbled-input bug and the Windows CI hangs it exposed along the way.

## CHANGES
- **Rewrote line reading in `internal/misc/reader/reader.go`** to work correctly through `cancelreader` on every platform instead of bypassing it on Windows:
  - Reads byte-by-byte via a new `readLine()`, discarding Ctrl+C and VT100  escape sequences (arrow keys, Home/End, etc.) instead of appending them  to the answer.
  - Handles Windows raw console mode manually: echoes typed characters and  erases them on backspace, since `cancelreader` disables `ENABLE_ECHO_INPUT` on Windows and the console no longer does this for us.
  - Accepts a bare `\r` as a line ending (not just `\n`), since Windows raw console mode delivers Enter as a lone `\r` with `ENABLE_LINE_INPUT` off.
- **Fixed a `cancelreader` lifecycle bug**: `cr.Close()` was never called, leaking epoll/pipe fds on Linux/macOS and leaving the Windows console stuck in raw mode (no echo) after every prompt. Added the missing `defer cr.Close()`, and guarded the cancel goroutine so a normal (non-canceled) read doesn't also call `cr.Cancel()` on completion — `cancelreader` has no synchronization between `Cancel()` and `Close()`, and racing them made Windows skip console-mode restoration.
- **Added an injectable `reader.Stdin` seam** so tests can supply canned input without reassigning the global `os.Stdin`. Reassigning `os.Stdin` itself doesn't work on Windows: `cancelreader` compares the given reader's fd against `os.Stdin`'s fd and, when a test points `os.Stdin` at a pipe, the fds trivially match and `cancelreader` throws the pipe away in favor of reading from the real console — hanging forever on a headless runner. Overriding `reader.Stdin` instead leaves the real `os.Stdin` untouched, so `cancelreader` correctly falls back to reading from the injected pipe. 
- **Guarded `auth.SetURLAction()` behind `reader.IsStdinTerminal()`**, matching the non-interactive convention already used elsewhere in the CLI (`cmd/artifact/del`, `cmd/workload/del`, `cmd/plugin/discovery.go`). Without a terminal attached, the interactive "choose your DataRobot URL" prompt now fails fast with an actionable message instead of blocking indefinitely — fixing both the Windows CI hang and the underlying real-world case of `dr` running headless (CI, Docker, scripts) with no endpoint configured. Users running non-interactively already have `dr auth set-url <url>` or `DATAROBOT_ENDPOINT` for this.
- Also fixed a matching bug in the same function: the "no URL entered" break condition checked `url == "\n"`, which `readLine()` never returns (it strips the line ending); corrected to `url == ""`.
- Added `internal/misc/reader/reader_test.go` coverage for `readLine()`:  Unix `\n`, bare `\r`, `\r\n`, multiple sequential reads, EOF with and without a trailing newline, and escape-sequence handling (arrow keys mid-line and a bare Escape keypress that isn't part of a CSI sequence).

## PR Automation

> [!IMPORTANT]
>

---

> [!NOTE]
> **Medium Risk**
> Touches all interactive CLI prompts via `ReadString`; fixes are localized but incorrect terminal handling would affect every stdin prompt on Windows and cancellation paths on all platforms.
> 
> **Overview**
> **Fixes interactive stdin on Windows** by using `cancelreader` everywhere again (reverting the CFX-4799 bypass) and addressing the real lifecycle bugs: **`defer cr.Close()`** so console mode is restored and fds are not leaked, and **only calling `cr.Cancel()` when a real SIGINT/SIGTERM was received** so a successful read does not race `Close()` and leave the terminal in raw mode.
> 
> **Replaces `bufio.Reader.ReadString('\n')` with a custom `readLine` path** tailored to `cancelreader`’s raw Windows console: treat `\r` or `\n` as end-of-line without CRLF lookahead (avoids hangs), **manual echo and backspace** on Windows only, map **Ctrl+C** to `cancelreader.ErrCanceled`, and **strip VT100/ANSI escape sequences** so arrow keys do not corrupt prompts (e.g. URLs). POSIX behavior stays largely unchanged because the tty still echoes.
> 
> Adds **unit tests** for `readLine` covering newlines/CRLF, EOF, escape discarding, and bare Escape handling.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 30b2c409e962ed9a04585b2081266d984901a2bc. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/internal/auth/auth.go
+++ b/internal/auth/auth.go
@@ -318,7 +318,6 @@ func askForNewHost() bool {
 
 	selectedOption, err := reader.ReadString()
 	if err != nil {
-		fmt.Println("reader.ReadString => err :", err)
 		return false
 	}
 
--- a/internal/auth/auth.go
+++ b/internal/auth/auth.go
@@ -318,6 +318,7 @@ func askForNewHost() bool {
 
 	selectedOption, err := reader.ReadString()
 	if err != nil {
+		fmt.Println("reader.ReadString => err :", err)
 		return false
 	}
 
--- a/internal/auth/auth.go
+++ b/internal/auth/auth.go
@@ -330,7 +330,7 @@ func SetURLAction() bool {
 			printSetURLPrompt()
 
 			url, err := reader.ReadString()
-			if err != nil || url == "\n" {
+			if err != nil || url == "" {
 				break
 			}
 
--- a/internal/auth/auth.go
+++ b/internal/auth/auth.go
@@ -325,6 +325,13 @@ func askForNewHost() bool {
 }
 
 func SetURLAction() bool {
+	if !reader.IsStdinTerminal() {
+		log.Error("No DataRobot URL configured and no terminal available to prompt for one. " +
+			"Run 'dr auth set-url <url>' or set DATAROBOT_ENDPOINT to configure it non-interactively.")
+
+		return false
+	}
+
 	if askForNewHost() {
 		for {
 			printSetURLPrompt()
--- a/internal/misc/reader/reader.go
+++ b/internal/misc/reader/reader.go
@@ -55,7 +55,6 @@ func ReadString() (string, error) {
 
 	str, err := readLine(reader)
 	if err != nil {
-		fmt.Println("readLine err => : ", err)
 		fmt.Println()
 	}
 
--- a/internal/misc/reader/reader.go
+++ b/internal/misc/reader/reader.go
@@ -66,10 +66,14 @@ const (
 	ctrlC     = 0x03 // ASCII ETX, sent by Ctrl+C
 	backspace = '\b'
 	del       = 0x7f // ASCII DEL, sent by Backspace on some terminals; also the exclusive upper bound of printable ASCII
+	esc       = 0x1b // ASCII ESC, begins VT100/ANSI escape sequences (arrows, Home/End, etc.)
 
 	printableASCIIMin = 0x20 // ' ', the lowest printable ASCII byte
 
 	eraseSequence = "\b \b" // backspace, overwrite with space, backspace again
+
+	csiFinalByteMin = 0x40 // '@', low end of the CSI final-byte range (ECMA-48)
+	csiFinalByteMax = 0x7e // '~', high end of the CSI final-byte range (ECMA-48)
 )
 
 // readLine reads bytes until '\n' or '\r', whichever comes first, without
@@ -96,18 +100,17 @@ func readLine(r *bufio.Reader) (string, error) {
 			return sb.String(), err
 		}
 
-		// Ctrl+C (ASCII ETX, 0x03). On POSIX this byte never reaches us —
-		// the tty driver intercepts it at the ISIG layer and raises a real
-		// SIGINT before it's ever placed in the input stream. On Windows,
-		// cancelreader disables ENABLE_PROCESSED_INPUT (required so it can
-		// implement its own Cancel()), which per Microsoft's docs means
-		// "CTRL+C is reported as keyboard input rather than as a signal" —
-		// so this is the only place Ctrl+C is ever observable on Windows.
-		if b == ctrlC {
-			return sb.String(), cancelreader.ErrCanceled
+		// Ctrl+C and escape sequences (arrow keys, Home/End, etc.) never
+		// belong in the answer — see handleControlByte.
+		if consumed, err := handleControlByte(r, b); consumed {
+			if err != nil {
+				return sb.String(), err
+			}
+
+			continue
 		}
 
-		if b == '\n' || b == '\r' {
+		if isLineEnd(b) {
 			if echo {
 				fmt.Println()
 			}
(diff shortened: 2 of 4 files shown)


taras-pokornyy:
@cursor review

internal/misc/reader/reader.go
```diff
@@ -67,6 +242,7 @@ func ReadString() (string, error) {
 func AskYesNo() bool {
 	line, err := ReadString()
 	if err != nil {
+		fmt.Println("err", err)
```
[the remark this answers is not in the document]
taras-pokornyy: @cursor review

internal/misc/reader/reader.go
```diff
@@ -28,37 +29,224 @@ import (
...
+		// with nothing pending — not from a genuine signal. Without this
+		// check, every successful read would still call cr.Cancel() here
+		// (close() unblocks a pending receive same as a real send would),
+		// racing the deferred cr.Close() above: cancelreader has no
+		// synchronization between Cancel() and Close(), and on Windows
+		// Close() skips console-mode restoration entirely if the race makes
+		// CloseHandle(cancelEvent) fail — the exact bug this file fixes.
+		if _, ok := <-cancelChan; ok {
+			cr.Cancel()
+		}
```
ajalon1, line 73: **Suggestion (optional, not requesting a change):** This comment says racing `Cancel()`/`Close()` risks skipping console-mode restoration on Windows. I traced `cancelreader` v0.2.2's Windows `Close()`: it only skips `resetConsole()` if `CloseHandle(cancelEvent)` itself errors (e.g. a genuine double-`Close()`), and `Cancel()` never closes that handle — it only calls `SetEvent` on it.

If `Cancel()` fires after the read already completed, its `blockingReadSignal` handshake has no receiver waiting, so it just times out harmlessly after 100ms. I couldn't find a code path in v0.2.2 where this specific race actually skips console restoration.

The `ok`-check guard itself is still correct and worth keeping — it stops an unnecessary `Cancel()` call on every successful read (100% of them, pre-fix) down to only the rare case where a real SIGINT/SIGTERM lands in the buffered channel right as a read completes. **I'd just soften the comment's claim ("the exact bug this file fixes") rather than presenting it as the confirmed mechanism.**

internal/misc/reader/reader.go
```diff
@@ -28,37 +29,224 @@ import (
...
+		b, err := r.ReadByte()
+		if err != nil {
+			return err
+		}
+
+		if b >= csiFinalByteMin && b <= csiFinalByteMax {
+			return nil
+		}
+	}
+}
```
ajalon1, line 228: **Suggestion (optional, not requesting a change):** Only CSI sequences (`ESC [ ...`) are stripped here. Confirmed via a scratch repro that an SS3-style sequence (`ESC O A`, sent by some terminals in "application cursor key" mode) leaks through as literal characters:

```
readLine on "ab\x1bOA cd\n" → "abOA cd"
```

Low-likelihood for a simple line prompt (that mode is normally opt-in for full-screen TUI apps, not typically active for a y/N or URL prompt), but it's the same class of bug this PR is otherwise fixing, just for a rarer encoding.

internal/misc/reader/reader.go
```diff
@@ -28,37 +29,224 @@ import (
...
+
+		if isBackspace(b) {
+			if echo {
+				eraseLastByte(&sb)
+			}
+
+			continue
+		}
+
+		sb.WriteByte(b)
```
ajalon1, line 150: **Nit:** Bytes below `0x20` other than the explicitly-handled ones (Ctrl+C, ESC, `\n`/`\r`, backspace/DEL) fall through to `sb.WriteByte(b)` unfiltered. Confirmed a stray Tab (`\t`) silently ends up in the answer:

```
readLine on "ab\tcd\n" → "ab\tcd" (not echoed, since it's outside the printable range, but present in the returned string)
```

Unlikely to matter for a URL/y-n prompt, but worth knowing about.

ajalon1 (commented):
I pulled the actual `windows-latest` CI run for this PR — `cmd/plugin/install` and `internal/auth`, the two suites that used to hang for 10 minutes, now complete in 0.37s and 0.12s respectively, so very nice.

Traced the `muesli/cancelreader` v0.2.2 source directly to check the PR description's technical claims; details in comments below.

Nice fix — the root-cause chain (Windows bypass → real bug it hid → cancelreader lifecycle bug it then exposed) is genuinely tricky and this holds up well under scrutiny.

Heads up: Cursor Bugbot's 4 flagged issues on this PR are all anchored to earlier commits (`6a8c31c99af2`, `30b2c409e962`, `f68953f67a24`), not the final 3 commits. I confirmed all 4 are already fixed in the current diff — but since Bugbot hasn't re-run against the final version, it might be worth a fresh `@cursor review` before merge just to have that anchored to what's actually being merged.

None of the comments below are blocking.

internal/misc/reader/reader.go
```diff
@@ -28,37 +29,224 @@ import (
...
+// override it to inject canned input instead of reassigning the os.Stdin
+// global. Reassigning os.Stdin itself doesn't work on Windows: cancelreader's
+// Windows implementation special-cases "is this reader's fd the same as
+// os.Stdin's fd?" and, when true, discards the reader it was given in favor
+// of opening the real console handle (CONIN$) directly — so a test-supplied
+// pipe assigned to os.Stdin gets ignored and the read blocks forever waiting
+// for real keystrokes. Overriding this var instead leaves the real os.Stdin
+// untouched, so the fds never match and cancelreader takes its generic
+// fallback path, which actually reads from the injected pipe.
+var Stdin io.Reader = os.Stdin
```
ajalon1, line 42: tl;dr - good catch

**Praise:** This is a well-targeted fix. Confirmed in `cancelreader`'s Windows source that reassigning the global `os.Stdin` (the old test pattern) creates a self-comparison bug: the library's fd check compares the passed reader's fd against `os.Stdin.Fd()`, and once you reassign `os.Stdin` itself, both sides of that comparison read the same variable — making it unconditionally true regardless of what you reassigned it to, which forces the CONIN$ real-console fallback every time. A distinct package variable sidesteps that entirely.

internal/misc/reader/reader.go
```diff
@@ -28,37 +29,224 @@ import (
...
+	cr, err := cancelreader.NewReader(Stdin)
 	if err != nil {
 		return "", err
 	}
 
+	// cancelreader must be closed. That never happened here: leaks epoll/pipe
+	// fds on Linux/macOS, and leaves Windows in raw console mode (no echo) —
+	// the garbled input CFX-4799 papered over by bypassing cancelreader on
+	// Windows instead of closing it.
+	defer cr.Close()
```
ajalon1, line 54: good catch

internal/auth/auth.go
```diff
@@ -325,12 +325,19 @@ func askForNewHost() bool {
 }
 
 func SetURLAction() bool {
+	if !reader.IsStdinTerminal() {
+		log.Error("No DataRobot URL configured and no terminal available to prompt for one. " +
+			"Run 'dr auth set-url <url>' or set DATAROBOT_ENDPOINT to configure it non-interactively.")
+
+		return false
+	}
```
ajalon1, line 333: tl;dr consistent, thank you

**Praise:** Confirmed this matches the existing convention in `cmd/artifact/del`, `cmd/workload/del`, and `cmd/plugin/discovery.go` exactly (check-before-prompt, explicit short-circuit), and it's placed correctly — before `askForNewHost()`, which also reads from stdin.

internal/misc/reader/reader_test.go
```diff
@@ -15,11 +15,94 @@
 package reader
```
ajalon1, line 15: **Praise:** Good coverage for a byte-level parser: CRLF variants, EOF with/without trailing newline, arrow keys mid-line, and the bare-Escape-keypress unread-byte case, which is an easy one to get wrong.

ajalon1 (approved):
thoughts in previous review, but LGTM

PR #691, opened 2026-07-21
[CFX-6984] Include DataRobot-deployed LLMs in dr llm-gateway list and select

chasdr:
## Summary

`dr llm-gateway list` and `dr llm-gateway select` only knew about LLM Gateway catalog models. On-prem and air-gapped installs where the gateway catalog is empty had nothing to list or select. This adds DataRobot-deployed LLMs (deployments whose champion model is a text-generation/chat model) alongside the gateway models, so those installs get a usable list and you can pick a deployed model by its id.

## How it works

`list` and `select` now pull from two sources: the LLM Gateway catalog and `/api/v2/deployments/` filtered to active text-generation deployments. The two are merged into one list. A `SOURCE` column (and a `source` field in JSON output) marks each entry `gateway` or `deployed`, and deployed entries carry a `deployment_id` so downstream tooling knows to route through the deployment instead of the gateway.

Fetching is best-effort. If one source is unavailable (an empty gateway on-prem, or no deployment access on SaaS), the command warns and lists the other, and only errors when both fail. When a source is down and you `select` an id that isn't in the surviving list, the error names the unavailable source instead of a bare "not found". `dr dotenv`'s model picker is unchanged (gateway-only).

## Technical Changes

- internal/drapi/llms.go: `GetDeployedLLMs` lists active `TextGeneration` deployments (server-side `championModelTargetType` filter plus a client re-filter for older builds that ignore it, and a label-to-id fallback). `GetLLMsAndDeployed` unions gateway + deployed, records per-source failures on `LLMList.Warnings`, and errors only if both sources fail.
- cmd/llm-gateway/list/cmd.go: JSON gains `source` + `deployment_id`; the table gains a `SOURCE` column, deployed rows show `-` for provider/model/context.
- cmd/llm-gateway/select/cmd.go + model.go: `select` accepts a deployment id; a not-found error names the unavailable source when one failed; the TUI picker labels deployed rows with the deployment id.
- docs/commands/llm-gateway.md, docs/commands/README.md: document the two sources, the `SOURCE` column, and the new JSON fields.
- internal/drapi/llms_test.go: 8 new drapi tests (mapping, target-type/active filter, pagination, soft-degrade + warnings, empty-label fallback), plus list/select coverage for the union rendering, the JSON contract, and the enriched select error.

Verified end-to-end against a live environment with deployed LLMs: `list` shows both sources, `select` persists a deployment id, and `list` then marks it selected.

### Screenshots

`dr llm list`

imageimage

`dr llm select`

image

---

> [!NOTE]
> **Medium Risk**
> Touches default LLM selection and a new deployments API path with partial-failure semantics; behavior change for on-prem users but well-tested and backward-compatible for gateway-only IDs.
> 
> **Overview**
> **`dr llm-gateway list`** and **`select`** now load LLMs from the gateway catalog and from active **TextGeneration** deployments, merged via **`GetLLMsAndDeployed`**. One failing source is warned and skipped; the command fails only when both fail. **`select`** appends those warnings when lookup fails on a partial list.
> 
> **List output** adds a **`SOURCE`** column and JSON **`source`** / **`deployment_id`** (deployed rows use `-` for provider/model in the table; the litellm sentinel stays in JSON). **`default-llm-id`** may be a gateway id or a deployment id. The TUI picker shows **`deployed · <id>`** for deployment rows. Docs and tests cover the union, soft-degrade behavior, and the JSON wire contract.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 194893e337e1d5b30393811f07a95f1a58b9da1d. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/cmd/llm-gateway/list/cmd.go
+++ b/cmd/llm-gateway/list/cmd.go
@@ -54,7 +54,7 @@ func Cmd() *cobra.Command {
 	cmd := &cobra.Command{
 		Use:          "list",
 		Aliases:      []string{"ls"},
-		Short:        "List available LLM Gateway models",
+		Short:        "List available LLMs",
 		Args:         cobra.NoArgs,
 		SilenceUsage: true,
 		PreRunE:      auth.EnsureAuthenticatedE,
--- a/cmd/llm-gateway/list/cmd.go
+++ b/cmd/llm-gateway/list/cmd.go
@@ -33,14 +33,19 @@ import (
 )
 
 // LLMOutput is the JSON representation of an LLM for --output-format json.
+// Source discriminates gateway catalog models from DataRobot-deployed LLMs;
+// DeploymentID is set only for deployed entries and is what downstream tooling
+// uses to write the deployed-model .env keys.
 type LLMOutput struct {
-	ID          string `json:"id"`
-	Name        string `json:"name"`
-	Provider    string `json:"provider"`
-	Model       string `json:"model"`
-	Description string `json:"description"`
-	ContextSize int    `json:"context_size"`
-	Selected    bool   `json:"selected"`
+	ID           string `json:"id"`
+	Name         string `json:"name"`
+	Source       string `json:"source"`
+	Provider     string `json:"provider"`
+	Model        string `json:"model"`
+	Description  string `json:"description"`
+	ContextSize  int    `json:"context_size"`
+	DeploymentID string `json:"deployment_id"`
+	Selected     bool   `json:"selected"`
 }
 
 func Cmd() *cobra.Command {
@@ -54,7 +59,7 @@ func Cmd() *cobra.Command {
 		SilenceUsage: true,
 		PreRunE:      auth.EnsureAuthenticatedE,
 		RunE: func(cmd *cobra.Command, _ []string) error {
-			llmList, err := drapi.GetLLMs()
+			llmList, err := drapi.GetLLMsAndDeployed()
 			if err != nil {
 				return err
 			}
@@ -90,13 +95,15 @@ func toLLMOutputs(llms []drapi.LLM, selectedID string) []LLMOutput {
 
 	for i, l := range llms {
 		outputs[i] = LLMOutput{
-			ID:          l.LlmID,
-			Name:        l.Name,
-			Provider:    l.Provider,
-			Model:       l.Model,
-			Description: l.Description,
-			ContextSize: l.ContextSize,
-			Selected:    l.LlmID == selectedID,
+			ID:           l.LlmID,
+			Name:         l.Name,
+			Source:       l.Kind,
+			Provider:     l.Provider,
+			Model:        l.Model,
+			Description:  l.Description,
+			ContextSize:  l.ContextSize,
+			DeploymentID: l.DeploymentID,
+			Selected:     l.LlmID == selectedID,
 		}
 	}
 
@@ -123,7 +130,7 @@ func formatContextSize(n int) string {
 }
 
 func printLLMTable(llms []drapi.LLM, selectedID string) {
-	fmt.Println(tui.SubTitleStyle.Render("LLM Gateway Models"))
+	fmt.Println(tui.SubTitleStyle.Render("Available LLMs"))
 
 	idStyle := tui.BaseTextStyle.
 		Foreground(tui.GetAdaptiveColor(tui.DrPurple, tui.DrPurpleDark)).
@@ -148,7 +155,7 @@ func printLLMTable(llms []drapi.LLM, selectedID string) {
 				return dimStyle
 			}
 		}).
-		Headers("ID", "NAME", "PROVIDER", "MODEL", "CONTEXT")
+		Headers("ID", "NAME", "SOURCE", "PROVIDER", "MODEL", "CONTEXT")
 
 	for _, l := range llms {
 		id := "  " + l.LlmID
@@ -156,7 +163,15 @@ func printLLMTable(llms []drapi.LLM, selectedID string) {
 			id = "* " + l.LlmID
 		}
 
-		t.Row(id, l.Name, l.Provider, l.Model, formatContextSize(l.ContextSize))
+		// Deployed rows carry no provider and only the litellm sentinel model;
+		// both are noise in the table (the SOURCE column already says
+		// "deployed"), so show "-" and keep the sentinel in JSON output only.
+		provider, model := l.Provider, l.Model
+		if l.Kind == drapi.LLMKindDeployed {
+			provider, model = "-", "-"
+		}
+
+		t.Row(id, l.Name, l.Kind, provider, model, formatContextSize(l.ContextSize))
 	}
 
 	rendered := t.Render()
--- a/internal/drapi/llms.go
+++ b/internal/drapi/llms.go
@@ -163,7 +163,7 @@ type deploymentList struct {
 // query params and return every deployment, so rows are re-filtered
 // client-side on target type and active status.
 func GetDeployedLLMs() ([]LLM, error) {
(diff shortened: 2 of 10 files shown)


internal/drapi/llms.go
```diff
@@ -103,3 +136,118 @@ func GetLLMs() (*LLMList, error) {
 
 	return &llmList, nil
```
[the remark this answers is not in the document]
chasdr: pre-existing in GetLLMs, not introduced by this PR. nothing reads LLMList.Count or TotalCount: list and select go through GetLLMsAndDeployed which sets its own count from len(), and dotenv reads .LLMs. harmless, leaving it. can make GetLLMs self-consistent in a followup if we want.

cmd/llm-gateway/list/cmd.go
```diff
@@ -148,15 +155,23 @@ func printLLMTable(llms []drapi.LLM, selectedID string) {
...
 		}
 
-		t.Row(id, l.Name, l.Provider, l.Model, formatContextSize(l.ContextSize))
+		// Deployed rows carry no provider and only the litellm sentinel model;
+		// both are noise in the table (the SOURCE column already says
+		// "deployed"), so show "-" and keep the sentinel in JSON output only.
+		provider, model := l.Provider, l.Model
+		if l.Kind == drapi.LLMKindDeployed {
+			provider, model = "-", "-"
+		}
```
ajalon1, line 172: huh

internal/drapi/llms.go
```diff
@@ -103,3 +136,118 @@ func GetLLMs() (*LLMList, error) {
...
+// GetDeployedLLMs lists DataRobot Deployments serving as chat LLMs (champion
+// model target type TextGeneration). The server-side championModelTargetType
+// filter is honored on recent platforms; older on-prem builds ignore unknown
+// query params and return every deployment, so rows are re-filtered
+// client-side on target type and active status.
+func GetDeployedLLMs() ([]LLM, error) {
+	url, err := config.GetEndpointURL("/api/v2/deployments/?championModelTargetType=" + targetTypeTextGeneration + "&limit=100")
+	if err != nil {
+		return nil, err
+	}
```
ajalon1, line 169: I honestly wonder if this could be pulled into its own `deployments` module, but maybe not yet.

internal/drapi/llms.go
```diff
@@ -103,3 +136,118 @@ func GetLLMs() (*LLMList, error) {
...
+}
+
+// GetLLMsAndDeployed returns the union of LLM Gateway catalog models and
+// DataRobot-deployed LLMs. Each source is best-effort: a single-source failure
+// is logged and the other source is still returned, so an empty or disabled
+// gateway (common on-prem) or missing deployment access does not blank the
+// list. An error is returned only when both sources fail.
+func GetLLMsAndDeployed() (*LLMList, error) {
+	gateway, gwErr := GetLLMs()
+	deployed, depErr := GetDeployedLLMs()
```
ajalon1, line 224: Sequential is OK? Don't need to run these in parallel?

internal/drapi/llms.go
```diff
@@ -103,3 +136,118 @@ func GetLLMs() (*LLMList, error) {
...
+	}
+
+	return deployed, nil
+}
+
+// GetLLMsAndDeployed returns the union of LLM Gateway catalog models and
+// DataRobot-deployed LLMs. Each source is best-effort: a single-source failure
+// is logged and the other source is still returned, so an empty or disabled
+// gateway (common on-prem) or missing deployment access does not blank the
+// list. An error is returned only when both sources fail.
```
ajalon1, line 221: I really appreciate the design here. warnings and errors.

internal/drapi/llms.go
```diff
@@ -103,3 +136,118 @@ func GetLLMs() (*LLMList, error) {
...
+
+	for url != "" {
+		var dl deploymentList
+
+		if err = GetJSON(url, "deployed LLMs", &dl); err != nil {
+			return nil, err
+		}
+
+		for _, d := range dl.Data {
+			if d.Model.TargetType != targetTypeTextGeneration || !strings.EqualFold(d.Status, "active") {
```
ajalon1, line 181: Why this extra check on TargetType? We're already querying with `championModelTargetType=TextGeneration`. Is there some weirdness in our /deployments API?

internal/drapi/llms.go
```diff
@@ -103,3 +136,118 @@ func GetLLMs() (*LLMList, error) {
...
+	TotalCount int          `json:"totalCount"`
+	Next       string       `json:"next"`
+	Previous   string       `json:"previous"`
+}
+
+// GetDeployedLLMs lists DataRobot Deployments serving as chat LLMs (champion
+// model target type TextGeneration). The server-side championModelTargetType
+// filter is honored on recent platforms; older on-prem builds ignore unknown
+// query params and return every deployment, so rows are re-filtered
+// client-side on target type and active status.
```
ajalon1, line 164: Interesting extra filter. How new is this addition to /deployments, I wonder.

ajalon1 (commented):
Screenshots of how it looks? @shreyaag-dr will be interested to see

internal/drapi/llms.go
chasdr: 👋 chas's claude here. `championModelTargetType` only landed on `/deployments/` around v2.34 (2024). older on-prem builds ignore the unknown param and hand back every deployment, so the client-side `targetType == TextGeneration` check is the belt to that server-side suspenders. redundant on SaaS, load-bearing air-gapped. (docstring says as much 6 lines up.)

internal/drapi/llms.go
chasdr: added 2024-08-13, ~v2.34-era. new enough that i wouldn't bet every install has it, which is the whole reason for the client re-filter your droid flagged one thread down.

internal/drapi/llms.go
chasdr: intentional. cold interactive path, both calls network-bound so ~2 RTTs worst case, and going parallel buys a sub-second win for goroutines plus a second token-verify round trip. not worth it on a list command. easy errgroup later if it ever lands on a latency gripe.

internal/drapi/llms.go
chasdr: agreed, and "not yet" is the right instinct. one function against `/deployments/` today. first time we grow a real deployments surface (get/describe/etc) it earns its own package. your droid can take that one.

internal/drapi/llms.go
chasdr: chas's claude accepts on behalf of the soft-degrade: warnings not errors, so an empty gateway on-prem still lists deployed models instead of face-planting.

cmd/llm-gateway/list/cmd.go
chasdr: guessing the "huh" is the `-` blanking: deployed rows have no provider and their model is always the sentinel `datarobot/datarobot-deployed-llm`, pure noise in a table whose SOURCE column already says "deployed". so the table shows `-` and JSON keeps the real sentinel for the skill. if the huh was aimed elsewhere, point your droid's cursor and i'll bite :D

chasdr:
> Screenshots of how it looks? @shreyaag-dr will be interested to see

Description updated with screenshots 👍   I happen to have the docs assist token which happens to have a bunch of real deployments to see - very convenient

cmd/llm-gateway/list/cmd.go
ajalon1: human principal here, I was just confused, all good :)

chasdr:
@ajalon1 @shreyaag-dr 
fyi this is blocking me from working on other tickets

ajalon1 (approved):
I was only wondering about the presentation layer; refactoring that is cheap.

shreyaag-dr (commented):
lgtm

PR #697, opened 2026-07-22
Feat: Allow dr auth login to accept hostname and default to https protocol

burkestar:
## RATIONALE

I keep tripping over `dr auth login some-host.datarobot.com` complaining its not a full URL.  It's a minor annoyance that is easily preventable.

This allows specifying a hostname only and using `https://` as the default implied protocol.

---

> [!NOTE]
> **Low Risk**
> Small, localized change to URL parsing for CLI config; defaulting to https is the expected behavior for DataRobot endpoints.
> 
> **Overview**
> **`SchemeHostOnly`** now trims whitespace and, when the input has no `://`, prepends **`https://`** before parsing. That lets **`dr auth login`** (and **`SetURLToConfig` / `SaveURLToConfig`**) accept values like `app.datarobot.com` or `app.datarobot.com/api/v2` the same as full `https://…` URLs.
> 
> Tests add cases for bare hostname and hostname-with-path; the invalid-input example is updated to **`not a url`** (spaces in the host) so it still fails after the implicit-scheme behavior.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 5a7c774837fd8028f50ee3437fffe12a84af5cd6. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/internal/config/api.go
+++ b/internal/config/api.go
@@ -30,8 +30,20 @@ const DRAPIURLSuffix = "/api/v2"
 var ErrInvalidURL = errors.New("Invalid URL.")
 
 // SchemeHostOnly takes a URL like: https://app.datarobot.com/api/v2 and just
-// returns https://app.datarobot.com (no trailing slash)
+// returns https://app.datarobot.com (no trailing slash).
+//
+// A bare hostname without a scheme (e.g. "app.datarobot.com") is also accepted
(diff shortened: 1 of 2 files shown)


c-h-russell-walker (approved):
LGTM and make sense 👍 

Thanks for the contribution.

PR #704, opened 2026-07-23
env tool mvp

brau0300:
# RATIONALE
mvp of a workload cli tool for managing environment vars on the workload

## CHANGES
support for:
dr workload env list <workload_id>
dr workload env set <workload_id> FOO=bar API_KEY=dr-credential:6a622b48767447ca7829c6ad/apiToken
dr workload env delete <workload_id> FOO

see detailed spec in the docs/development/env_spec.md

---

> [!NOTE]
> **Medium Risk**
> Changes trigger rolling workload redeploys and artifact clones/locks; mistakes can affect running services, though guards, confirmation, and broad httptest coverage reduce operational risk.
> 
> **Overview**
> Adds **`dr workload env`** under the workload command with **`list`**, **`set`**, and **`delete`** for the primary container’s environment variables on the artifact the workload runs.
> 
> **`set`** and **`delete`** implement the artifact edit flow (patch draft in place, clone when locked), then optionally roll out via a shared **`rollout`** helper: confirm redeploy (`--yes` / non-interactive), lock clones only at deploy time, block duplicate in-flight replacements (skipped for **`--stage`**), and optional **`--wait`** polling. **`set`** parses `KEY=VALUE` and `dr-credential:<id>/<key>`, validates K8s-style names and credential IDs before writes; **`delete`** refuses no-op removals before cloning locked artifacts.
> 
> New **`internal/workload`** pieces cover env mutation on raw artifact specs, credential lookup, replacement start/wait (including **`errored`** as failure), and text/JSON rendering. **`docs/development/env_spec.md`** documents the live API contract and corner cases.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 085801619546367c6ae32cda789a3a20bb5da6ea. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


chasdr (commented):
hi! 👋 chas's claude here, did a pass on this for chas.

really clean for an mvp. the api signatures are verified against staging and documented inline, the raw-JSON traversal preserves sidecar containers untouched, secrets never get printed or logged, and the tests actually drive the http layer via httptest instead of mocking it. `task lint` clean, all green. no blockers from me.

three small things inline: one worth a look, two nits.

since chas mentioned this is hackathon work and any ideas are fair game, a few cool-takes to point your cursor agent at if you've got time:

- **`--env-file .env`**: bulk-import a dotenv in one shot. it's a thin wrapper over the `parseEnvArg` + `UpsertEnvironmentVars` path you already have, and it's a killer demo moment ("point it at my .env and redeploy").
- **show the env diff in the confirm prompt**: right now the y/N just names the target artifact. you've already got the old vars (`PrimaryEnvironmentVars`) and the new ones in hand, so printing `+ FOO=1  /  LOG_LEVEL debug→info  /  - BAR` before the prompt is the bit that makes an audience nod. pairs perfectly with a `--dry-run` that computes the same diff and exits before writing.
- **`env edit` (kubectl-style)**: dump current env into `$EDITOR` as a dotenv, apply the diff on save. reuses list + the diff + upsert/remove. flashiest one if you've got the time.
- **credential key hinting**: you smartly skip validating the credential *key* (that mapping drifts, good call). but since `GetCredential` already fetches the credential, on a rollout failure you could surface "credential X has keys: [awsAccessKeyId, ...]" so a typo'd key doesn't just silently crash-loop the container later.

nice work, solid base. 🚀

cmd/workload/env/internal/rollout/rollout.go
```diff
@@ -0,0 +1,175 @@
...
+
+		return nil
+	}
+
+	confirmed, err := confirmRollout(cmd, opts.Yes, workloadID, targetArtifactID)
+	if err != nil || !confirmed {
+		return err
+	}
+
+	if needsLock {
```
chasdr, line 69: one race worth a look: when `needsLock` is true, `targetArtifactID` is the fresh clone, and this locks it *before* the `GuardNoActiveReplacement` call just below. if a replacement sneaks in between the upfront guard and here, the guard fails after the clone's already locked. and locked artifacts can't be deleted, so it's orphaned for good.

would moving the guard above this `if needsLock` block close it? the clone stays a deletable draft on a guard failure, and the lock still lands before `StartReplacement`.

cmd/workload/env/set/cmd.go
```diff
@@ -0,0 +1,295 @@
...
+
+// envVarNamePattern is Kubernetes' own container env var name rule
+// (IsEnvVarName in k8s.io/apimachinery/pkg/util/validation): starts with a
+// letter, underscore, or dot, followed by letters, digits, underscores,
+// dots, or dashes. Verified live against staging that the platform accepts
+// -- and silently writes -- names violating this (spaces, a leading digit,
+// dashes) at PATCH time with no complaint; rejecting here catches a typo
+// immediately instead of it surfacing later as a replacement failure or a
+// crash-looping container with no obvious link back to this command.
+var envVarNamePattern = regexp.MustCompile(`^[-._a-zA-Z][-._a-zA-Z0-9]*$`)
```
chasdr: tiny one: this regex matches k8s `IsEnvVarName` exactly (leading dash allowed), but the comment above says "starts with a letter, underscore, or dot" and the error message says "must not start with a digit". both slightly under-describe what the regex actually permits. the regex is right, just the prose drifts. word tweak whenever you're in here.

cmd/workload/env/set/cmd.go
```diff
@@ -0,0 +1,295 @@
...
+	workloadID := args[0]
+
+	stageFlag, _ := cmd.Flags().GetBool("stage")
+
+	// Fail fast, before parsing/validating anything or mutating any
+	// artifact, if a rollout couldn't happen right now anyway. Skipped when
+	// staging: staging never touches the live rollout machinery, so it's
+	// safe to prepare a follow-up edit while an earlier replacement settles.
+	if !stageFlag {
+		if err := rollout.GuardNoActiveReplacement(workloadID); err != nil {
```
chasdr, line 157: nit, mostly a question: this network guard runs before the local `KEY=VALUE` parse loop below, so a typo'd arg costs a round-trip before the local error fires. parsing first would fail faster on bad input. though i get it if the "don't do any work if a rollout can't happen anyway" ordering is deliberate.

PR #719, opened 2026-07-28
[CMPT-6969] feat(pipeline): add `dr pipeline clone` command

sunny2get:
## Summary

Adds `dr pipeline clone <pipeline-id> [--name]`, mirroring the new pipelines-api endpoint `POST /api/v2/pipelines/{id}/clone`. Clones a pipeline into a new **draft**, copying its source code, description, latest input params, and assigned image. The name defaults to `Clone of <source name>` and is overridable via `--name`; the server auto-suffixes on collision so repeated clones never fail.

> Depended on the pipelines-api clone endpoint (datarobot/pipelines-api#234), which has now **merged** — this is ready to merge/release once it reaches the target environment.

## Changes

- **`internal/pipeline`**: `ClonePipeline(pipelineID, name)` → `POST .../clone` with a JSON body (`name` omitempty), mirroring `LockPipeline`.
- **`cmd/pipeline/clone`**: cobra subcommand following the **write-command** pattern — returns `fmt.Errorf("clone pipeline: %w", err)` and renders via `RenderCreateResponse`. No read-command friendly-404 handler, so JSON stdout is never corrupted (consistent with #707).
- **Tests**: `internal/pipeline/clone_test.go` (httptest: default name omits body key, `--name` override, 404) + `cmd/pipeline/clone/cmd_test.go` (JSON/human render, invalid-output + blank-name rejection).
- **`docs/commands/pipeline.md`**: clone row.

## Test

`go build ./...`, `go vet`, `gofmt` clean; `go test ./cmd/pipeline/clone/... ./internal/pipeline/...` passes.

---

> [!NOTE]
> **Low Risk**
> Additive CLI and API client for a new write endpoint; no changes to auth or existing pipeline flows beyond registering a subcommand.
> 
> **Overview**
> Adds **`dr pipeline clone <pipeline-id>`** so users can duplicate an existing pipeline into a new **draft** via `POST /api/v2/pipelines/{id}/clone`, copying source, description, latest input, and image as described in the command help.
> 
> The client sends an optional **`--name`** in JSON (`name` omitted when unset so the API can default to `Clone of <source>`); whitespace-only `--name` is rejected locally. Success output reuses **`RenderCreateResponse`** (same shape as create/lock). The subcommand is registered under `dr pipeline`, includes auth, output format, and telemetry, and is documented in the pipeline command table.
> 
> **`internal/pipeline.ClonePipeline`** mirrors **`LockPipeline`**’s JSON POST pattern. Tests cover default vs override name, 404 handling, rendering, and flag validation.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit cc6277a1c7203eda22043fcf4f955a697c14d82b. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/cmd/pipeline/clone/cmd_test.go
+++ b/cmd/pipeline/clone/cmd_test.go
@@ -1,0 +1,88 @@
+// Copyright 2026 DataRobot, Inc. and its affiliates.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package clone
+
+import (
+	"encoding/json"
+	"io"
+	"testing"
+	"time"
+
+	"github.com/datarobot/cli/cmd/pipeline/internal/testutil"
+	"github.com/datarobot/cli/internal/outputformat"
+	"github.com/datarobot/cli/internal/pipeline"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+// sample is a clone response: always a fresh draft, so Version/Status are
+// intentionally zero-valued (the server returns none for a draft, same as
+// POST /pipelines).
+func sample() pipeline.CreateResponse {
+	return pipeline.CreateResponse{
+		PipelineID: "p-2",
+		Name:       "Clone of wf",
+		Mode:       "draft",
+		TaskNames:  []string{"e1", "e2"},
+		CreatedAt:  time.Date(2026, 4, 30, 10, 0, 0, 0, time.UTC),
+	}
+}
+
+func TestPrintCloneJSON(t *testing.T) {
+	output := testutil.CaptureStdout(t, func() {
+		require.NoError(t, pipeline.RenderCreateResponse(outputformat.OutputFormatJSON, sample()))
+	})
+
+	var parsed map[string]any
+
+	require.NoError(t, json.Unmarshal([]byte(output), &parsed))
+	assert.Equal(t, "p-2", parsed["id"])
+	assert.Equal(t, "Clone of wf", parsed["name"])
+	assert.Equal(t, "draft", parsed["mode"])
+}
+
+func TestPrintCloneHuman(t *testing.T) {
+	output := testutil.CaptureStdout(t, func() {
+		require.NoError(t, pipeline.RenderCreateResponse(outputformat.OutputFormatText, sample()))
+	})
+
+	assert.Contains(t, output, "Clone of wf")
+	assert.Contains(t, output, "draft")
+	assert.Contains(t, output, "e1, e2")
+}
+
+func TestCmd_RejectsInvalidOutput(t *testing.T) {
+	cmd := Cmd()
+	cmd.SetArgs([]string{"abc", "--output-format", "yaml"})
+	cmd.SetOut(io.Discard)
+	cmd.SetErr(io.Discard)
+	cmd.PreRunE = nil
+
+	err := cmd.Execute()
+	require.Error(t, err)
+	assert.Contains(t, err.Error(), "invalid output format")
+}
+
+func TestCmd_RejectsBlankName(t *testing.T) {
+	cmd := Cmd()
+	cmd.SetArgs([]string{"abc", "--name", "   "})
+	cmd.SetOut(io.Discard)
+	cmd.SetErr(io.Discard)
+	cmd.PreRunE = nil
+
+	err := cmd.Execute()
+	require.Error(t, err)
+	assert.Contains(t, err.Error(), "--name must not be blank")
+}
--- a/docs/commands/pipeline.md
+++ b/docs/commands/pipeline.md
@@ -76,6 +76,7 @@ dr pipeline lock <pipeline-id>
 | `dr pipeline create`    | `POST   /api/v2/pipelines`           | Upload a Python file to register a new pipeline. |
 | `dr pipeline list`      | `GET    /api/v2/pipelines`           | Paginated list with mode filtering.              |
 | `dr pipeline get`       | `GET    /api/v2/pipelines/{id}`      | Pipeline detail including all versions.          |
+| `dr pipeline clone`     | `POST   /api/v2/pipelines/{id}/clone` | Clone a pipeline into a new draft (source, description, latest input, image). |
 | `dr pipeline update`    | `PATCH  /api/v2/pipelines/{id}`      | Re-upload a file to append a new version.        |
 | `dr pipeline delete`    | `DELETE /api/v2/pipelines/{id}`      | Remove a pipeline and all of its versions.       |
 | `dr pipeline lock`      | `PATCH  /api/v2/pipelines/{id}/mode` | Promote a draft to locked mode.                  |
@@ -542,8 +543,8 @@ If the task ID is not found, the command prints `Task not found: <task-id>` and 
 | Status | Cause                                                                          |
 |--------|--------------------------------------------------------------------------------|
 | `400`  | Invalid Python file or mismatched pipeline name.                               |
-| `404`  | The provided `<pipeline-id>`, version, or run does not exist.                  |
-| `409`  | Tried to update a `locked` pipeline, or cancel an already-terminal run.        |
+| `404`  | The provided `<pipeline-id>`, version, run, or task does not exist.            |
+| `409`  | Tried to update a `locked` pipeline; cancelled an already-terminal run; requested `run task result` before the task reached `COMPLETED`; or addressed a fan-out `run task` (`get`/`logs`/`result`) without `--node-id` (message lists the candidate node ids). |
 
 ## See also
 
--- a/internal/pipeline/clone_test.go
+++ b/internal/pipeline/clone_test.go
@@ -1,0 +1,120 @@
+// Copyright 2026 DataRobot, Inc. and its affiliates.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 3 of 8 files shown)


internal/pipeline/pipeline.go
```diff
@@ -258,6 +258,33 @@ func LockPipeline(pipelineID string) (*CreateResponse, error) {
...
+
+// ClonePipeline issues POST /api/v2/pipelines/{pipeline_id}/clone to copy an
+// existing pipeline into a new draft (source, description, latest input, and
+// assigned image). An empty name lets the server default to
+// "Clone of <source name>". The response mirrors a create payload.
+func ClonePipeline(pipelineID, name string) (*CreateResponse, error) {
+	endpoint, err := config.GetEndpointURL("/api/v2/pipelines/" + pipelineID + "/clone")
+	if err != nil {
+		return nil, err
+	}
```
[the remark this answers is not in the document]
sunny2get: Applied — `ClonePipeline` now `url.PathEscape`s the pipeline id before building the endpoint (mirrors internal/workload/workload.go).

internal/pipeline/clone_test.go
```diff
@@ -0,0 +1,101 @@
...
+
+	defer srv.Close()
+
+	installEndpoint(t, srv.URL)
+
+	got, err := ClonePipeline("p-1", "")
+	require.NoError(t, err)
+	assert.Equal(t, "p-2", got.PipelineID)
+	assert.Equal(t, "Clone of wf", got.Name)
+	assert.Equal(t, "draft", got.Mode)
```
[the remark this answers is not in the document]
sunny2get: A clone always returns a fresh draft, so the server sends no version/status (same as POST /pipelines) — the fixture omits them deliberately. I made that explicit: the test now asserts `got.Version == 0` / `got.Status == ""`. Adding version:1/status:READY would misrepresent the clone as a locked v1.

cmd/pipeline/clone/cmd_test.go
```diff
@@ -0,0 +1,85 @@
...
+
+func sample() pipeline.CreateResponse {
+	return pipeline.CreateResponse{
+		PipelineID: "p-2",
+		Name:       "Clone of wf",
+		Mode:       "draft",
+		TaskNames:  []string{"e1", "e2"},
+		CreatedAt:  time.Date(2026, 4, 30, 10, 0, 0, 0, time.UTC),
+	}
+}
```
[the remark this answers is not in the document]
sunny2get: Same rationale — the clone is a draft, so Version/Status are intentionally zero-valued (documented with a comment on the fixture). Populating them would portray the clone as locked, which it never is.

HanuMayil (approved):

internal/pipeline/pipeline.go
```diff
@@ -258,6 +258,37 @@ func LockPipeline(pipelineID string) (*CreateResponse, error) {
...
+}
+
+// ClonePipeline issues POST /api/v2/pipelines/{pipeline_id}/clone to copy an
+// existing pipeline into a new draft (source, description, latest input, and
+// assigned image). An empty name lets the server default to
+// "Clone of <source name>". The response mirrors a create payload.
+func ClonePipeline(pipelineID, name string) (*CreateResponse, error) {
+	// Escape the caller-supplied id so reserved path characters can't alter
+	// the request path (mirrors internal/workload/workload.go).
+	escapedID := url.PathEscape(pipelineID)
```
ajalon1: 🤖 Good call adding `url.PathEscape` here — this is the right defense against path-injection via the caller-supplied pipeline ID.

**Observation (out of scope, non-blocking):** The sibling functions in this file — `GetPipeline` (line 131), `UpdatePipeline` (line 149), `DeletePipeline` (line 175), and `LockPipeline` (line 256) — all concatenate the raw `pipelineID` directly into the URL path without escaping. This is the exact issue Copilot flagged on the first revision. Consider a follow-up to apply `url.PathEscape` (or a shared `escapeID` helper like `internal/workload/workload.go` has) consistently across all of them.

internal/pipeline/clone_test.go
```diff
@@ -0,0 +1,120 @@
...
+
+	installEndpoint(t, srv.URL)
+
+	got, err := ClonePipeline("p-1", "")
+	require.NoError(t, err)
+
+	assert.Equal(t, http.MethodPost, gotMethod)
+	assert.Equal(t, "/api/v2/pipelines/p-1/clone", gotPath)
+
+	// No name override → the body must omit "name" entirely.
```
ajalon1, line 62: 🤖 The path-escaping defense added to `ClonePipeline` (line 275 of `pipeline.go`) is the key security improvement in this PR, but no test verifies it works. All three test cases use `"p-1"`, which has no reserved characters.

Consider adding a case that passes an ID with path-special characters (e.g. `"p/1"` or `"..%2F.."`) and asserts the server receives the escaped path (`/api/v2/pipelines/p%2F1/clone`) rather than a rewritten one. This would lock in the defense against regressions.

cmd/pipeline/clone/cmd_test.go
```diff
@@ -0,0 +1,88 @@
...
+	"github.com/datarobot/cli/internal/pipeline"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+// sample is a clone response: always a fresh draft, so Version/Status are
+// intentionally zero-valued (the server returns none for a draft, same as
+// POST /pipelines).
+func sample() pipeline.CreateResponse {
+	return pipeline.CreateResponse{
```
ajalon1, line 34: 🤖 The zero-valued `Version`/`Status` here is well-documented and intentional — a clone is always a fresh draft, same as `POST /pipelines`.

Minor UX note (non-blocking, consistent with `create`): `printCreateResponseHuman` in `pipeline_output.go` always prints `Version: %d` and `Status: %s`, so the human output for a clone will show `Version: 0` and `Status:` (empty). Not wrong, but potentially confusing to users. A future improvement could suppress these fields for drafts, or show a placeholder like `—`.

ajalon1 (commented):
👨🏽‍🚀 The tl;dr? Security seems fine, couple small issues, I would suggest adding another test. 🤖?

## Security + Code Review: `dr pipeline clone`

**Verdict: No security findings.** The new code is well-defended and consistent with the existing `lock` command pattern.

### Security summary

| Area | Result |
|---|---|
| **Path injection** | `url.PathEscape` on the pipeline ID before building the endpoint — good. Actually *better* than the sibling `Get`/`Update`/`Delete`/`Lock` functions which don't escape. |
| **Auth** | `PreRunE: auth.EnsureAuthenticatedE` enforced. |
| **JSON output purity** | `log.Infof` in `doJSON` writes to **stderr** (verified via `internal/log/logger.go`); JSON rendering goes to stdout via `fmt.Println`. No stdout corruption. |
| **Info disclosure** | `--name` is correctly excluded from telemetry. `pipeline_id` sent to telemetry is consistent with existing `lock` command. Debug dumps use `config.RedactedReqInfo`. |
| **SSRF** | Endpoint built from configured base URL, not user-supplied. |
| **Input validation** | `--name` trimmed; blank rejected when flag explicitly set; `omitempty` ensures empty name omitted from JSON body. |
| **Supply chain** | No new dependencies. |

### Code review observations (non-blocking)

1. **No test for the path-escaping defense** — the key security improvement in this PR has no test verifying it works with reserved characters.
2. **Pre-existing inconsistency** — sibling functions (`GetPipeline`, `UpdatePipeline`, `DeletePipeline`, `LockPipeline`) still concatenate raw IDs into the URL path.
3. **Human output shows `Version: 0` / `Status:` (empty) for drafts** — consistent with `create`, documented as intentional by the author.
4. **Docs change bundles unrelated 404/409 error-description updates** — reasonable but tangential to clone.

docs/commands/pipeline.md
```diff
@@ -542,8 +543,8 @@ If the task ID is not found, the command prints `Task not found: <task-id>` and
 | Status | Cause                                                                          |
 |--------|--------------------------------------------------------------------------------|
 | `400`  | Invalid Python file or mismatched pipeline name.                               |
-| `404`  | The provided `<pipeline-id>`, version, or run does not exist.                  |
-| `409`  | Tried to update a `locked` pipeline, or cancel an already-terminal run.        |
+| `404`  | The provided `<pipeline-id>`, version, run, or task does not exist.            |
+| `409`  | Tried to update a `locked` pipeline; cancelled an already-terminal run; requested `run task result` before the task reached `COMPLETED`; or addressed a fan-out `run task` (`get`/`logs`/`result`) without `--node-id` (message lists the candidate node ids). |
```
ajalon1, line 547: 🤖 These 404/409 error-description updates (adding "task" to 404, expanding 409 with `run task result` / fan-out `--node-id` details) seem tangential to the clone command. Reasonable to bundle, but worth confirming they reflect current API behavior and aren't from a different change set.

sunny2get: yes, this is something I am adding which was a leftover commit for the previous change

cmd/pipeline/clone/cmd_test.go
sunny2get: I like the idea of showing `-` , will update

ajalon1 (approved):

internal/pipeline/clone_test.go
sunny2get: good point, adding

cmd/pipeline/clone/cmd_test.go
sunny2get: Done in 1627773 — `printCreateResponseHuman` now renders the `—` placeholder when Version is 0 or Status is empty, so a clone (and a freshly-created draft) shows `Version: —` / `Status: —` instead of `Version: 0` and a blank status line. This shared renderer also backs `create`/`update`/`lock`, so they get the same treatment; JSON output is untouched (still emits raw `version`/`status`). Added draft + locked render tests that assert the placeholder is attached to those specific fields (not merely present somewhere in the output).

internal/pipeline/clone_test.go
sunny2get: Done in 1627773 — added `TestClonePipeline_EscapesPipelineID`, which clones with id `p/1` and asserts the server receives `/api/v2/pipelines/p%2F1/clone`. It captures `r.RequestURI` (the raw wire target) rather than `r.URL.Path` — the latter is already decoded, so it would read `.../p/1/clone` and pass even if the escaping regressed. This locks in the defense against regressions as you suggested.

internal/pipeline/pipeline.go
sunny2get: Done in 0c870dc — bundled this follow-up per the request. Extracted a shared `escapeID` helper (mirroring `internal/workload/workload.go`) and applied it in `GetPipeline`, `UpdatePipeline`, `DeletePipeline`, `LockPipeline`, and `ClonePipeline`, so every path builder in the file escapes the caller-supplied id consistently. It is a no-op for normal 24-char ids and is kept in its own commit, separate from the clone fixes, for easier review.

PR #724, opened 2026-07-30
[CFX-6318] fix(auth): open the browser instead of routing users to the terminal

cdevent:
# RATIONALE

UX research (P4, week of 2026-05-26) found `dr auth` "routed the user to the terminal" instead of opening a browser on a fresh install. Tracing the fresh-install path turned up four separate causes, plus two adjacent bugs.

Jira: [CFX-6318](https://datarobot.atlassian.net/browse/CFX-6318)

## CHANGES

**Flow change** — fresh install, before → after:

| Before | After |
| --- | --- |
| ASCII `[1] US / [2] EU / [3] JP` menu read off stdin | TUI list picker, US Cloud preselected (one keystroke) |
| `Please visit this link…` printed **before** the browser was touched | Browser opened first; link shown as a labelled fallback |
| `open.Open` errors discarded — silent no-op on Windows | Returns an error; failure is reported with the link promoted into a box |
| No spinner, no timeout — bare URL then blocked forever | Spinner in both paths, 5-minute bound |

Root causes: the env picker ran from the `EnsureAuthenticatedE` `PreRunE` on ~70 commands, so it fired before any TUI could render (the wizard's picker was unreachable on a fresh install); and Windows ran `exec.Command("start", url)`, but `start` is a `cmd.exe` builtin with no executable, so it always failed silently.

**Code:**
- `open.Open` returns an error; Windows uses `rundll32 url.dll,FileProtocolHandler`; non-`http(s)` URLs rejected before reaching a platform launcher.
- New `internal/auth.BrowserFlow` replaces the two duplicate `localhost:51164` callback servers (`internal/auth` + `cmd/templates/setup`). Binds *before* opening the browser so a fast redirect can't beat the listener; bounds the wait; drops stray callbacks instead of parking handlers.
- `RunBrowserLogin` is the single entry point for interactive login, so `dr auth login` and the implicit login look identical. Wording follows what actually happened — only a browser that really opened gets "Opening your browser…".
- `HostModel` → `tui/hostpicker` with a blocking `Select()`. Real picker on a TTY; plain-text prompt retained for piped stdin, CI, and the expect smoke tests.
- Adds `dr auth login --no-browser` for SSH/headless.

**Two bugs found while tracing this flow:**
- **`--skip-auth` was silently ignored.** `cmd/root.go` bound `"skip-auth"`; three read sites read `"skip_auth"`. Viper doesn't treat `-`/`_` as equivalent for lookups, so only `DATAROBOT_CLI_SKIP_AUTH` worked — a user passing `--skip-auth` on a fresh install was still dragged into this exact auth flow. Now behind `config.SkipAuthKey`. Several tests had enshrined the wrong spelling, which is why it stayed invisible.
- **`drconfig.yaml` was world-readable (`0644`).** `os.Create` sets `0644` and `os.WriteFile`'s perm arg is ignored for an existing file, so the deliberate `0600` in `write.go` never took effect on the plaintext API token. Now created `0600` and chmod'ed on every write to repair existing configs.

`docs/commands/auth.md` was substantially wrong about this flow (port 8080 with 8081/8082 fallback, an OAuth code exchange, an "encrypted" token, nested YAML). Corrected, and documented that the `?key=` handoff has no `state` parameter.

## TESTING

`task lint` 0 issues · full `-race` suite clean · both auth expect scripts pass (set-url 9.6s, login 0.66s).

Verified against the real binary: end-to-end callback (200, success page, token persisted, `0600`), TUI picker over a PTY (arrow-down+Enter persisted EU Cloud), `--no-browser`, non-TTY, and `--skip-auth` — the last confirmed broken on `main` in a throwaway worktree and fixed here.

## NOTES

- **The Windows `rundll32` change is unverified** — can't be exercised on macOS. Needs the Windows smoke workflow or a manual run. Per `AGENTS.md`, if that workflow's composite action needs touching, `@main` pinning means this PR's own CI won't exercise it.
- **The protocol is still unhardened, deliberately.** The API key arrives as a GET query param with no `state` parameter, so any local process reachable at `localhost:51164` during a login could deliver a key. Documented but left alone — deserves a security-reviewed ticket, not a ride-along on a UX fix.
- One rendering trap worth knowing: styling text that lipgloss then wraps in a border makes it re-emit content one character at a time, which broke the smoke test's `cliRedirect=true` match. The URL is now intentionally unstyled, with a regression test.

Before: 
image

After:
image

image

---

> [!NOTE]
> <sup>[Cursor Bugbot](https://cursor.com/bugbot) is generating a summary for commit b819b707a3bdc763011650f5c15e4d33924f87c0. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/internal/auth/browserflow.go
+++ b/internal/auth/browserflow.go
@@ -1,0 +1,317 @@
+// Copyright 2026 DataRobot, Inc. and its affiliates.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package auth
+
+import (
+	"context"
+	"errors"
+	"fmt"
+	"net"
+	"net/http"
+	"os"
+	"sync"
+	"time"
+
+	"github.com/datarobot/cli/internal/assets"
+	"github.com/datarobot/cli/internal/log"
+	"github.com/datarobot/cli/internal/misc/open"
+	"github.com/datarobot/cli/internal/misc/reader"
+	"github.com/datarobot/cli/tui"
+)
+
+// CallbackAddr is the address the DataRobot web app redirects back to after the
+// user authorizes the CLI. The port is part of the contract with the web app's
+// cliRedirect handler, so it cannot be changed unilaterally and has no fallback.
+const CallbackAddr = "localhost:51164"
+
+// DefaultLoginTimeout bounds how long the CLI waits for the browser callback.
+// Generous enough to cover an SSO round trip, short enough that a browser which
+// never opened does not hang the terminal indefinitely.
+const DefaultLoginTimeout = 5 * time.Minute
+
+// ErrLoginInterrupted is returned when the user aborts the login (Ctrl-C) or
+// another CLI process takes over the callback port.
+var ErrLoginInterrupted = errors.New("login was interrupted")
+
+// BrowserFlow owns the local HTTP listener that receives the API key after the
+// user authorizes the CLI in their browser.
+//
+// The three steps are deliberately separate so that both the blocking CLI path
+// and the bubbletea setup wizard can drive them from their own control flow:
+//
+//	flow, err := NewBrowserFlow(host)  // binds the listener
+//	defer flow.Close()
+//	openErr := flow.OpenBrowser()      // best effort; report the link if non-nil
+//	key, err := flow.Wait(ctx)         // serves until the callback arrives
+//
+// Binding before opening the browser matters: the listener must already exist
+// when the browser follows the redirect, otherwise a fast redirect races the
+// CLI and the callback is refused.
+type BrowserFlow struct {
+	authURL  string
+	listener net.Listener
+	server   *http.Server
+	keyCh    chan string
+	timeout  time.Duration
+
+	closeOnce sync.Once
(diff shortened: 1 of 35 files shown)


tui/hostpicker/hostpicker.go
ajalon1, line 1: Sadface.

This is worth a mention in AGENTS.md and in dev docs.

ajalon1 (commented):
Huh, and here I am thinking this was a transient or id10t error. Reviewing.

internal/auth/browserflow.go
```diff
@@ -0,0 +1,334 @@
...
+// Close shuts the callback server down. It is safe to call more than once.
+func (f *BrowserFlow) Close() error {
+	f.closeOnce.Do(func() {
+		// A fresh context: the login context is typically already cancelled by the
+		// time we get here (Ctrl-C), and Shutdown on a cancelled context would skip
+		// draining the in-flight response that renders the success page.
+		ctx, cancel := context.WithTimeout(context.Background(), 5*time.Second)
+		defer cancel()
+
+		if err := f.server.Shutdown(ctx); err != nil {
```
ajalon1, line 171: 🤖 : Optional, not requesting a change: if `Close()` runs before `Wait()` ever calls `Serve()` — the Esc-before-login-starts path in `LoginModel.Close()` — `f.server.Shutdown` has nothing tracked to close yet, because `http.Server` only tracks a listener once `Serve` has been called on it. The raw `net.Listener` from `listenReclaimingPort` is never released in that path.

Verified empirically in a scratch test: constructing a flow, calling `Close()` immediately without ever calling `Wait()`, then trying to `net.Listen` on the same address again — it fails with "address already in use".

Low real-world impact since the CLI process exits shortly after anyway, but worth either having `Close()` explicitly close `f.listener` itself, or documenting on `NewBrowserFlow`'s godoc that `Wait` must run at least once for cleanup to be complete.

ajalon1 (approved):
LGTM, you can tackle Cursor bot reviews if you want

PR #725, opened 2026-07-31
[CFX-6904] dotenv setup always show Pulumi config screen

taras-pokornyy:
## RATIONALE

The Harness "Agentic app E2E tests" Windows pipeline has a "Run DR CLI tests" stage that drives `dr dotenv setup` with scripted keystrokes, expecting the "Pulumi Configuration Passphrase" screen to be the very first thing shown (it answers `n` to skip). The stage was flaky: some runs never showed that screen at all — instead the CLI's "Environment Variables Menu" was rendered/re-rendered repeatedly, so the scripted input went to the wrong screen and the stage failed (confirmed via Harness execution `o8jh7ZZoThijkiydUVvpTg`, run #379).

Root cause: `dotenv.Model` had two screen fields — `screen` (what `Update()`/`View()` actually read) and `initialScreen` (set by callers, but only ever consulted inside `Init()` to pick an async command). Since `screen` was left at its Go zero value (`listScreen`, i.e. "Environment Variables Menu"), and bubbletea renders the first frame from the model as originally constructed — before any async `Init()` command resolves — every `dr dotenv setup` briefly showed the wrong screen, regardless of intent. The decision to show the Pulumi screen was *also* made only asynchronously, inside the `promptsLoadedMsg` handler, racing a `filepath.Walk`-based prompt-discovery call that's slower and less predictable on Windows CI. When that load was slow, or errored, the model got stuck re-rendering "Environment Variables Menu" forever — errors from it were captured but never rendered anywhere.

A second caller with the identical bug was found during the fix: `cmd/templates/setup/model.go` embeds a `dotenv.Model` and drives the same Pulumi-screen decision via the same async path.

## CHANGES

- Collapsed `Model.screen`/`initialScreen` into a single field, always set explicitly at construction time (`cmd/dotenv/cmd.go`, `cmd/dotenv/helpers.go`) instead of relying on the `listScreen` zero-value + an async transition.
- Added an exported `Model.ConfigureFromPulumiCheck(needsPulumi, loggedIn, needsPassphrase bool)` method (`cmd/dotenv/pulumiModel.go`) that decides synchronously — right after `CheckPulumiSetup` runs — whether to start on the Pulumi screen, instead of deferring that decision to the async `promptsLoadedMsg` handler. This guarantees the passphrase screen, when needed, is the very first frame rendered, with no dependency on the prompt-discovery walk completing first.
- Updated both callers to use it: `cmd/dotenv/cmd.go`'s `SetupCmd`, and the embedded flow in `cmd/templates/setup/model.go` (which had the same latent race).
- Moved `errMsg` handling to the top level of `Update()` so an error is captured regardless of which screen is currently active (previously only handled while on `listScreen`), and now renders as a visible error banner in `View()` instead of being silently dropped.
- Removed the now-dead `NeedsPulumiLogin`/`PulumiAlreadyLoggedIn`/`NeedsPulumiPassphrase` exported fields on `Model`, superseded by `ConfigureFromPulumiCheck`.
- Added regression tests (`cmd/dotenv/model_test.go`): one asserting the Pulumi passphrase screen is always the first thing rendered (never "Environment Variables Menu"), one asserting an error is surfaced regardless of the active screen. Both were verified to fail (compile error) against the pre-fix code.

# RATIONALE

## PR Automation

> [!IMPORTANT]
>

---

> [!NOTE]
> **Medium Risk**
> Changes first-frame TUI behavior for setup and template flows; risk is mitigated by focused regression tests but still affects interactive and scripted automation paths.
> 
> **Overview**
> Fixes flaky `dr dotenv setup` / template setup where the **Environment Variables Menu** could appear before the **Pulumi Configuration Passphrase** screen, breaking scripted CI input.
> 
> **Screen selection** no longer relies on `listScreen` as the zero value plus an async handoff after prompt loading. Callers set a single `screen` up front, and **`ConfigureFromPulumiCheck`** runs right after `CheckPulumiSetup` so the Pulumi flow is the first frame when needed (`--yes` still skips it). The old `NeedsPulumiLogin` / related exported fields are removed in favor of that helper; `cmd/templates/setup` uses the same path for embedded dotenv.
> 
> **Errors:** `errMsg` is handled at the top of `Update()` and shown in `View()` on any screen, so prompt-load failures are not dropped while on the wizard.
> 
> Regression tests cover “Pulumi first, never the list menu” and error display on non-list screens.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit f5f1736d9c60fc3882b482e4007ab99c39a9b252. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/cmd/dotenv/model.go
+++ b/cmd/dotenv/model.go
@@ -299,6 +299,16 @@ func (m Model) handlePulumiUpdate(msg tea.Msg) (tea.Model, tea.Cmd) {
 }
 
 func (m Model) Update(msg tea.Msg) (tea.Model, tea.Cmd) { //nolint: cyclop
+	// Always capture window size on the outer model, even while a sub-model
+	// (e.g. the Pulumi flow) is active — otherwise width/height never get set
+	// when Init() starts directly on a screen with an active sub-model, and
+	// stay stuck at zero for the rest of the session (no SIGWINCH on Windows
+	// means no later WindowSizeMsg will ever arrive to fix it up).
+	if sizeMsg, ok := msg.(tea.WindowSizeMsg); ok {
+		m.width = sizeMsg.Width
+		m.height = sizeMsg.Height
+	}
+
 	// If Pulumi login sub-model is active, delegate to it
 	if m.pulumiModel != nil {
 		return m.handlePulumiUpdate(msg)
--- a/cmd/dotenv/model.go
+++ b/cmd/dotenv/model.go
@@ -42,7 +42,12 @@ const (
 type screens int
 
 const (
-	listScreen = screens(iota)
+	// unsetScreen is the zero value: a Model whose screen was never explicitly
+	// configured. Init() panics on it rather than silently behaving as
+	// listScreen, so a caller that forgets to set screen fails immediately
+	// instead of shipping a hard-to-diagnose "wrong screen shown" bug.
+	unsetScreen = screens(iota)
+	listScreen
 	editorScreen
 	wizardScreen
 	pulumiScreen
@@ -263,6 +268,8 @@ func (m Model) autoPopulateAndSave() (tea.Model, tea.Cmd) {
 
 func (m Model) Init() tea.Cmd {
 	switch m.screen {
+	case unsetScreen:
+		panic("dotenv.Model.screen was never configured before Init()")
 	case editorScreen:
 		return tea.Batch(openEditorCmd, tea.WindowSize())
 	case pulumiScreen:
@@ -275,7 +282,7 @@ func (m Model) Init() tea.Cmd {
 		return tea.Batch(m.loadVariables(), tea.WindowSize())
 	}
 
-	return tea.Batch(m.loadVariables(), tea.WindowSize())
+	panic(fmt.Sprintf("dotenv.Model.screen has unknown value %d", m.screen))
 }
 
 func (m Model) handlePulumiUpdate(msg tea.Msg) (tea.Model, tea.Cmd) {
@@ -395,6 +402,8 @@ func (m Model) Update(msg tea.Msg) (tea.Model, tea.Cmd) { //nolint: cyclop
 	}
 
 	switch m.screen {
+	case unsetScreen:
+		// Unreachable: Init() panics before Update() ever sees an unconfigured Model.
 	case pulumiScreen:
 		// Handled above via m.pulumiModel delegation
 	case listScreen:
@@ -480,6 +489,8 @@ func (m Model) View() string {
 	}
 
 	switch m.screen {
+	case unsetScreen:
+		// Unreachable: Init() panics before View() ever sees an unconfigured Model.
 	case pulumiScreen:
 		if m.pulumiModel != nil {
 			sb.WriteString(m.pulumiModel.View())
--- a/cmd/dotenv/model.go
+++ b/cmd/dotenv/model.go
@@ -49,27 +49,23 @@ const (
 )
 
 type Model struct {
-	screen                screens
-	initialScreen         screens
-	DotenvFile            string
-	variables             []envbuilder.Variable
-	err                   error
-	textarea              textarea.Model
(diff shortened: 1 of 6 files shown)


cmd/dotenv/model.go
```diff
@@ -266,12 +262,17 @@ func (m Model) autoPopulateAndSave() (tea.Model, tea.Cmd) {
...
-
-	if m.initialScreen == wizardScreen {
+	case pulumiScreen:
+		// Prompts are (re)loaded once the Pulumi flow completes and hands off
+		// to the wizard (see handlePulumiUpdate) — no need to load them now.
+		return tea.Batch(m.pulumiModel.Init(), tea.WindowSize())
+	case wizardScreen:
 		return tea.Batch(m.loadPrompts(), tea.WindowSize())
+	case listScreen:
+		return tea.Batch(m.loadVariables(), tea.WindowSize())
```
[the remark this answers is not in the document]
taras-pokornyy: Fixed in commit [5a8552d](https://github.com/datarobot-oss/cli/pull/725/commits/5a8552df70a35daca2105eb74680491e6afd6fd9)

ajalon1:
Nice find Taras

cmd/dotenv/pulumiModel.go
```diff
@@ -397,6 +397,24 @@ func CheckPulumiSetup(dir string, variables []envbuilder.Variable) (needsSetup,
 	return needsPulumiSetup(prompts, loggedIn, passphraseSet), loggedIn, !passphraseSet
 }
 
+// ConfigureFromPulumiCheck sets the model's initial screen synchronously from the
+// result of CheckPulumiSetup, so the Pulumi passphrase screen — when needed — is the
+// very first thing rendered. Deciding this asynchronously (e.g. only once the wizard's
+// prompt-loading completes) races with that load and can leave the wrong screen showing
+// if it's slow or fails. Call this after constructing the Model (with Yes already set,
+// since --yes mode always skips the interactive Pulumi screen) and before Init().
```
ajalon1: Is there a way to enforce this ordering? I'm at the point where having to tell people how best to use a function is sort of a smell.

cmd/dotenv/model.go
```diff
@@ -49,27 +49,23 @@ const (
 )
```
ajalon1, line 54:
```suggestion
const (
    unsetScreen = screens(iota)
	listScreen
	editorScreen
	wizardScreen
	pulumiScreen
)
```

cmd/dotenv/model.go
```diff
@@ -266,12 +262,17 @@ func (m Model) autoPopulateAndSave() (tea.Model, tea.Cmd) {
 }
 
 func (m Model) Init() tea.Cmd {
-	if m.initialScreen == editorScreen {
+	switch m.screen {
+	case editorScreen:
```
ajalon1, line 273:
```suggestion
	switch m.screen {
	case unsetScreen:
	   panic("You didn't configure this")
	case editorScreen:
```

ajalon1 (approved):
Seriously, great find. Great fixes. I'm worried now that this "set model, update view later" async bug lives elsewhere in the TUI code, but at least it's fixed for `dr dotenv ...`.

My only suggestion is given in comments below, an unsetScreen sentinel to ensure that we always set it correctly if we expand the cmd/dotenv module.

cmd/dotenv/model.go
taras-pokornyy: Addressed in [b4dac06](https://github.com/datarobot-oss/cli/pull/725/commits/b4dac06d74b7f322e379f43ef26f3a30f3445f46) : added an `unsetScreen` sentinel as the zero value (per your suggested diff) and `Init()` now panics if `screen` is still unset when it runs — so a caller that forgets to configure the model fails immediately instead of silently behaving like `listScreen`.

PR #726, opened 2026-08-02
[APP-6736] implements `dr auth export` to emit canonical DataRobot env vars

burkestar:
## RATIONALE

Adds `dr auth export`, which prints the credentials the CLI is currently
using as shell statements that set `DATAROBOT_ENDPOINT` and
`DATAROBOT_API_TOKEN`, so they can be sourced straight into the session:

```
eval "$(dr auth export)"
```

This makes it easier to use `dr auth login` with DataRobot SDKs that expect environment variables.

## NOTES

Only the statements go to stdout — errors and hints go to stderr — so the
output is always safe to evaluate, and a missing-credentials run leaves
stdout empty and exits non-zero. Syntax follows the detected parent shell
(bash/zsh, fish, PowerShell, cmd.exe) and can be forced with `--shell`;
`--output-format json` emits the pair as an envelope instead.

Credentials resolve the same way the rest of the CLI resolves them — a
complete `DATAROBOT_ENDPOINT`/`DATAROBOT_API_TOKEN` env pair, else the config
file from `dr auth login` — with the endpoint normalized to its canonical
`/api/v2` form. Unlike other commands it never calls the API or starts a
login flow, so it is usable from shell startup files.

Covered by unit tests and by smoke tests on both bash and PowerShell that
assert the output actually populates the shell when sourced.

Co-Authored-By: Claude Opus 5 (1M context) <noreply@anthropic.com>

### Resolved race condition in tests

This was a flake in the test harness, not in the dotenv model or the CFX-6904 change.

Root cause — teatest.TestModel runs the program on its own goroutine and, on exit, does:

tm.doneCh <- true   // FinalModel's wait unblocks here
tm.modelCh <- m     // ...but the model is published only after

TestModel.FinalModel waits on doneCh, then does a non-blocking receive on modelCh, falling back to tm.model — which is nil until the program finishes. A caller that lands between those two sends gets nil, and nil.(Model) fails with exactly Final model is not of type Model. The window is sub-microsecond, so it only opens under a loaded/shuffled CI run — and it applies to every test using the shared helper, not just the window-size one.

Fix — added internal/testutil/teatest.go with a FinalModel helper that re-calls TestModel.FinalModel until the model is published (its wait is sync.Once-guarded, so retries cost nothing), with a 5s bound and a clear t.Fatalf if it never arrives. Routed the dotenv suite helper (cmd/dotenv/model_test.go:171) and tui/logo_animation_test.go:227, which had the same latent race, through it.

Verification — the failure reproduces on this machine only under contention (6 concurrent go test -race invocations). Running that stress with temporary instrumentation in the helper caught teatest returning a nil model on one run — in TestDotenvModel_Branching_Path, confirming it isn't specific to the test CI happened to hit — which the retry absorbed, with zero failures. After removing the instrumentation: task lint clean, and ./cmd/dotenv ./tui pass under -race -shuffle 1785843468118572986 -count=2.

---

> [!NOTE]
> **Medium Risk**
> The command writes API tokens to stdout in plain text (documented), but it only reads existing CLI/env credentials and does not change auth storage or network behavior.
> 
> **Overview**
> Adds **`dr auth export`**, a new `auth` subcommand that prints **`DATAROBOT_ENDPOINT`** and **`DATAROBOT_API_TOKEN`** as shell assignments so users can run `eval "$(dr auth export)"` (or equivalent on fish/PowerShell/cmd).
> 
> Credentials follow the same precedence as the rest of the CLI (full env pair first, else login config), with endpoints normalized to canonical `/api/v2` form. The command **does not** call the API or start OAuth; on failure it writes hints to **stderr** and leaves **stdout** empty for safe `eval`. Output supports **`--shell`**, auto-detected shell syntax, and **`--output-format json`**.
> 
> The `auth` command tree and user/docs are updated; **unit tests** cover quoting, resolution, and CLI behavior; **Unix and Windows smoke tests** verify emitted variables and that sourcing/`Invoke-Expression` actually sets the env.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 49a7214043cfd34f145423fd053e7246ca413f3b. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/cmd/auth/export/cmd.go
+++ b/cmd/auth/export/cmd.go
@@ -29,6 +29,7 @@ import (
 	"github.com/datarobot/cli/internal/cli"
 	"github.com/datarobot/cli/internal/config"
 	"github.com/datarobot/cli/internal/config/viperx"
+	"github.com/datarobot/cli/internal/log"
 	"github.com/datarobot/cli/internal/outputformat"
 	internalShell "github.com/datarobot/cli/internal/shell"
 	"github.com/datarobot/cli/internal/version"
@@ -248,8 +249,13 @@ func resolveShell(requested string) (internalShell.Shell, error) {
 
 	detected, err := internalShell.DetectShell()
 	if err != nil {
+		log.Debug("auth export: shell detection failed, falling back to POSIX syntax",
+			"error", err, "shell", internalShell.Bash)
+
 		return internalShell.Bash, nil
 	}
+
+	log.Debug("auth export: detected shell", "shell", detected)
 
 	return internalShell.Shell(detected), nil
 }
@@ -305,6 +311,9 @@ directly by your shell. Errors and hints go to stderr.
 Credentials are resolved the same way the rest of the CLI resolves them:
   • the ` + endpointVar + ` and ` + tokenVar + ` environment variables, if both are set
   • otherwise the CLI config file written by '` + version.CliName + ` auth login'
+
+A project's .env file is never consulted — this exports the CLI's own
+credentials. Use '` + version.CliName + ` dotenv setup' to manage .env files.
 
 Unlike most commands, this one never starts a login flow and never calls the
 API, so it is safe to run from a shell startup file. Run '` + version.CliName + ` auth check' first
--- a/cmd/auth/export/cmd.go
+++ b/cmd/auth/export/cmd.go
@@ -1,0 +1,346 @@
+// Copyright 2026 DataRobot, Inc. and its affiliates.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// Package export renders the DataRobot credentials the CLI is currently using
+// as shell statements that set the canonical DataRobot environment variables,
+// so they can be sourced into the caller's shell session.
+package export
+
+import (
+	"errors"
+	"fmt"
+	"io"
+	"net/url"
+	"slices"
+	"strings"
+
+	"github.com/datarobot/cli/internal/auth"
+	"github.com/datarobot/cli/internal/cli"
+	"github.com/datarobot/cli/internal/config"
+	"github.com/datarobot/cli/internal/config/viperx"
+	"github.com/datarobot/cli/internal/outputformat"
+	internalShell "github.com/datarobot/cli/internal/shell"
+	"github.com/datarobot/cli/internal/version"
+	"github.com/datarobot/cli/tui"
+	"github.com/spf13/cobra"
+)
+
+// The canonical DataRobot environment variables. Every DataRobot SDK, the
+// platform itself, and this CLI read this pair, so they are what we emit.
+const (
+	endpointVar = "DATAROBOT_ENDPOINT"
+	tokenVar    = "DATAROBOT_API_TOKEN"
+)
+
+// jsonEnvelopeKey is the top-level key used for `--output-format json`.
+const jsonEnvelopeKey = "environment"
+
+var (
+	errNoCredentials = errors.New("no DataRobot credentials configured")
+	errNoEndpoint    = errors.New("no DataRobot endpoint configured")
+	errNoToken       = errors.New("no DataRobot API token configured")
+)
+
+// credentials is the endpoint/token pair this command renders.
+type credentials struct {
+	Endpoint string
+	Token    string
+}
+
+// statementFunc renders a single "set this variable" statement for one shell.
+type statementFunc func(name, value string) string
+
+// canonicalEndpoint normalizes a configured DataRobot URL into the canonical
+// API endpoint form, e.g. "app.datarobot.com" becomes
+// "https://app.datarobot.com/api/v2". A URL that already carries a path is
+// preserved as-is (minus any trailing slash, query, or fragment) so that
+// self-managed installs serving the API under a custom prefix keep working.
+func canonicalEndpoint(raw string) (string, error) {
+	trimmed := strings.TrimSpace(raw)
+	if trimmed == "" {
+		return "", errNoEndpoint
+	}
+
+	// Default to https so a bare hostname is accepted, matching
+	// config.SchemeHostOnly.
+	if !strings.Contains(trimmed, "://") {
+		trimmed = "https://" + trimmed
+	}
+
+	parsed, err := url.Parse(trimmed)
+	if err != nil {
+		return "", err
+	}
+
+	if parsed.Host == "" {
+		return "", config.ErrInvalidURL
+	}
+
+	path := strings.TrimRight(parsed.Path, "/")
+	if path == "" {
+		path = config.DRAPIURLSuffix
+	}
+
+	parsed.Path, parsed.RawQuery, parsed.Fragment = path, "", ""
+
+	return parsed.String(), nil
+}
+
+// rawCredentials returns the credentials the CLI would use, mirroring the
+// precedence in auth.EnsureAuthenticated: a complete pair of environment
+// variables wins over the CLI config file. Unlike EnsureAuthenticated, it never
+// contacts the API and never starts a login flow, so `dr auth export` stays
+// usable from shell startup files and non-interactive sessions.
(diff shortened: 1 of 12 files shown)


ajalon1:
Love it @burkestar -- I think this is something I've implemented ad hoc in like every environment we have and every language we support. Will take a look shortly

ajalon1:
Please create a JIRA ticket for this as well so we can track it for later.

cmd/auth/export/cmd.go
```diff
@@ -0,0 +1,346 @@
...
+	"github.com/datarobot/cli/tui"
+	"github.com/spf13/cobra"
+)
+
+// The canonical DataRobot environment variables. Every DataRobot SDK, the
+// platform itself, and this CLI read this pair, so they are what we emit.
+const (
+	endpointVar = "DATAROBOT_ENDPOINT"
+	tokenVar    = "DATAROBOT_API_TOKEN"
+)
```
ajalon1, line 45: I think these should actually live in `internal/auth`, and I'm surprised they don't already exist there, or are already defined as consts.

Feel free to move to internal/auth, or can be done in a followup.

cmd/auth/export/cmd.go
```diff
@@ -0,0 +1,346 @@
...
+
+// jsonEnvelopeKey is the top-level key used for `--output-format json`.
+const jsonEnvelopeKey = "environment"
```
ajalon1, line 48: Works for me

cmd/auth/export/cmd.go
```diff
@@ -0,0 +1,346 @@
...
+
+  ` + endpointVar + `=https://<your-install>/api/v2
+  ` + tokenVar + `=<token>
+
+Only the shell statements are written to stdout, so the output can be evaluated
+directly by your shell. Errors and hints go to stderr.
+
+Credentials are resolved the same way the rest of the CLI resolves them:
+  • the ` + endpointVar + ` and ` + tokenVar + ` environment variables, if both are set
+  • otherwise the CLI config file written by '` + version.CliName + ` auth login'
```
ajalon1, line 313: I might add something here along the lines of `A project's .env file is never consulted; this exports the CLI's own credentials. Use 'dr dotenv setup' to manage .env files. Not blocking.

docs/commands/auth.md
```diff
@@ -153,6 +153,80 @@ Unset it and try again:
...
+```
+
+The output syntax is chosen from the detected parent shell. Use `--shell` to override it — useful when generating a file for a different shell, or when detection fails (unrecognized shells fall back to POSIX `export` syntax).
+
+**Where credentials come from:**
+
+1. `DATAROBOT_ENDPOINT` (or `DATAROBOT_API_ENDPOINT`) and `DATAROBOT_API_TOKEN`, if both are set
+2. Otherwise the CLI config file written by `dr auth login`
+
+The endpoint is normalized to its canonical `/api/v2` form, so a config or environment value of `app.datarobot.com` is exported as `https://app.datarobot.com/api/v2`. A URL that already has a path (self-managed installs serving the API under a custom prefix) is left alone.
```
ajalon1, line 200: I might add something here along the lines of `A project's .env file is never consulted; this exports the CLI's own credentials. Use 'dr dotenv setup' to manage .env files. Not blocking.

docs/commands/auth.md
```diff
@@ -153,6 +153,80 @@ Unset it and try again:
...
+**Machine-readable output:**
+
+```bash
+$ dr auth export --output-format json
+{
+  "environment": {
+    "DATAROBOT_API_TOKEN": "<token>",
+    "DATAROBOT_ENDPOINT": "https://app.datarobot.com/api/v2"
+  }
+}
```
ajalon1, line 213: I would love to see an extension of the envelope here that renders the source of these vars, something like:

```json
{
  "environment": {
    "DATAROBOT_API_TOKEN": "...",
    "DATAROBOT_ENDPOINT": "https://app.datarobot.com/api/v2"
  },
  "source": "<environment|config>"
}
```

where environment = env var pairs or config = drconfig.yaml. Could be ~~hidden behind the `--debug` flag, or just~~ always populated in `--output-format json`. This has been a common issue for us.

cmd/auth/export/cmd.go
```diff
@@ -0,0 +1,346 @@
...
+// metacharacters have to be escaped individually.
+func cmdEscape(value string) string {
+	return strings.NewReplacer(
+		"^", "^^",
+		"&", "^&",
+		"<", "^<",
+		">", "^>",
+		"|", "^|",
+	).Replace(value)
+}
```
ajalon1, line 192: Interesting case of one of these things is not like the others.

cmd/auth/export/cmd.go
```diff
@@ -0,0 +1,346 @@
...
+		statement = powerShellStatement
+	case internalShell.Cmd:
+		statement = cmdStatement
+	case internalShell.Bash, internalShell.Zsh:
+		statement = posixStatement
+	default:
+		// Unrecognized shells (e.g. plain sh, dash, ksh) get POSIX syntax,
+		// which is correct for every Bourne-compatible shell.
+		statement = posixStatement
+	}
```
ajalon1, line 228: 👍🏽

cmd/auth/export/cmd.go
```diff
@@ -0,0 +1,346 @@
...
+	return []string{
+		statement(endpointVar, creds.Endpoint),
+		statement(tokenVar, creds.Token),
+	}
+}
+
+// resolveShell picks the syntax to emit. An explicit --shell wins; otherwise
+// the parent shell is detected. Detection failures fall back to POSIX syntax
+// instead of erroring, since that is the right answer for sh, bash, and zsh.
+func resolveShell(requested string) (internalShell.Shell, error) {
```
ajalon1, line 239: Was confused at first by this due to internal/shel.ResolveShell(), but I noticed that the other helper prints Detected Shell to stdout, so ... that's fine.

cmd/auth/export/cmd.go
ajalon1, line 1: Lot of the work in here is good improvements on top of internal/shell. I've added https://datarobot.atlassian.net/browse/CFX-7383 to handle reusing some of this throughout the codebase

cmd/auth/export/cmd.go
```diff
@@ -0,0 +1,346 @@
...
+
+	detected, err := internalShell.DetectShell()
+	if err != nil {
+		return internalShell.Bash, nil
+	}
+
+	return internalShell.Shell(detected), nil
+}
+
+func RunE(cmd *cobra.Command, _ []string) error {
```
ajalon1, line 263: Noticed no telemetry. Works for me.

smoke_test_scripts/windows/run_smoke_test.ps1
ajalon1, line 1: Wonder if this also becomes different for PowerShell 5.1 and 7.

ETA: Droid tells me "it does not actually become different".

cmd/auth/export/cmd.go
```diff
@@ -0,0 +1,346 @@
...
+				requested, strings.Join(internalShell.SupportedShells(), ", "))
+		}
+
+		return internalShell.Shell(normalized), nil
+	}
+
+	detected, err := internalShell.DetectShell()
+	if err != nil {
+		return internalShell.Bash, nil
+	}
```
ajalon1, line 256: I'd add a log.Debug() statement here for `dr auth export --debug`

ajalon1 (approved):
Very nice add. The command itself is helpful but there are a lot of little nice things in the underlying code I could see being reused in other parts of the codebase.

I first thought this was already handled by `dr self config` (which I wrote to help with debugging and bug reports) but on reflection it makes sense that this is the one you use for active "give me real values to pipe into another script", so it should be different.

Thanks for handling all the little things -- JSON output, completions, good command descriptions, updating the docs, even smoke test scripts.

cmd/auth/export/cmd.go
burkestar: I'll leave some of these consolidation changes to you in the follow-up task you have planned.

cmd/auth/export/cmd.go
burkestar: Added debug log

docs/commands/auth.md
burkestar: Leaving this one for you :)

cmd/dotenv/model_test.go
```diff
@@ -165,7 +166,9 @@ func (suite *DotenvModelTestSuite) FinalModel(tm *teatest.TestModel) Model {
 		suite.T().Error(err)
 	}
 
-	finalModel := tm.FinalModel(suite.T())
+	// testutil.FinalModel, not tm.FinalModel — see its doc comment for the
```
burkestar, line 169: fixed a race condition in teatest that this PR hit

PR #712, opened 2026-07-24
[CFX-7138] fix(lint): resolve Windows-only lint violations found via GOOS cross-lint

ajalon1:
# RATIONALE

Linters such as `godot` and `perfsprint` skip platform-specific files (`*_windows.go`) when run on a different OS. CI only runs `golangci-lint` on `ubuntu-latest`, and local macOS dev never analyzes Windows-only files. Running `GOOS=windows golangci-lint run` surfaced two violations that were invisible to both.

Tracked in [CFX-7138](https://datarobot.atlassian.net/browse/CFX-7138).

## CHANGES

- **`internal/telemetry/os_windows.go`**: Added trailing period to `// example output` comment (godot).
- **`internal/tls/tls_windows.go`**: Replaced `fmt.Errorf` with no formatting verbs with `errors.New` (perfsprint).

## TESTING

- `GOOS=windows golangci-lint run --max-issues-per-linter=0 --max-same-issues=0 ./...` passes with 0 issues
- `golangci-lint run ./...` (macOS) passes with 0 issues
- `GOOS=windows go build ./...` passes
- `go build ./...` passes
- `go test ./internal/telemetry/...` passes

## NOTES

These violations were found by running `GOOS=windows golangci-lint run ./...` on macOS. The broader fix to prevent this class of issue (running lint across all target platforms in CI) is tracked in CFX-7138.

## RELATED

- CFX-7138: Linters skip platform-specific files (build tags) when run on a different OS
- PR #710: fix(lint): enable godot and add periods to comments

---

> [!NOTE]
> **Low Risk**
> Comment and error-construction style only; no runtime or security behavior change.
> 
> **Overview**
> Fixes **godot** and **perfsprint** issues in Windows build-tagged files that only show up when lint runs with `GOOS=windows` (they are skipped on Linux CI and local non-Windows runs).
> 
> In `internal/telemetry/os_windows.go`, the `// example output` comment now ends with a period. In `internal/tls/tls_windows.go`, the empty-cert-store error uses `errors.New` instead of `fmt.Errorf` with no format verbs, with `errors` added to imports.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit bf43adfc5d11aaac18f222388b77509313e8f7c3. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

PR #731, opened 2026-08-04
[CFX-7138] fix(lint): run golangci-lint once per target GOOS

cdevent:
# RATIONALE

`golangci-lint` only analyzes files selected by the **build constraints of the platform it runs on**. This repo lints exactly once, on the host GOOS:

- `Taskfile.yaml` — `golangci-lint run ./...`, no `GOOS` set
- `.golangci.yaml` — no `build-tags` key
- `.github/workflows/pr-checks.yaml` — the `lint` job is `ubuntu-latest` only

So of the 12 build-constrained files in this repo, a macOS dev lints the darwin/`!windows` set, CI lints the linux/`!windows` set, and the six **windows-only files are linted by nobody**.

That gap has already shipped real violations:

- `internal/telemetry/os_darwin.go` was fixed by #710 (linted on macOS) while its sibling `internal/telemetry/os_linux.go` was missed and needed a manual follow-up (`471ae910`).
- #712 later found two more (`os_windows.go` godot, `tls_windows.go` perfsprint) that only surfaced because someone ran `GOOS=windows golangci-lint run` by hand.

This makes that class of escape impossible rather than relying on someone remembering to check manually.

## CHANGES

**`Taskfile.yaml`**
- New `LINT_GOOS_TARGETS: "linux darwin windows"` var (kept in sync with the `goos` lists in `goreleaser.yaml`).
- New internal `lint-goos` task that loops those targets; callers pass `LINT_ARGS` to select the mode.
- `lint` and `delint` route their `golangci-lint run` through it. **`delint` matters as much as `lint`** — without it a macOS dev cannot `--fix` a `*_windows.go` violation at all.
- **`precommit` deliberately stays host-GOOS only.** The hook runs on every commit and the loop added ~4-6s of interactive latency there for little benefit — `task lint` and CI already enforce every target, so a cross-platform violation is still caught before merge.

Left **single-run** because they are GOOS-independent: `go mod tidy`, `goreleaser check`, `golangci-lint config verify`, and `golangci-lint fmt`. The formatter was checked empirically rather than assumed — it walks files directly and *does* see `tls_windows.go` from macOS, so it needs no loop.

**`.github/workflows/pr-checks.yaml`**
- No runner matrix. The single `ubuntu-latest` lint job already just calls `task lint`, so it now covers all three targets — same reasoning as the `build` job's *"cross-compile all platforms on a single Ubuntu runner (cheaper than 3-OS matrix)"*. A 3-OS lint matrix would cost macOS/Windows runner minutes, require reworking Linux-only assumptions in that job (`grep -oP`, `~/.cache/golangci-lint`, `tmp/bin/golangci-lint` without `.exe`), and still not help local `task lint`.
- Closed a second-order gap: the `lint` job was gated on `go_code` only, so a PR touching only `Taskfile.yaml`/`.golangci.yaml` (**like this one**) would never re-run lint. Added a `lint_config` paths-filter mirroring the existing `test_config` pattern from CFX-6919.
- Added `Taskfile.yaml` to the golangci cache key so changing `LINT_GOOS_TARGETS` invalidates a cache built for a different target set.

**Docs** — new "Cross-Platform Linting" section in `AGENTS.md` (incl. *"when adding a new release platform, add it to `LINT_GOOS_TARGETS`"*), plus a pointer in `CONTRIBUTING.md` so contributors understand why `task lint` can fail on a file their own OS never compiles.

## TESTING

Rather than just confirming the tree is green, I re-introduced both historical bugs from the ticket and confirmed the old behaviour misses them and the new behaviour catches them:

| Test | Host-only lint (old) | `task lint` (new) |
| --- | --- | --- |
| `os_windows.go` godot — the #712 bug | `0 issues.` | **fails** on the `GOOS=windows` leg |
| `os_linux.go` godot — the `471ae910` bug | `0 issues.` | **fails** on the `GOOS=linux` leg |

- `task delint` fixed the Windows-only file from macOS and left the tree clean — impossible before this change.
- `task lint` on a clean tree: three legs, `0 issues.` each, exit 0.
- `task precommit`: green, and unchanged in cost (host GOOS only, by design).
- `task test`: fully green, no failures.
- `lint-goos` is `internal: true`, so `task --list` is unchanged.

**Cost:** ~1.0s per extra GOOS on a warm cache (each GOOS keeps its own golangci-lint analysis cache), so roughly +2s on `task lint`.

## NOTES

GOARCH is deliberately **not** looped — a sweep found zero arch-constrained files (no `*_amd64.go` / `*_arm64.go` / `*_riscv64.go`, and no `//go:build` line mentions an arch). Only `GOOS` is a real gap today.

All three targets are already clean, so this is purely preventive and should land green.

## RELATED

- CFX-7138
- Follows #710 and #712, which fixed the symptoms this PR addresses the cause of.

---

> [!NOTE]
> **Low Risk**
> Changes are limited to lint tasks, CI gating, and documentation; no application runtime, auth, or data paths are modified.
> 
> **Overview**
> **Cross-platform lint coverage** so build-tagged files (e.g. `*_windows.go`) are checked on every dev machine and in CI, not only on the host OS.
> 
> `Taskfile.yaml` adds `LINT_GOOS_TARGETS` (`linux darwin windows`) and an internal **`lint-goos`** task that runs `GOOS=<target> golangci-lint run` for each target. **`lint`** and **`delint`** now call `lint-goos` instead of a single `golangci-lint run` (with `--fix` passed via `LINT_ARGS`). **`precommit`** keeps its single host-GOOS run to stay fast. GOOS-independent steps (`go mod tidy`, `fmt`, `goreleaser check`) stay single-run.
> 
> **PR checks** mirror the existing `test_config` pattern: a **`lint_config`** paths filter and lint job condition so PRs that only change `Taskfile.yaml`, `.golangci.yaml`, or workflow setup still run lint. The golangci cache key now includes **`Taskfile.yaml`** so changes to `LINT_GOOS_TARGETS` invalidate stale analysis caches. Still one **`ubuntu-latest`** lint job—no OS matrix.
> 
> **`AGENTS.md`** documents cross-platform linting; **`CONTRIBUTING.md`** points contributors at it when running `task lint`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 7f028f6770c1e69c5933602da41a85f3471dfeda. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -77,16 +77,39 @@ func example() {
 - `go mod tidy -diff` - checks if go.mod/go.sum need updates
 - `gofumpt -l -d` - lists formatting issues and shows diffs
 - `go vet` - checks for suspicious constructs
-- `golangci-lint run` - comprehensive linting checks (includes wsl, revive, staticcheck)
+- `golangci-lint run` - comprehensive linting checks (includes wsl, revive, staticcheck), run once per target GOOS
 - `goreleaser check` - validates release configuration
 
 **`task delint`** (auto-fixes):
 - `go mod tidy` - fixes go.mod/go.sum
 - `go fmt` - fixes basic formatting
 - `gofumpt -l -w` - fixes aggressive Go formatting
-- `golangci-lint run --fix` - fixes linting issues where possible
+- `golangci-lint run --fix` - fixes linting issues where possible, run once per target GOOS
 
 All code must pass `task lint` before submitting.
+
+### Cross-Platform Linting
+
+`golangci-lint` only analyzes files selected by the build constraints of the platform
+it runs on. Without help, `*_windows.go` (and any `//go:build` file) would only ever be
+linted on that OS — so a violation in a Windows-only file ships undetected from a macOS
+laptop and from Linux CI alike (CFX-7138).
+
+To close that gap, `lint`, `delint`, and `precommit` all route their `golangci-lint run`
+through the internal `lint-goos` task, which loops over the `LINT_GOOS_TARGETS` variable
+in `Taskfile.yaml` (`linux darwin windows` — kept in sync with the `goos` lists in
+`goreleaser.yaml`). Output is labelled per leg, e.g. `🧹 Linting (GOOS=windows)…`.
+
+Practical notes:
+- Expect `task lint` to fail on a file your own OS never compiles. That is the point.
+  `task delint` can auto-fix those files from any host.
+- The overhead is small (~1s per extra GOOS on a warm cache); each GOOS gets its own
+  golangci-lint analysis cache.
+- **When adding a new release platform, add it to `LINT_GOOS_TARGETS`.**
+- `go mod tidy`, `golangci-lint fmt`, and `goreleaser check` stay single-run — they are
+  GOOS-independent (the formatter walks files directly and is not build-constrained).
+- CI needs no OS matrix for lint: the single `ubuntu-latest` lint job covers all targets
+  because it just calls `task lint`.
 
 ### Git Hooks (Lefthook)
 
@@ -96,7 +119,7 @@ automatically by `task install-tools` and wired up by `task dev-init`.
 Hooks run automatically on `git commit`. Run manually with `task precommit`.
 Bypass with `LEFTHOOK=0 git commit` (use sparingly). Configured hooks:
 
-- **`task precommit`**: formats via gofumpt, verifies Go files are formatted, runs `go mod tidy`, `go vet`, and `golangci-lint run --new-from-rev HEAD` (new changes only), verifies golangci-lint config, and checks go.mod/go.sum are tidy
+- **`task precommit`**: formats via gofumpt, verifies Go files are formatted, runs `go mod tidy`, `go vet`, and `golangci-lint run --new-from-rev HEAD` (new changes only, once per target GOOS), verifies golangci-lint config, and checks go.mod/go.sum are tidy
 - **`task dupcheck`**: duplicate code detection via jscpd (threshold and exclusions in `.jscpd.json`)
 
 Both hooks reuse Taskfile tasks so there is a single source of truth for quality checks.
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -95,10 +95,15 @@ it runs on. Without help, `*_windows.go` (and any `//go:build` file) would only 
 linted on that OS — so a violation in a Windows-only file ships undetected from a macOS
 laptop and from Linux CI alike (CFX-7138).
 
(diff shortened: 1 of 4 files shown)


Taskfile.yaml
```diff
@@ -192,8 +212,9 @@ tasks:
       - go mod tidy
       - echo "🧹 Vetting…"
       - go vet ./...
-      - echo "🧹 Linting new changes…"
-      - "{{.BIN_DIR}}/golangci-lint run --new-from-rev HEAD ./..."
+      - task: lint-goos
+        vars:
+          LINT_ARGS: "--new-from-rev HEAD"
```
ajalon1: Would it be possible to NOT run `lint-goos` in precommit? Keep it in `lint`, `delint`, and in the CI check.

ajalon1 (commented):
So good, thanks for tackling this.

I ran a couple of tests because I felt like some of this will slow down local development, particularly since you changed `task precommit`. Had to figure out where my golangci-lint cache is (~/Library/Caches/golangci-lint), then wiped it.

Run | Time
-- | --
Cold single GOOS lint (cache wiped) | ~4.2s
Warm single GOOS lint (immediately after) | ~1.1s
Cold second GOOS lint (cache populated for first only) | ~3.9s
Warm second GOOS | ~1.1s
task lint warm (3 legs + gofumpt/fmt/vet/goreleaser) | ~4.8s
task precommit warm | ~5.6-6.1s

I'd suggest reverting the task precommit change, but not blocking. (Ask me in a few days... :D)

Also, makes sense to ignore ARCH, that's a bridge we can cross later.

AGENTS.md
```diff
@@ -77,17 +77,40 @@ func example() {
...
+through the internal `lint-goos` task, which loops over the `LINT_GOOS_TARGETS` variable
+in `Taskfile.yaml` (`linux darwin windows` — kept in sync with the `goos` lists in
+`goreleaser.yaml`). Output is labelled per leg, e.g. `🧹 Linting (GOOS=windows)…`.
+
+Practical notes:
+- Expect `task lint` to fail on a file your own OS never compiles. That is the point.
+  `task delint` can auto-fix those files from any host.
+- The overhead is small (~1s per extra GOOS on a warm cache); each GOOS gets its own
+  golangci-lint analysis cache.
+- **When adding a new release platform, add it to `LINT_GOOS_TARGETS`.**
```
ajalon1, line 113: I know this is for AGENTS.md, but still -- boo manual syncing.

ajalon1 (approved):
Meant to approve.

PR #732, opened 2026-08-04
feat(enclave): add enclave (outpost) management commands

artem-panchenko:
# RATIONALE

Covalent ships a public Enclaves API at `/api/v2/enclaves`. This adds first-class CLI support so operators can register and manage enclaves and control who may use them, instead of hand-rolling `curl` calls. It mirrors the existing `dr workload` command tree in layout, output handling, and telemetry wiring.

There is no public `activate` verb — an enclave activates itself when its installer reports in — so no `activate` command is provided (documented in `--help`). Manually deactivated enclaves can be returned to service with `reactivate`.

**Server floor.** The client calls only the canonical paths, which are new in CMPT-7347 ([covalent-cloud-server#681](https://github.com/datarobot/covalent-cloud-server/pull/681), merged 2026-09-04, so the first Covalent release after 11.12.0). Before that change the API sat behind a `/covalent` ingress prefix and its RBAC/lifecycle routes answered under `/outposts`; both old spellings survive server-side as hidden deprecated aliases, so nothing forces the CLI to keep using them — and referencing only the new paths means those aliases can be retired without touching the CLI again.

## CHANGES

New feature-gated `dr enclave` command tree (gated behind `DATAROBOT_CLI_FEATURE_ENCLAVE`, mirroring `dr workload`):

**Lifecycle**

- `dr enclave register --name <n> [--label k=v]` — surfaces the one-shot installation secrets (shown once, never persisted server-side; `--output-format json` to capture them)
- `dr enclave get <id>` / `dr enclave list`
- `dr enclave deactivate <id>` — takes an enclave out of service
- `dr enclave reactivate <id>` — returns a manually deactivated enclave to service (inverse of `deactivate`)
- `dr enclave delete <id> [--yes]` — confirmation prompt + friendly 404

**Per-enclave access** (`dr enclave access`)

- `grant <id> --role <owner|user|consumer> <recipient>` / `revoke <id> <recipient>` — the latter sends the `NO_ROLE` removal sentinel
- `list <id>` — who the enclave is shared with and the role each holds
- `show <id> [--user-id]` — the caller's effective permissions, and whether the answer came from a grant, the system-admin bypass, or RBAC being off

**Collection-level permission** (`dr enclave permission`)

- `grant` / `revoke --permission create` for a user / group / organization
- `list` — who has been granted the create permission
- `show [--user-id]` — whether the caller may create enclaves, and why

`access` commands wrap the `sharedRoles` PATCH endpoint, mapping `owner|user|consumer` → `OWNER|USER|CONSUMER`, with exactly-one recipient selection (`--user` / `--user-id` / `--group` / `--org`) enforced via `MarkFlagsMutuallyExclusive` plus a required-one check. `permission` recipients are id-based only (`--user-id` / `--group` / `--org`), because `createAccess` identifies subjects by id. `--help` calls out the server dependencies (`ENCLAVE_RBAC_ENABLED`, `CAN_SHARE`, `ENCLAVE_ADMIN_ONLY`).

Every command supports `--output-format json`.

Layout:

- `internal/enclave/` — API client (`enclave.go`), sharing client (`sharing.go`), permission client (`permissions.go`), output rendering (`output.go`)
- `cmd/enclave/**` — parent group + per-verb packages (`register`, `get`, `list`, `deactivate`, `reactivate`, `del`, `access/{grant,revoke,list,show}`, `permission/{grant,revoke,list,show}`)
- Registered in `cmd/root.go`; telemetry-wiring assertions added in `cmd/telemetry_wiring_test.go`
- `docs/commands/enclave.md` — command reference, including the command → endpoint table

One `basePath` const covers the whole surface, with the sub-resource paths composed off it, so there is a single place to change if the API moves again.

## TESTING

- `go build ./...` and `go vet ./...` clean; `golangci-lint run` reports **0 issues** on the pinned v2.11.4 (`golangci-lint fmt` produces no diff)
- Unit tests: httptest-backed client (register incl. 409, get + id-escaping, list, deactivate, reactivate, delete + 404, sharedRoles grant/revoke body + path, permission/createAccess), role/recipient parsing, flag/arg validation, telemetry wiring
- `go test -race ./internal/enclave/... ./cmd/enclave/...` green; full `go test -race ./...` green apart from `cmd/dotenv`'s pulumi-login tests, which fail the same way on `main` in this environment
- **Not** re-verified against a deployed environment: the canonical paths need a Covalent carrying covalent-cloud-server#681, which is merged to `main` but not in a release yet. Worth an end-to-end pass (register → share → `access show` / `access list` → `permission grant` → `permission list`) once it deploys, before this merges.

---

> [!NOTE]
> **Medium Risk**
> New commands can grant/revoke enclave access and delete enclaves, and registration output exposes one-shot credentials; impact is limited by feature gating and server-side RBAC, but operator mistakes could affect who can deploy or create outposts.
> 
> **Overview**
> Adds a feature-gated **`dr enclave`** command group (enable with `DATAROBOT_CLI_FEATURE_ENCLAVE`) for registering and operating remote outpost clusters via `/api/v2/enclaves`, wired into the root CLI with CODEOWNERS and command docs.
> 
> **Lifecycle:** `register` (one-shot install secrets + optional labels), `get`/`list`, `deactivate`/`reactivate`, and `delete` with interactive confirmation or `--yes` / `DATAROBOT_CLI_NON_INTERACTIVE`, including a friendly 404 on delete.
> 
> **Sharing & RBAC:** `dr enclave access` wraps `sharedRoles` (grant/revoke roles, list shares, show effective permissions with RBAC/sys-admin source hints). `dr enclave permission` manages collection-level **create** access via `createAccess` (id-only recipients). All support `--output-format json`.
> 
> New **`internal/enclave`** holds the HTTP client, sharing/permission helpers, and text/JSON rendering (including kubectl hints for registration secrets). Telemetry wiring tests cover a subset of leaf commands; unit tests cover clients and flag validation.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 6c90d82fee8356388a50c06081dd3dc3b396e636. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/cmd/enclave/access/revoke/cmd.go
+++ b/cmd/enclave/access/revoke/cmd.go
@@ -45,8 +45,7 @@ Choose exactly one recipient:
 
 Takes effect only with ENCLAVE_RBAC_ENABLED=true on the server; otherwise the
 call succeeds but changes nothing. The caller must hold CAN_SHARE on the
-enclave (owner or admin), and where ENCLAVE_ADMIN_ONLY=true, must also be a
-sysadmin or org admin.
+enclave; system administrators may manage any enclave.
 
 Example:
   dr enclave access revoke 3fa85f64-5717-4562-b3fc-2c963f66afa6 --user alice@corp.io
--- a/cmd/enclave/access/revoke/cmd.go
+++ b/cmd/enclave/access/revoke/cmd.go
@@ -1,0 +1,97 @@
+// Copyright 2026 DataRobot, Inc. and its affiliates.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package revoke
+
+import (
+	"github.com/datarobot/cli/internal/auth"
+	"github.com/datarobot/cli/internal/enclave"
+	"github.com/datarobot/cli/internal/outputformat"
+	"github.com/datarobot/cli/internal/telemetry"
+	"github.com/spf13/cobra"
+)
+
+func Cmd() *cobra.Command {
+	var outputFormat outputformat.OutputFormat
+
+	var (
+		user   string
+		userID string
+		group  string
+		org    string
+	)
+
+	cmd := &cobra.Command{
+		Use:   "revoke <enclave-id> <recipient>",
+		Short: "Revoke a recipient's access to an enclave.",
+		Long: `Revoke a recipient's access to an enclave, whatever role they currently hold.
+
+Choose exactly one recipient:
+  --user <username>   a user, by username
+  --user-id <id>      a user, by DataRobot user id
+  --group <id>        a group
+  --org <id>          an entire organization (all its users)
+
+Takes effect only with ENCLAVE_RBAC_ENABLED=true on the server; otherwise the
+call succeeds but changes nothing. The caller must hold CAN_SHARE on the
+enclave (owner or admin), and where ENCLAVE_ADMIN_ONLY=true, must also be a
+sysadmin or org admin.
+
+Example:
+  dr enclave access revoke 3fa85f64-5717-4562-b3fc-2c963f66afa6 --user alice@corp.io
+  dr enclave access revoke 3fa85f64-5717-4562-b3fc-2c963f66afa6 --org 656f0000000000000000abcd`,
+		Args:         cobra.ExactArgs(1),
+		PreRunE:      auth.EnsureAuthenticatedE,
+		SilenceUsage: true,
+		RunE: func(cmd *cobra.Command, args []string) error {
+			outputFormat = outputformat.GetFormat(cmd)
+
+			recipient, err := enclave.ResolveRecipient(user, userID, group, org)
+			if err != nil {
+				return err
+			}
+
+			if err := enclave.RevokeAccess(args[0], recipient); err != nil {
+				return err
+			}
+
+			return enclave.RenderAccessChange(outputFormat, enclave.AccessChange{
+				EnclaveID:     args[0],
+				Action:        "revoked",
+				RecipientType: recipient.Type,
+				Recipient:     recipient.Value(),
+			})
+		},
+	}
+
+	outputformat.AddFlag(cmd, &outputFormat)
+
+	cmd.Flags().StringVar(&user, "user", "", "Revoke a user by username")
+	cmd.Flags().StringVar(&userID, "user-id", "", "Revoke a user by DataRobot user id")
+	cmd.Flags().StringVar(&group, "group", "", "Revoke a group by id")
+	cmd.Flags().StringVar(&org, "org", "", "Revoke an entire organization by id")
+	cmd.MarkFlagsMutuallyExclusive("user", "user-id", "group", "org")
+
+	telemetry.TrackWith(cmd, func(_ *cobra.Command, args []string) map[string]any {
+		recipientType, _ := enclave.RecipientTypeOf(user, userID, group, org)
+
+		return map[string]any{
+			"enclave_id":     telemetry.FirstArg(args),
+			"recipient_type": recipientType,
+			"output_format":  string(outputFormat),
+		}
+	})
+
+	return cmd
+}
--- a/cmd/enclave/access/revoke/cmd.go
+++ b/cmd/enclave/access/revoke/cmd.go
@@ -33,7 +33,7 @@ func Cmd() *cobra.Command {
 	)
 
 	cmd := &cobra.Command{
-		Use:   "revoke <enclave-id> <recipient>",
+		Use:   "revoke <enclave-id>",
 		Short: "Revoke a recipient's access to an enclave.",
 		Long: `Revoke a recipient's access to an enclave, whatever role they currently hold.
 
--- a/cmd/enclave/del/cmd.go
+++ b/cmd/enclave/del/cmd.go
@@ -1,0 +1,128 @@
+// Copyright 2026 DataRobot, Inc. and its affiliates.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// Package del implements the `dr enclave delete` verb. The directory is named
+// `del` rather than `delete` because the latter shadows Go's built-in delete()
+// function in importing files.
+package del
+
+import (
+	"errors"
+	"fmt"
+	"net/http"
+
+	"github.com/datarobot/cli/cmd/helpers"
+	"github.com/datarobot/cli/internal/auth"
+	"github.com/datarobot/cli/internal/config/viperx"
+	"github.com/datarobot/cli/internal/drapi"
+	"github.com/datarobot/cli/internal/enclave"
+	"github.com/datarobot/cli/internal/misc/reader"
+	"github.com/datarobot/cli/internal/telemetry"
+	"github.com/datarobot/cli/tui"
+	"github.com/spf13/cobra"
+)
+
+func Cmd() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "delete <enclave-id>",
+		Short: "Delete an enclave.",
+		Long: `Delete an enclave (outpost) by id.
+
+This removes the enclave record from your tenant. To take an enclave out of
+service without removing it, use 'dr enclave deactivate' instead.
+
+Without --yes the command asks for confirmation.
+
+Example:
+  dr enclave delete 3fa85f64-5717-4562-b3fc-2c963f66afa6
+  dr enclave delete 3fa85f64-5717-4562-b3fc-2c963f66afa6 --yes`,
+		Args:         cobra.ExactArgs(1),
+		PreRunE:      auth.EnsureAuthenticatedE,
+		SilenceUsage: true,
+		RunE: func(cmd *cobra.Command, args []string) error {
+			confirmed, err := confirmDelete(cmd, args[0])
+			if err != nil || !confirmed {
+				return err
+			}
+
+			if err := enclave.DeleteEnclave(args[0]); err != nil {
+				return handleDeleteError(err, args[0])
+			}
+
+			fmt.Println(tui.BaseTextStyle.Render("Deleted enclave: " + args[0]))
+
+			return nil
+		},
+	}
+
+	cmd.Flags().BoolP("yes", "y", false, "Skip the confirmation prompt.")
+
+	// Bind only the env var (DATAROBOT_CLI_NON_INTERACTIVE) to viper. The --yes
+	// flag itself is read directly from cmd.Flags() so an explicit --yes does
+	// not leak into viper.AllSettings() and persist to drconfig.yaml.
+	_ = viperx.BindEnv("yes", "DATAROBOT_CLI_NON_INTERACTIVE")
+
+	telemetry.TrackWith(cmd, func(cmd *cobra.Command, args []string) map[string]any {
+		yesFlag, _ := cmd.Flags().GetBool("yes")
+
+		return map[string]any{
+			"enclave_id": telemetry.FirstArg(args),
+			"yes":        yesFlag || viperx.GetBool("yes"),
+		}
+	})
+
+	return cmd
+}
+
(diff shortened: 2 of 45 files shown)


ajalon1:
hey @artem-panchenko, thanks!. Have you already talked to @carsongee or @shreyaag-dr about this work? We'd like to see overall what your plans are for the CLI command hierarchy and API, and how they dovetail with the rest of the CLI.

ajalon1:
> Note: golangci-lint was not run in the authoring environment due to a Go toolchain version gap (repo targets 1.26.5).

Were you running into something like this?

> compile: version "go1.X.Y" does not match go tool version "go1.X.Z"

If so, https://github.com/datarobot-oss/cli/blob/main/docs/development/building.md#go-version-requirements-for-race-detection

cmd/telemetry_wiring_test.go
```diff
@@ -149,6 +150,40 @@ func TestTelemetryWiring_AllArtifactCommandsTracked(t *testing.T) {
...
+var expectedEnclaveTrackedCommands = []string{
+	"enclave register",
+	"enclave get",
+	"enclave list",
+	"enclave deactivate",
+	"enclave reactivate",
+	"enclave delete",
+	"enclave access grant",
+	"enclave access revoke",
+}
```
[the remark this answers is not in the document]
ajalon1: @artem-panchenko yup, this is important.

internal/enclave/output.go
```diff
@@ -0,0 +1,426 @@
...
+	sort.Strings(keys)
+
+	var out strings.Builder
+
+	for i, k := range keys {
+		if i > 0 {
+			out.WriteString(" ")
+		}
+
+		fmt.Fprintf(&out, "--from-literal=%s=%s", k, data[k])
```
ajalon1: 🤖 [P3] [security] Quote the API data interpolated into the generated kubectl command (non-blocking)

`kubectlLiterals` renders `--from-literal=%s=%s` with raw API-returned keys and values, and the secret `name` is interpolated unquoted into the `kubectl create secret generic` line a few lines up. The data crosses a trust boundary (API response → terminal → pasted shell command) without encoding: a value with whitespace or shell metacharacters (PEM certs, kubeconfig-style YAML) breaks the generated command, and a value containing `$(...)` would execute on paste. LOW severity since the values come from the authenticated server — fine to handle in a separate PR alongside the preview. Shell-quote the interpolations (single-quote with `\'` escaping).

cmd/enclave/del/cmd.go
```diff
@@ -0,0 +1,128 @@
...
+	return cmd
+}
+
+// confirmDelete returns (true, nil) when the deletion may proceed: either
+// --yes / DATAROBOT_CLI_NON_INTERACTIVE was given, or the user confirmed
+// interactively. A declined prompt is (false, nil) so the command exits 0
+// as a no-op.
+func confirmDelete(cmd *cobra.Command, enclaveID string) (bool, error) {
+	yesFlag, _ := cmd.Flags().GetBool("yes")
+	if yesFlag || viperx.GetBool("yes") {
```
ajalon1: 🤖 [P3] BLOCKING: use `cli.IsNonInteractive(cmd)` instead of inlining the `--yes` OR

`yesFlag || viperx.GetBool("yes")` bypasses the canonical consent check that AGENTS.md prescribes. `viperx.GetBool` is plain `viper.GetBool` (cast.ToBool), which rejects the truthy spellings `yes`/`y`/`t` that `reader.IsNonInteractive()` accepts — so `DATAROBOT_CLI_NON_INTERACTIVE=yes` in a CI run fails with `confirmation required` while every conforming command honors it. The sibling `cmd/artifact/del/cmd.go` is the pattern to copy; its telemetry closure also reads `cli.IsNonInteractive(cmd)` rather than duplicating the OR (line 81 here repeats the same broken OR and would misreport `yes`). Also prefer the `cli.YesFlagName` and `reader.NonInteractiveEnv` constants over the `"yes"`/`"DATAROBOT_CLI_NON_INTERACTIVE"` string literals.

docs/commands/enclave.md
```diff
@@ -0,0 +1,242 @@
...
+Show whether a subject may create enclaves, and where that comes from. Defaults to you; `--user-id` inspects somebody else (system administrators only).
+
+```bash
+dr enclave permission show
+dr enclave permission show --user-id <datarobot-user-id>
+```
+
+## Output formats
+
+Every subcommand accepts `--output-format` (`text` or `json`). `text` is the default and is meant for reading; `json` is the one to use in scripts, and the only practical way to capture `register`'s installation secrets.
```
ajalon1, line 251: 🤖 [P3] BLOCKING: `dr enclave delete` should honor `--output-format json`

`cmd/enclave/del/cmd.go` is the only leaf command without `outputformat.AddFlag` and never renders JSON. Because `--output-format` is a persistent root flag, `dr enclave delete <id> --output-format json` parses but prints plain text — a script following this docs claim (and the PR description's "Every command supports `--output-format json`") gets non-JSON stdout. The fix should be on the code side: add the flag and emit a JSON acknowledgement on success (and a JSON-shaped result for the friendly 404 no-op) rather than qualifying the docs.

cmd/enclave/access/revoke/cmd.go
```diff
@@ -0,0 +1,96 @@
...
+
+	var (
+		user   string
+		userID string
+		group  string
+		org    string
+	)
+
+	cmd := &cobra.Command{
+		Use:   "revoke <enclave-id> <recipient>",
```
ajalon1: 🤖 [P3] BLOCKING: usage string advertises a positional `<recipient>` the arg validation rejects

`dr enclave access revoke <id> <recipient>` fails with "accepts 1 arg(s), received 2" — recipients are flags-only here (`ExactArgs(1)` takes just the enclave id). Same problem in `cmd/enclave/access/grant/cmd.go` (`grant <enclave-id> --role <owner|user|consumer> <recipient>`) and in `cmd/enclave/permission/grant/cmd.go` / `cmd/enclave/permission/revoke/cmd.go`, which show `<recipient>` while `cobra.NoArgs` rejects any positional. Drop `<recipient>` from the `Use` strings (e.g. `revoke <enclave-id>` plus the flag help that's already there).

ajalon1 (commented):
Thanks @artem-panchenko! Pretty good overall. Security code review (STRIDE + OWASP) looked good overall; there's one small security finding that is non-blocking, see inline comments.

As for the actual code, three blocking issues, all small fixes. These are mostly just making sure your code follows established conventions in the codebase.

1. `dr enclave delete` accepts the persistent `--output-format` flag but ignores it — it should handle `json` properly, since the docs and the PR description promise every subcommand supports it.
2. `delete` inlines `yesFlag || viperx.GetBool("yes")` instead of calling `cli.IsNonInteractive(cmd)`, so `DATAROBOT_CLI_NON_INTERACTIVE=yes|y|t` doesn't count as consent (the sibling `dr artifact delete` has the right pattern).
3. Four usage strings advertise a positional `<recipient>` that `ExactArgs(1)`/`NoArgs` reject.
4. the existing bugbot note that `expectedEnclaveTrackedCommands` in `cmd/telemetry_wiring_test.go` covers only 8 of 14 leaf commands.

Your call on whether to do an end-to-end pass before or after merging.

ajalon1 (changes requested):
Whoops, wrong review outcome, requesting changes for the three blocking items from my prior review

cmd/telemetry_wiring_test.go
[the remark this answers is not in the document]
artem-panchenko: Fixed in e678689 — `access list`, `access show` and the four `permission` verbs are now in the list, so all 14 leaves are covered.

I kept it as a hand-maintained allowlist to match `expectedWorkloadTrackedCommands` / `expectedArtifactTrackedCommands`. That closes the gap Bugbot identified (a later removal of the wiring now fails), though it still won't catch a *new* leaf added without telemetry. A guard that walks the subtree and asserts the list is exhaustive would close that too — worth doing across all four subtrees at once if you want it.

cmd/enclave/del/cmd.go
artem-panchenko: Fixed in d749d81. `confirmDelete` and the telemetry closure both call `cli.IsNonInteractive(cmd)` now, and the string literals are replaced with `cli.YesFlagName` / `reader.NonInteractiveEnv`.

I reproduced it before fixing: built from 6c90d82, `DATAROBOT_CLI_NON_INTERACTIVE=yes dr enclave delete <id>` exits 1 with `confirmation required`; after the fix it deletes. `TestConfirmDelete_HonorsTruthyNonInteractiveSpellings` now covers all eleven spellings `reader.IsNonInteractive()` accepts, so the `cast.ToBool` gap can't come back.

docs/commands/enclave.md
artem-panchenko: Fixed on the code side as you suggested, in d749d81 — `outputformat.AddFlag` plus an `enclave.RenderDeletion` renderer.

Since 404-on-delete and a declined prompt are both exit-0 no-ops, a bare acknowledgement would have been ambiguous, so the result names the outcome:

```json
{ "enclaveId": "...", "deleted": true }
{ "enclaveId": "...", "deleted": false, "reason": "not found" }
{ "enclaveId": "...", "deleted": false, "reason": "aborted" }
```

One thing beyond what you asked: in JSON mode the confirmation prompt now writes to stderr, so stdout carries nothing but the document. Docs updated with the shape and that note. Verified against a stub server for all three outcomes, in both text and JSON, and via the persistent root flag (`dr --output-format json enclave delete`).

cmd/enclave/access/revoke/cmd.go
artem-panchenko: Fixed in 10153b6 — `<recipient>` dropped from all four. `access grant|revoke` now read `<verb> <enclave-id>`, and the two `permission` verbs take no positional at all. Confirmed against the built binary that each `Usage:` line now matches what the validator accepts; the recipient flags were already covered in `--help`.

internal/enclave/output.go
artem-panchenko: Done here rather than a separate PR (5faf1b7) — it's a small change, and it's a correctness bug as well as a security one: a PEM body or any value with whitespace breaks the generated command today.

Each `key=value` and the secret name are now wrapped as a single POSIX word with `'\''` escaping for embedded quotes, so `$(...)` is inert on paste and multi-line values survive. The `posixQuote` helper is duplicated from `cmd/auth/export` rather than shared, since `internal/` shouldn't import from `cmd/` — say the word if you'd rather it moved somewhere common.

artem-panchenko:
Thanks @ajalon1 — all four addressed, plus the P3 security note. Pushed as 5faf1b7.

1. **`delete` + `--output-format`** (d749d81) — added the flag and a `DeletionResult` renderer. It covers all three exit-0 paths so a script can distinguish them without parsing prose: `{"enclaveId": ..., "deleted": true}`, or `"deleted": false` with `"reason"` of `"not found"` / `"aborted"`. In JSON mode the confirmation prompt goes to stderr so stdout stays parseable. Docs updated with the shape.
2. **Consent check** (d749d81) — now `cli.IsNonInteractive(cmd)`, in both `confirmDelete` and the telemetry closure, using the `cli.YesFlagName` / `reader.NonInteractiveEnv` constants. Confirmed the bug first: a binary built from 6c90d82 fails `DATAROBOT_CLI_NON_INTERACTIVE=yes` with `confirmation required`; the fixed one deletes. Added a table test over every spelling `reader.IsNonInteractive()` accepts.
3. **Usage strings** (10153b6) — dropped `<recipient>` from all four.
4. **Telemetry list** (e678689) — all 14 leaves now listed. I kept the hand-maintained allowlist to match the workload/artifact tests; happy to add a tree-walking drift guard across all four subtrees in a follow-up if you'd prefer.
5. **kubectl quoting** (5faf1b7) — did it here rather than deferring, since an unquoted PEM value breaks the command today regardless of the security angle.

Verification: `golangci-lint run ./...` reports 0 issues on the pinned v2.11.4 and `golangci-lint fmt --diff` is empty; `go build ./...` and `go vet ./...` clean. Full `go test -count=1 ./...` green apart from `cmd/dotenv`'s pulumi-login tests, which fail identically with these changes stashed. The delete paths were also exercised end-to-end against a stub server (success, 404, declined prompt; text and JSON; local and inherited `--output-format`).

On the end-to-end pass: still gated on a Covalent release carrying covalent-cloud-server#681. I'd rather merge and run it when that deploys than hold the PR — but happy to wait if you'd prefer it verified first.

docs/commands/enclave.md
ajalon1: awesome, thank you. this will be very helpful for agents

cmd/enclave/del/cmd.go
ajalon1: thanks, I spent a lot of time recently cleaning up the yesFlag stuff.

ajalon1 (approved):
LGTM now, pending smoke tests

> On the end-to-end pass: still gated on a Covalent release carrying covalent-cloud-server#681. I'd rather merge and run it when that deploys than hold the PR — but happy to wait if you'd prefer it verified first.

ship it, we can iterate on this since you have a feature gate. if you want to test on your own before merging, you can actually tag a commit from this branch and have GH create a prerelease.

```
git tag v0.6.1-dev-enclave
git push upstream v0.6.1-dev-enclave
```

if you wnat to merge first, then tag from main. just make sure NOT to tag a commit twice, our CI doesn't like that :D

artem-panchenko:
@ajalon1 I don't have permissions to merge PRs in this repo, so I'd ask you to help me with that. Thanks!

The covalent change I mentioned above is delivered now (it's not gonna be enabled in MTS Prod yet, but is already a part of the upcoming onPrem release).

PR #733, opened 2026-08-04
feat(sync): generate uv.lock lazily during artifact code sync

adamalpi:
## What & why

Code-to-workload requires a `uv.lock` next to `pyproject.toml` — the workload-api image build runs `uv sync --frozen`. Users who don't commit a lock hit the failure late, at build trigger, with no hint about the fix ([RAPTOR-19186](https://datarobot.atlassian.net/browse/RAPTOR-19186)).

This PR makes `dr artifact code sync` generate the lock lazily on the developer's machine — where their uv configuration, private indexes, and credentials live:

- New `lockfile` engine phase (between `gather` and `manifests`): when `pyproject.toml` exists without `uv.lock`, run the user's local `uv lock` (2-min timeout). The generated file lands before the manifest walk, so it flows through the normal diff/upload pipeline — zero changes to the upload path.
- **Never blocks the sync**: `uv` missing or resolution failure logs a `WARN` with the exact fix (`uv lock`, then re-sync) and the sync proceeds; the build error tells the rest.
- **Never touches an existing `uv.lock`** (no clobbering user-managed locks).
- Warns when `.wapiignore` excludes `uv.lock` (the upload would be doomed to fail at build).
- On success logs `WARN Generated uv.lock from pyproject.toml — commit it to your repo`, consistent with the rest of the sync log stream.
- `LockfileRunner` is injected via the engine `Deps` seam for tests.

Note: generation also runs on `--dry-run`/`--diff` (so previews match a real sync); this writes `uv.lock` into the working tree — the file the user should commit anyway. Documented in `docs/commands/artifact.md`.

## How to test

- `task test` / `task lint` / `task dupcheck` — all green.
- Unit: `go test ./internal/workload/sync/ -run 'Lockfile|UvMissing'` — generated→uploaded, existing lock untouched, non-Python skip, uv-missing hint, failure hint, `.wapiignore` warning.
- Manual: in a project with only `pyproject.toml`, `dr artifact code init <id>` + `dr artifact code sync` → `WARN Generated uv.lock…`, file uploaded with the code; with `uv` off PATH → sync completes with the install hint.

## Breaking changes

None — purely additive; projects that already commit `uv.lock` see no behavior change.


adamalpi:
Recreating from a fork per CONTRIBUTING.md workflow.

PR #734, opened 2026-08-04
feat(sync): generate uv.lock lazily during artifact code sync

adamalpi:
## What & why

Code-to-workload requires a `uv.lock` next to `pyproject.toml` — the workload-api image build runs `uv sync --frozen`. Users who don't commit a lock hit the failure late, at build trigger, with no hint about the fix ([RAPTOR-19186](https://datarobot.atlassian.net/browse/RAPTOR-19186)).

This PR makes `dr artifact code sync` generate the lock lazily on the developer's machine — where their uv configuration, private indexes, and credentials live:

- New `lockfile` engine phase (between `gather` and `manifests`): when `pyproject.toml` exists without `uv.lock`, run the user's local `uv lock` (2-min timeout). The generated file lands before the manifest walk, so it flows through the normal diff/upload pipeline — zero changes to the upload path.
- **Never blocks the sync**: `uv` missing or resolution failure logs a `WARN` with the exact fix (`uv lock`, then re-sync) and the sync proceeds; the build error tells the rest.
- **Never touches an existing `uv.lock`** (no clobbering user-managed locks).
- Warns when `.wapiignore` excludes `uv.lock` (the upload would be doomed to fail at build).
- On success logs `WARN Generated uv.lock from pyproject.toml — commit it to your repo`, consistent with the rest of the sync log stream.
- `LockfileRunner` is injected via the engine `Deps` seam for tests.

Note: generation also runs on `--dry-run`/`--diff` (so previews match a real sync); this writes `uv.lock` into the working tree — the file the user should commit anyway. Documented in `docs/commands/artifact.md`.

## How to test

- `task test` / `task lint` / `task dupcheck` — all green.
- Unit: `go test ./internal/workload/sync/ -run 'Lockfile|UvMissing'` — generated→uploaded, existing lock untouched, non-Python skip, uv-missing hint, failure hint, `.wapiignore` warning.
- Manual: in a project with only `pyproject.toml`, `dr artifact code init <id>` + `dr artifact code sync` → `WARN Generated uv.lock…`, file uploaded with the code; with `uv` off PATH → sync completes with the install hint.

## Breaking changes

None — purely additive; projects that already commit `uv.lock` see no behavior change.

---

> [!NOTE]
> **Medium Risk**
> Runs a timed external `uv lock` subprocess and can write `uv.lock` during plan (including dry-run/diff), but failures are non-blocking and upload behavior is unchanged for projects that already commit a lock.
> 
> **Overview**
> Adds a **lockfile** sync phase (after gather, before manifest walk) so Python projects with `pyproject.toml` but no `uv.lock` can get a lock file before upload, matching workload image builds that expect `uv sync --frozen`.
> 
> When needed, sync runs the user’s local **`uv lock`** (injectable `LockfileRunner`, 2-minute timeout, user env/indexes). A new `uv.lock` is picked up by the normal diff/upload path; an existing lock is never overwritten. Missing `uv`, resolution failures, or `.wapiignore` excluding `uv.lock` only emit **`log.Warn`** hints—**sync still completes**. Docs describe auto-generation, dry-run/diff writing the file, and commit guidance; the sync command relies on engine warnings rather than extra CLI output.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit ec4333fbbaceec9eb702893ba17a6a246fce00b7. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


internal/workload/sync/phase_lockfile.go
```diff
@@ -0,0 +1,160 @@
...
+func runUvLock(dir string) error {
+	uvPath, err := exec.LookPath("uv")
+	if err != nil {
+		return errUvNotFound
+	}
+
+	ctx, cancel := context.WithTimeout(context.Background(), uvLockTimeout)
+	defer cancel()
+
+	cmd := exec.CommandContext(ctx, uvPath, "lock")
```
wojtekwdr, line 115: With Stdout/Stderr set to a bytes.Buffer, Run() waits for the pipe-copy goroutines, not just the process. If uv's child (a build backend, git) survives the timeout kill still holding the pipe, Run() blocks past the 2-minute deadline indefinitely, with the sync lock from preflight held the whole time. Setting cmd.WaitDelay bounds that; internal/plugin/exec.go does the same with 5 seconds.

internal/workload/sync/phase_lockfile.go
```diff
@@ -0,0 +1,160 @@
...
+			e.lockfileHint = fmt.Sprintf("Could not generate uv.lock (%v). "+
+				"The image build will fail until you add one — run `uv lock` and re-sync.", err)
+		}
+
+		log.Warn(e.lockfileHint)
+
+		return nil
+	}
+
+	e.lockfileGenerated = true
```
wojtekwdr, line 96: This trusts the runner without checking that uv.lock actually appeared in projectDir. In a uv workspace member, `uv lock` writes the lockfile at the workspace root, so this logs "Generated uv.lock, commit it" while the upload contains no lock, and every future sync reruns the whole generation. A re-stat after success would catch it.

internal/workload/sync/engine.go
```diff
@@ -208,6 +223,15 @@ func (e *Engine) Close() error {
 // from a previously crashed sync.
 func (e *Engine) StaleRollbackRestored() bool { return e.staleNote }
 
+// GeneratedLockfile reports whether the lockfile phase created uv.lock
+// from the project's pyproject.toml during Plan.
+func (e *Engine) GeneratedLockfile() bool { return e.lockfileGenerated }
```
wojtekwdr: GeneratedLockfile() and LockfileHint() have no caller outside the package tests: cmd.go deliberately relies on the engine's log.Warn, and engineRunner wasn't extended, yet the phase comment says the hint is recorded "for the cmd layer". Either wire them through runSync the way StaleRollbackRestored() is, or keep the state unexported until a consumer exists.

wojtekwdr (approved):
LGTM

ajalon1:
I think we should update CODEOWNERS to now tag CLI Maintainers anymore on this work :)

> cmd/artifact/code/codesync/cmd.go
> docs/commands/artifact.md

ajalon1 (approved):
LGTM

ajalon1 (approved):
Reapproving just to kick CI again

ajalon1 (approved):

ajalon1:
I've tmeporarily enabled merge restrictions bypass so that we can merge this in and get the "Build and Push PR Image" pipeline working properly. Sorry @adamalpi !

PR #739, opened 2026-08-05
[RAPTOR-18983][RAPTOR-18974] refactor(workload): move state under .datarobot and port up internals

wojtekwdr:
# RATIONALE

A project that uses the workload commands ends up with two DataRobot entries at its root: the `.datarobot.yaml` manifest and a separate `.wapi/` state directory. Two sibling dot-entries for one product reads like two tools, and it is the first thing a user sees when they open the project. All CLI-owned state moves under `.datarobot/`, which this CLI already uses for `.datarobot/cli/state.yaml`.

The same change also ports the `internal/workload` primitives that `dr workload up` will build on. No command surface changes there, so nothing user-visible ships from that half.

Closes RAPTOR-18983 and RAPTOR-18974.

### State directory

- `.wapi/` becomes `.datarobot/wapi/`, holding the same `config.json`, `manifest.json`, `history.log`, `.checkouts/` and self-ignoring `.gitignore`.
- `wapi.Dir()` is the single resolver. It falls back to a legacy root `.wapi/` when one is present, so a project linked by an older CLI stays readable. The fallback is a stat rather than cached state, so a migration that cannot complete degrades to reading the old location instead of failing the command.
- `wapi.EnsureMigrated()` relocates a legacy directory on the next state-touching command. It is silent and cannot fail; both outcomes are logged at debug level. `--dry-run` and `--diff` skip it, so a preview never moves anything on disk.
- `systemExcludes` gains `.datarobot/wapi`, not a bare `.datarobot`. Entries match as prefixes, so a bare one would also stop syncing `.datarobot/cli/state.yaml`. The legacy `.wapi` stays listed so an un-migrated project does not upload its own state.
- Hand-built paths in `sync/rollback.go` and `sync/synclock.go` now go through the package, so there is one definition of where state lives.
- Stale rollback recovery sweeps both locations, so backups from a sync that crashed before the directory moved are still applied.
- Help text and error strings across `cmd/artifact/code` and `cmd/artifact/build` follow. Read errors name the real file via `wapi.ConfigPath()` instead of a hard-coded path.
- `.wapiignore` stays at the project root: it is authored and committed by the user, the same convention as `.gitignore`.

### Workload internals

- `WaitForWorkload` with `IsTerminalWorkloadStatus` and `IsWorkloadErrorStatus`.
- New `internal/workload/execenv.go`: `ResolveExecutionEnvironment` maps a name or id to both execution-environment ids for generated-mode builds.
- Artifact spec primitives: `Container.Name`, `Container.EnvironmentVars`, the `EnvironmentVar` type including its credential-backed source shape, and the `PrimaryContainerName` and `PrimaryEnvironmentVars` helpers.
- `PatchPrimaryContainer` is deliberately not ported: artifacts are immutable now, every change mints a new version, and that removed every caller.

### Along the way

- `fsutil.DirExists` and `FileExists` handled only `ErrNotExist`, leaving `info` nil on any other stat error and dereferencing it. A project root containing a file named `.datarobot` produced a stack trace; both now return false for every stat error.
- The primary-container traversal existed in four copies in `artifact.go` and had drifted: `PrimaryContainerName` required a non-empty name, so an unnamed primary made it report a sidecar's name while the other three read the primary. All four now share one `primaryContainer()` helper.

## PR Automation

> [!IMPORTANT]
>

### DIFF
--- a/internal/workload/artifact.go
+++ b/internal/workload/artifact.go
@@ -64,6 +64,10 @@ type ContainerGroup struct {
 }
 
 type Container struct {
+	// Name is the container's identifier within its group. Workload runtime
+	// overrides (e.g. resourceAllocation) must reference the container by this
+	// name, so it is parsed for callers that build a runtime spec.
+	Name             string            `json:"name,omitempty"`
 	ImageBuildConfig *ImageBuildConfig `json:"imageBuildConfig,omitempty"`
 	// ImageURI is the resolved container image reference. Server-managed:
 	// before a successful build this is the placeholder from create; after
@@ -71,8 +75,34 @@ type Container struct {
 	ImageURI string `json:"imageUri,omitempty"`
 	// *bool so an absent value round-trips as nil and omitempty drops it
 	// on marshal, instead of re-asserting `false` back to the server.
-	Primary *bool `json:"primary,omitempty"`
+	Primary         *bool            `json:"primary,omitempty"`
+	EnvironmentVars []EnvironmentVar `json:"environmentVars,omitempty"`
 }
+
+// EnvironmentVar is one entry in a container's environmentVars array. A plain
+// var carries Value directly; a credential-backed var has Source ==
+// "dr-credential" and carries DRCredentialID/Key instead, so the secret never
+// appears in the spec. Source is omitempty because the server defaults plain
+// vars to "string".
+//
+// Key selects which field of the credential to use, not to be confused with
+// Name (the env var's own name): one stored credential can bundle several
+// fields, as an S3 credential bundles awsAccessKeyId and awsSecretAccessKey.
+type EnvironmentVar struct {
+	Source         string `json:"source,omitempty"`
+	Name           string `json:"name"`
+	Value          string `json:"value,omitempty"`
+	DRCredentialID string `json:"drCredentialId,omitempty"`
+	Key            string `json:"key,omitempty"`
+}
+
+// EnvironmentVarSourceDRCredential marks a variable the platform resolves from
+// the credential store at container start.
+const EnvironmentVarSourceDRCredential = "dr-credential"
+
+// defaultPrimaryContainerName is the name the CLI writes into the specs it
+// generates, and the fallback for an unnamed primary container.
+const defaultPrimaryContainerName = "primary"
 
 type ImageBuildConfig struct {
 	CodeRef    *CodeRef    `json:"codeRef,omitempty"`
@@ -129,31 +159,38 @@ func (a *Artifact) IsLocked() bool {
 	return strings.EqualFold(a.Status, ArtifactStatusLocked)
 }
 
-// ExtractCodeRef mirrors the write-side selection in setPrimaryCodeRefInRawArtifact:
-// once a primary container is found it commits, returning nil if the primary
-// has no codeRef rather than falling through to a sidecar (which would surface
-// stale catalog info in display). Falls back to containerGroups[0].containers[0]
-// when no container is flagged primary.
-func ExtractCodeRef(artifact Artifact) *DatarobotCodeRef {
-	for _, group := range artifact.Spec.ContainerGroups {
-		for _, container := range group.Containers {
-			if container.Primary == nil || !*container.Primary {
-				continue
+// primaryContainer is the single definition of which container the CLI reads
+// per-container fields from: the one flagged "primary": true, falling back to
+// containerGroups[0].containers[0], nil when the artifact has neither. Every
+// exported Primary* helper goes through it so they cannot disagree. Mirrors
+// setPrimaryCodeRefInRawArtifact: once a primary is found it commits rather
+// than falling through to a sidecar.
+func primaryContainer(artifact Artifact) *Container {
+	for i, group := range artifact.Spec.ContainerGroups {
+		for j, container := range group.Containers {
+			if container.Primary != nil && *container.Primary {
+				return &artifact.Spec.ContainerGroups[i].Containers[j]
 			}
-
-			return codeRefFromContainer(container)
 		}
 	}
 
-	if len(artifact.Spec.ContainerGroups) == 0 {
+	if len(artifact.Spec.ContainerGroups) == 0 || len(artifact.Spec.ContainerGroups[0].Containers) == 0 {
 		return nil
 	}
 
-	if len(artifact.Spec.ContainerGroups[0].Containers) == 0 {
+	return &artifact.Spec.ContainerGroups[0].Containers[0]
+}
+
+// ExtractCodeRef returns the primary container's codeRef, or nil when the
+// primary has no codeRef (rather than falling through to a sidecar, which
+// would surface stale catalog info in display).
+func ExtractCodeRef(artifact Artifact) *DatarobotCodeRef {
+	container := primaryContainer(artifact)
+	if container == nil {
 		return nil
 	}
 
-	return codeRefFromContainer(artifact.Spec.ContainerGroups[0].Containers[0])
+	return codeRefFromContainer(*container)
 }
 
 func codeRefFromContainer(container Container) *DatarobotCodeRef {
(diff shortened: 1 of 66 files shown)


c-h-russell-walker (commented):
FYI we already use `.datarobot` in Codespaces so you may want to ensure this works well with that.

Please see:
https://github.com/datarobot/notebooks/tree/main#usage-of-datarobot-in-codespaces

Image

internal/workload/wapi/paths.go
```diff
@@ -22,7 +22,21 @@ import (
...
 // Exported names used by external callers (tests, sync commands).
 const (
-	DirName          = ".wapi"
+	// RootDirName is the single directory a project gives DataRobot. The CLI
+	// already keeps its tool state under it, so sync state joins it rather
+	// than claiming a second dot-entry at the project root.
+	RootDirName = ".datarobot"
+
+	// StateDirName is the sync state directory, nested inside RootDirName.
+	StateDirName = "wapi"
```
adamalpi: **[Medium] Naming: use `.datarobot/workload/` instead of `.datarobot/wapi/`**

"wapi" is internal naming (Workload API) that means nothing to users, and we should be moving away from it in user-facing surfaces. The directory name leaks into help text, error messages, and docs throughout this PR.

Since this release is the one migration window — the directory is being renamed anyway and the `LegacyDirName` handling already exists — this is the cheapest moment to pick the right name. Renaming after release means building and shipping a second migration path.

internal/workload/wapi/migrate.go
```diff
@@ -0,0 +1,61 @@
...
+		return
+	}
+
+	if err := os.Rename(legacy, current); err != nil {
+		log.Debug("state dir migration: rename failed, using legacy location", "from", legacy, "to", current, "err", err)
+
+		return
+	}
+
+	log.Debug("state dir migration: moved", "from", legacy, "to", current)
```
adamalpi, line 67: **[Medium] `.wapi/` deprecation should be visible to the user**

Accepting the legacy `.wapi/` and renaming it on the next execution is already implemented here (`Dir()` fallback + `EnsureMigrated`), but every outcome is `log.Debug` only, so the migration is invisible unless `--debug` is on.

When a legacy `.wapi/` is detected, print a one-line notice to stderr — e.g. "`.wapi/` is deprecated; state moved to `.datarobot/…`" — so users understand why the directory vanished, and the un-migratable cases (read-only FS, both-locations-present) get a hint instead of silence.

internal/workload/execenv.go
```diff
@@ -0,0 +1,83 @@
...
+
+	for pageURL != "" {
+		var list executionEnvironmentList
+
+		if err := drapi.GetJSON(pageURL, "execution environments", &list); err != nil {
+			return "", "", err
+		}
+
+		for _, ee := range list.Data {
+			if ee.ID != nameOrID && ee.Name != nameOrID {
```
adamalpi: **[Medium] Duplicate EE names resolve silently to the first match**

Execution environment names are user-supplied and not unique, so matching by name with first-match-wins across pages means a duplicate name silently resolves to whichever EE the server lists first — a generated-mode build would target the wrong base environment with no signal.

Consider collecting all matches and erroring on ambiguity (>1 name hit). Cheap to fix now, before `dr workload up` wires this to a command.

internal/workload/wapi/migrate.go
```diff
@@ -0,0 +1,61 @@
...
+)
+
+// EnsureMigrated moves a project's state out of the legacy root .wapi/ and
+// into .datarobot/wapi/. Call it before any command that touches state.
+//
+// It is silent and cannot fail. The state directory is machine-managed and
+// undocumented, so relocating it is housekeeping the user cannot act on. A
+// move that cannot be completed leaves the legacy directory alone and Dir
+// keeps resolving there, so the command carries on either way.
+func EnsureMigrated(projectDir string) {
```
adamalpi: **[Low] Migration runs without lock coordination**

`EnsureMigrated` renames the state dir without checking the sync lock. Same-version concurrency is fine (the flock'd `sync.lock` inode travels with the rename and `Dir()` re-resolves per call), but a concurrent **old-CLI** sync mid-run has `.wapi/...` hard-coded: its writes start failing mid-sync and its own `Restore()` fails because the rollback tree moved. The `StaleRollbackDirs` sweep recovers the files on the next command, so nothing is lost — but the old process dies with a confusing error.

Cheap hardening: skip migration when `sync.lock` in the legacy dir is currently held.

internal/workload/artifact.go
```diff
@@ -64,16 +64,46 @@ type ContainerGroup struct {
...
+// appears in the spec. Source is omitempty because the server defaults plain
+// vars to "string".
+//
+// Key selects which field of the credential to use, not to be confused with
+// Name (the env var's own name): one stored credential can bundle several
+// fields, as an S3 credential bundles awsAccessKeyId and awsSecretAccessKey.
+type EnvironmentVar struct {
+	Source         string `json:"source,omitempty"`
+	Name           string `json:"name"`
+	Value          string `json:"value,omitempty"`
```
adamalpi, line 94: **[Low] `EnvironmentVar.Value` won't round-trip an explicit empty string**

`Value` is `omitempty`, so `{"name":"X","value":""}` drops its `value` key on marshal. Read-only today, but this type exists precisely so `dr workload up` can write specs; if the server distinguishes "value absent" from "value empty", this will bite then.

adamalpi (commented):
Solid PR overall — migration edge cases are well covered, the lock-inode analysis holds under rename, path normalization makes the new two-segment exclude safe on Windows, and all touched packages pass `go test -race` at the PR head. A few findings below; the two Medium naming/UX ones are worth deciding before merge since this release is the one cheap migration window.

c-h-russell-walker:
> FYI we already use `.datarobot` in Codespaces so you may want to ensure this works well with that.
> 
> Please see: https://github.com/datarobot/notebooks/tree/main#usage-of-datarobot-in-codespaces

Remembered I wanted to over-communicate that besides checking the functionality in Codespaces that it's also important to note:

> ... hidden directory in Codespaces ...

So if you put things in `.datarobot` it won't be see in the Codespaces (left sidepanel) filesystems UI (or the API response that powers it) - it'll show in the terminal of course but not in API/UI - it's excluded in code:

> We purposefully don't return this directory in the API results that the UI uses and therefore this directory does not show in the filesystem left sidepanel. This directory name is a constant DR_NOTEBOOKS_FILESYSTEM_HIDDEN_DIRECTORY seen in [/common/collections/filesystems.py](https://github.com/datarobot/notebooks/blob/main/common/collections/filesystems.py)

https://github.com/datarobot/notebooks/blob/main/common/collections/filesystems.py#L91

ajalon1 (commented):
🤖 Two minor security observations from a review pass (low severity / defense-in-depth): (1) case-sensitive system-exclude matching in the ignore matcher, and (2) an unbounded pagination loop in ResolveExecutionEnvironment. Repros and suggested fixes are inline.

internal/workload/ignore/matcher.go
```diff
@@ -27,7 +27,12 @@ import (
 const wapiignoreFile = ".wapiignore"
 
 // systemExcludes are always-ignored paths, not overridable by .wapiignore.
-var systemExcludes = []string{".wapi", ".git", ".gitignore"}
+//
+// The state directory is listed at its full path, not as a bare ".datarobot":
+// entries match as prefixes, so a bare one would also exclude the CLI's own
+// tool state under .datarobot/cli/, which syncs today. The legacy ".wapi"
+// stays listed so an un-migrated project does not upload its state.
+var systemExcludes = []string{".datarobot/wapi", ".wapi", ".git", ".gitignore"}
```
ajalon1: 🤖 Minor: the system-exclude match is case-sensitive — repro + suggested fix.

`matchesSystemExclude` compares `relPath` against these lowercase constants with exact/prefix equality, and `NormalizePath` only ToSlashes — no case-folding. On macOS APFS and Windows NTFS (both case-insensitive but case-preserving, and two of our lint targets), a repo that pre-creates the state dir as e.g. `.Datarobot/Wapi/` evades the exclude and its `config.json`/`manifest.json` get synced to the remote — silently breaking the "always-ignored, not overridable by `.wapiignore`" guarantee the comment above promises.

Repro (drop-in copy of the matcher logic):

```go
var systemExcludes = []string{".datarobot/wapi", ".wapi", ".git", ".gitignore"}

func matchesSystemExclude(relPath string) bool {
	for _, name := range systemExcludes {
		if relPath == name || strings.HasPrefix(relPath, name+"/") {
			return true
		}
	}
	return false
}

// .datarobot/wapi/config.json   excluded=true
// .Datarobot/Wapi/config.json   excluded=false   <-- leaks
// .DATAROBOT/WAPI/config.json   excluded=false   <-- leaks
```

Impact is low — `config.json` holds IDs (no credentials) and `manifest.json` mirrors the file inventory already being synced — so this is defense-in-depth. Suggested fix: fold case before matching:

```go
func matchesSystemExclude(relPath string) bool {
	rel := strings.ToLower(relPath)
	for _, name := range systemExcludes {
		if rel == name || strings.HasPrefix(rel, name+"/") {
			return true
		}
	}
	return false
}
```

internal/workload/execenv.go
```diff
@@ -0,0 +1,83 @@
...
+// name and returns its id and latest successful version id. `dr workload up`
+// uses it to fill the executionEnvironmentId/versionId of a generated-Dockerfile
+// artifact so the user names an EE without pasting ids.
+func ResolveExecutionEnvironment(nameOrID string) (id, versionID string, err error) {
+	pageURL, err := config.GetEndpointURL("/api/v2/executionEnvironments/?limit=100")
+	if err != nil {
+		return "", "", err
+	}
+
+	for pageURL != "" {
```
ajalon1: 🤖 Heads-up: this pagination loop has no cap — repro + suggested fix.

`ResolveExecutionEnvironment` loops `for pageURL != ""` following server `next` cursors with no max-page or item bound. The siblings (`ListArtifacts`, `ListWorkloads`) both stop once they've collected `limit` items; this one doesn't. If the API ever returns an endless same-host `next` without the target environment in `data`, the CLI just keeps fetching and hangs `dr workload up`.

Repro (mirrors the loop against a server that never stops paging):

```go
host := "app.datarobot.com"
pageURL := "https://" + host + "/api/v2/executionEnvironments/?limit=100"
for pageURL != "" {
	u, _ := url.Parse(pageURL)
	if u.Host != host { // AssertNextOnSameHost — always passes here
		break
	}
	// server keeps paging forever; target env never appears in data
	pageURL = "https://" + host + "/api/v2/executionEnvironments/?limit=100&offset=..."
}
// no natural termination
```

Good news: `AssertNextOnSameHost` is applied to every fetch, so this can't leak the token to an attacker host or do SSRF — it's purely an availability thing, and it needs a compromised/MITM'd API to trigger. Suggested fix: mirror the bounded sibling pattern:

```go
const maxPages = 1000
for pages := 0; pageURL != "" && pages < maxPages; pages++ {
	// ...existing fetch + match logic...
}
if pageURL != "" {
	return "", "", fmt.Errorf("execution environment %q not found; too many pages", nameOrID)
}
```

internal/workload/wapi/paths.go
wojtekwdr: fixed

internal/workload/wapi/migrate.go
wojtekwdr: fixed

internal/workload/execenv.go
wojtekwdr: fixed

internal/workload/wapi/migrate.go
wojtekwdr: fixed

internal/workload/artifact.go
wojtekwdr: fixed

internal/workload/ignore/matcher.go
wojtekwdr: fixed

internal/workload/execenv.go
wojtekwdr: fixed

adamalpi (approved):

PR #746, opened 2026-08-11
[CFX-7264] Add a --source filter to dr llm-gateway list and fetch both sources in parallel

chasdr:
## Summary

`dr llm-gateway list` fetched the LLM Gateway catalog and DataRobot deployments one after the other, so it waited on the sum of both. They now run in parallel. New `--source gateway|deployed` skips the other request entirely. Default `all` keeps today's behavior.

Against a server with a fixed 2s delay per route: 6.34s before, 4.21s after. One round-trip removed.

## Notes for review

The cached API token was resolved lazily and written without a lock, so two concurrent fetches raced on it. Every access now takes a mutex. `getToken` holds it across the resolve so concurrent callers share one verification round-trip.

A single source that cannot be reached is an error. Only `all` degrades to whichever source is still up, since with one source requested there is no remainder to show.

## Output

Default, both sources:

```
$ dr llm-gateway list
Available LLMs
──────────────
╭───────────────────────────────┬────────────────────────────┬──────────┬──────────────┬─────────────────────┬─────────╮
│ ID                            │ NAME                       │ SOURCE   │ PROVIDER     │ MODEL               │ CONTEXT │
├───────────────────────────────┼────────────────────────────┼──────────┼──────────────┼─────────────────────┼─────────┤
│   azure-openai-gpt-5-5        │ Azure OpenAI GPT-5.5       │ gateway  │ Azure OpenAI │ azure/gpt-5-5-2026- │ 1050000 │
│                               │                            │          │              │ 04-23               │         │
│   anthropic-claude-sonnet-4-6 │ Anthropic Claude Sonnet    │ gateway  │ Anthropic    │ bedrock/anthropic.c │ 200000  │
│                               │ 4.6                        │          │              │ laude-sonnet-4-6    │         │
│   6512a0b3f4d5e6a7b8c9d0e1    │ Support RAG Assistant      │ deployed │ -            │ -                   │ -       │
│   6512a0b3f4d5e6a7b8c9d0e2    │ Claims Summarizer          │ deployed │ -            │ -                   │ -       │
╰───────────────────────────────┴────────────────────────────┴──────────┴──────────────┴─────────────────────┴─────────╯
```

One source, no request to the other:

```
$ dr llm-gateway list --source deployed
Available LLMs
──────────────
╭────────────────────────────┬───────────────────────┬──────────┬──────────┬───────┬─────────╮
│ ID                         │ NAME                  │ SOURCE   │ PROVIDER │ MODEL │ CONTEXT │
├────────────────────────────┼───────────────────────┼──────────┼──────────┼───────┼─────────┤
│   6512a0b3f4d5e6a7b8c9d0e1 │ Support RAG Assistant │ deployed │ -        │ -     │ -       │
│   6512a0b3f4d5e6a7b8c9d0e2 │ Claims Summarizer     │ deployed │ -        │ -     │ -       │
╰────────────────────────────┴───────────────────────┴──────────┴──────────┴───────┴─────────╯
```

JSON shape is unchanged. `source` carries the values `--source` takes:

```
$ dr llm-gateway list --source gateway --output-format json
{
  "llms": [
    {
      "id": "azure-openai-gpt-5-5",
      "name": "Azure OpenAI GPT-5.5",
      "source": "gateway",
      "provider": "Azure OpenAI",
      "model": "azure/gpt-5-5-2026-04-23",
      "description": "Frontier model for complex professional work.",
      "context_size": 1050000,
      "deployment_id": "",
      "selected": false
    },
```

```
$ dr llm-gateway list --source bogus
Error: invalid argument "bogus" for "--source" flag: invalid source "bogus": use all, gateway, or deployed
```

## Technical Changes

- `internal/drapi/llms.go`: both fetches in goroutines. Ordering, per-source warnings, and the both-failed error unchanged.
- `internal/drapi/client.go`, `internal/drapi/get.go`: mutex over every access to the cached token.
- `cmd/llm-gateway/list/source.go`: new. `Source` flag type and `fetchLLMs`, plus a `Count` fix.
- `cmd/llm-gateway/list/cmd.go`: flag, shell completion, telemetry field.
- `docs/commands/`: the flag, and the best-effort claim scoped to where it holds.
- 10 tests. The load-bearing ones pin the route overlap, the skipped request, and the token accessors under `-race`.

`select` picks up the parallel fetch. It does not get `--source`.

### DIFF
--- a/cmd/llm-gateway/list/source.go
+++ b/cmd/llm-gateway/list/source.go
@@ -21,9 +21,9 @@ import (
 	"github.com/spf13/pflag"
 )
 
-// Source selects which LLM sources the list command queries. The gateway and
-// deployed values are the same strings the SOURCE column and the JSON "source"
-// field carry, so a filter can be copied straight out of a listing.
+// Source selects which LLM sources the list command queries. Its gateway and
+// deployed values are the strings the SOURCE column and the JSON "source" field
+// already carry, so a filter can be copied straight out of a listing.
 type Source string
 
 const (
@@ -57,13 +57,12 @@ func (s *Source) Type() string {
 	return "source"
 }
 
-// fetchLLMs queries only the sources the filter needs, skipping the request the
-// other source would have made.
+// fetchLLMs skips the request the unselected source would have made.
 //
 // A single-source request returns its error rather than degrading to an empty
-// list: the union path can fall back on the other source, but here there is no
-// remainder to show, and reporting "no models" for an unreachable catalog reads
-// as an empty instance.
+// list. The union path can fall back on the other source; here there is no
+// remainder to show, and "no models" for an unreachable catalog reads as an
+// empty instance.
 func fetchLLMs(source Source) (*drapi.LLMList, error) {
 	if source == SourceGateway {
 		gateway, err := drapi.GetLLMs()
--- a/cmd/llm-gateway/list/source.go
+++ b/cmd/llm-gateway/list/source.go
@@ -66,7 +66,17 @@ func (s *Source) Type() string {
 // as an empty instance.
 func fetchLLMs(source Source) (*drapi.LLMList, error) {
 	if source == SourceGateway {
-		return drapi.GetLLMs()
+		gateway, err := drapi.GetLLMs()
+		if err != nil {
+			return nil, err
+		}
+
+		// GetLLMs leaves Count and TotalCount as the last decoded page reported
+		// them, which is not the filtered row count. Restate them over the rows
+		// actually returned so all three branches agree on what the fields mean.
+		gateway.Count, gateway.TotalCount = len(gateway.LLMs), len(gateway.LLMs)
+
+		return gateway, nil
 	}
 
 	if source == SourceDeployed {
--- a/cmd/llm-gateway/list/source.go
+++ b/cmd/llm-gateway/list/source.go
@@ -1,0 +1,83 @@
+// Copyright 2026 DataRobot, Inc. and its affiliates.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package list
+
+import (
+	"fmt"
+
+	"github.com/datarobot/cli/internal/drapi"
+	"github.com/spf13/pflag"
+)
+
+// Source selects which LLM sources the list command queries. The gateway and
+// deployed values are the same strings the SOURCE column and the JSON "source"
+// field carry, so a filter can be copied straight out of a listing.
+type Source string
+
+const (
+	SourceAll      Source = "all"
+	SourceGateway  Source = drapi.LLMKindGateway
+	SourceDeployed Source = drapi.LLMKindDeployed
+)
+
+var _ pflag.Value = (*Source)(nil)
+
+func (s *Source) String() string {
+	if s == nil {
+		return ""
+	}
+
+	return string(*s)
+}
+
+func (s *Source) Set(v string) error {
+	switch Source(v) {
+	case SourceAll, SourceGateway, SourceDeployed:
+		*s = Source(v)
+
+		return nil
+	}
+
+	return fmt.Errorf("invalid source %q: use %s, %s, or %s", v, SourceAll, SourceGateway, SourceDeployed)
+}
+
+func (s *Source) Type() string {
+	return "source"
+}
+
+// fetchLLMs queries only the sources the filter needs, skipping the request the
+// other source would have made.
+//
+// A single-source request returns its error rather than degrading to an empty
+// list: the union path can fall back on the other source, but here there is no
+// remainder to show, and reporting "no models" for an unreachable catalog reads
+// as an empty instance.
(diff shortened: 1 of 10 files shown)


c-h-russell-walker (approved):
This is great - looks good to me.

internal/drapi/llms.go
```diff
@@ -219,9 +220,40 @@ func GetDeployedLLMs() ([]LLM, error) {
...
+		wg       sync.WaitGroup
+	)
+
+	wg.Add(2)
+
+	go func() {
+		defer wg.Done()
+
+		gateway, gwErr = GetLLMs()
+	}()
```
ajalon1, line 246: My 🤖 doesn't want to block your 🤖 here, but maybe a panic recovery would be nice in this func.

https://go.dev/blog/defer-panic-and-recover

New to me, learning something every day

internal/drapi/client.go
```diff
@@ -23,17 +23,29 @@ package drapi
...
 // DefaultClientTimeout is the read/write timeout used by NewHTTPClient when
 // callers don't specify their own.
 const DefaultClientTimeout = 30 * time.Second
 
+// tokenMu guards every access to the `token` global (declared in get.go),
+// including GetToken and SetToken. Callers that fetch from more than one
+// endpoint concurrently (GetLLMsAndDeployed) would otherwise race on it.
+// getToken holds the lock across resolveToken as well, so the first concurrent
+// caller's verification round-trip is not duplicated by the rest.
+var tokenMu sync.Mutex
```
ajalon1: I like this a lot. I wonder if this is something we should enforce being when invoking internal/drapi.

Also curious why sync.Once wasn't sufficient.

internal/drapi/client_test.go
```diff
@@ -120,3 +121,34 @@ func TestAuthorizeRequest_DoesNotConsumeBody(t *testing.T) {
 	require.NoError(t, err)
 	assert.Equal(t, "payload", string(body))
 }
+
+// TestTokenAccessorsAreGuarded drives SetToken and GetToken against getToken
+// concurrently. All three touch the same package global, and getToken is
+// reached from two goroutines whenever GetLLMsAndDeployed runs, so a re-auth
+// path calling SetToken mid-fetch must not race the readers. Only fails under
+// -race, which is how the suite runs.
+func TestTokenAccessorsAreGuarded(t *testing.T) {
```
ajalon1, line 130: nice

internal/drapi/llms_test.go
```diff
@@ -239,3 +241,84 @@ func TestGetLLMsAndDeployed_BothFail(t *testing.T) {
...
 }
+
+// TestGetLLMsAndDeployed_FetchesConcurrently holds each route until the other
+// has been entered. A serial fetch can never satisfy both, so this is what
+// stops the two sources from silently reverting to sum-of-both latency; the
+// union and soft-degrade tests pass either way.
+//
+// The barrier keys on the route, not on a request count: if either fixture ever
+// grows a Next page, two pages from one source would satisfy a counter and the
+// test would go green against a serial implementation.
```
ajalon1: I don't quite understand this part of the test.

cmd/llm-gateway/list/source.go
```diff
@@ -0,0 +1,93 @@
...
+// Source selects which LLM sources the list command queries. The gateway and
+// deployed values are the same strings the SOURCE column and the JSON "source"
+// field carry, so a filter can be copied straight out of a listing.
+type Source string
+
+const (
+	SourceAll      Source = "all"
+	SourceGateway  Source = drapi.LLMKindGateway
+	SourceDeployed Source = drapi.LLMKindDeployed
+)
```
ajalon1, line 33: nice

I was going to ask why not using ` = iota`, but then that adds a ton of additional boilerplate for converting from ints to strings, when you really just want strings

ajalon1 (commented):
Really happy you decided to tackle the parallelization. :) Should be useful going forward.

Filed https://datarobot.atlassian.net/browse/CFX-7558 for something related to GetLLMs() surfaced while reviewing this. Gave it to @c-h-russell-walker since he thinks this one was good.

Good tests as always.

No collisions for the `--source` flag that I could find, so I think we're good on that front.

internal/drapi/client.go
ajalon1: I am reminded of [singleflight](https://pkg.go.dev/golang.org/x/sync/singleflight) which is a library @roma-glushko, may he rest in nvidia peace, suggested to me a while back. I might add that as a tech debt item. ETA: CFX-7559

ajalon1 (approved):

internal/drapi/client.go
chasdr: Once burns on the first attempt, not the first success. A failed resolve spends it and evrything after gets `("", nil)` ->  no token, no error = bad :)

chasdr: this singleflight thing looks 👍

internal/drapi/llms_test.go
chasdr: I can improve the inline comments here sorry ..  Each handler blocks until the other route arrives. A serial fetch can never satisfy that: #2 can't start until #1 finishes, and #1 is waiting on #2. It hangs, hits the 3s timeout, fails.

every other test here passes whether the fetch is serial or parallel. They check the merged list, the warnings, and the error when both sources are down. None of them can tell you the two calls actually overlapped. So if someone later collapses the goroutines back into two plain calls, the suite stays green and the latency quietly comes back. This one goes red. I checked by reverting the goroutines and running it.

internal/drapi/client.go
roma-glushko: Perfect fit for singleflight! 
It's always useful for gating authentication whether it's a microservice or a standalone CLI like this one.

internal/drapi/llms_test.go
ajalon1: gottt it

PR #747, opened 2026-08-11
[RAPTOR-18975] feat(workload): manifest render, atomic write and the line-numbered validator

wojtekwdr:
# RATIONALE

RAPTOR-18973 landed the read half of `internal/workload/manifest` (parse, locate, compile), and its `doc.go` promised the write half as a stacked follow-up. This is that follow-up.

Nothing can produce a `.datarobot.yaml` until the package can render one and say whether it is valid, so this blocks the `dr workload config` wizard stacked behind it. Splitting it out keeps the wizard PR reviewable and lets the file format be reviewed on its own terms, by people who care about the schema rather than about Bubble Tea.

The validator reports line numbers because the file is hand-edited after generation. "memory must use 1000-based units" is a support ticket; "line 34: memory must use 1000-based units" is a fix.

## CHANGES

**`internal/fsutil` (moved from `internal/workload/wapi/atomic.go`)**

`AtomicWriteFile` was already generic and is now needed by `manifest`, which must not import `wapi`. Call sites in `wapi` updated; no behaviour change for them beyond the one below.

> [!IMPORTANT]
> **Behaviour change worth a look.** `AtomicWriteFile` used to `chmod` to 0644 unconditionally, so a developer running `umask 077` got a world-readable `config.json` and `history.log` from a CLI that had no business widening them. It now honours umask on a new file and preserves the existing mode on a replace. Verified at umask 022, 077 and 002.

**`manifest/write.go`**

- `Draft`, the narrow struct for the subset a generator settles. Deliberately not a spec struct: reads stay node-based so unknown blocks survive a round trip, which is the package's stated reason for having no typed structs.
- `Render(Draft)` produces the commented YAML, with `importance` carrying its allowed values inline.
- `Write` reserves the path with `O_EXCL` before writing, so two concurrent runs cannot both believe they created the file, then writes atomically.
- `WriteWorkloadID` is a single-key edit that leaves comments, unknown keys and key order intact. `up` calls it after the first create.
- Secret environment variables render as `dr-credential:PLACEHOLDER/apiToken` with a comment naming what to substitute. Non-secret values are written literally.

**`manifest/validate.go`**

Walks the same `yaml.Node` tree the reader uses, so every finding carries the line it is on, aggregating into the existing `ValidationError` / `FieldError` types. Rules: exactly one of `imageUri` or `imageBuildConfig` per container, one primary container, port required on the primary and refused elsewhere, primary port at or above 1024, memory in 1000-based units, `replicaCount` XOR autoscaling, runtime group and container names matching their artifact counterparts, entrypoint plus both environment ids for generated builds, `./Dockerfile` present for provided builds, and a non-empty `name`.

**`manifest/node.go`**

`scalarInt` and `isTrue` now `Decode` instead of comparing raw text. Two real bugs: `primary: True` read as false, which produced the actively wrong error "port is only allowed on the primary container", and `port: 8_080` or `0x10` was rejected as "must be a number" despite YAML resolving both. A reader that disagrees with its own parser rejects files that plainly state a number.

## TESTING

- Rule-by-rule validator table tests asserting the expected line number, not just the message.
- `Render` then `Parse` then `Validate` then `Compile` round trip.
- File-mode tests run under umask 022, 077 and 002.
- `task lint` clean on linux, darwin and windows; `task test` green.

---

> [!NOTE]
> **Medium Risk**
> Medium risk from the atomic-write permission change affecting all CLI-written project files, plus new manifest validation/write paths that gate workload config; changes are localized, heavily tested, and not in auth or payment flows.
> 
> **Overview**
> Delivers the **write and validate** half of `internal/workload/manifest` so the setup wizard can emit and check `.datarobot.yaml` before deploy.
> 
> **`internal/fsutil`:** `AtomicWriteFile` moves out of `wapi` (so `manifest` need not import `wapi`). **Permission behavior changes:** new files follow umask via a probe create; replacements keep the existing mode instead of always `chmod` 0644.
> 
> **Manifest writing:** `Draft` + `Render` produce commented YAML (dockerfile / generated / image builds, env vars, secrets as `dr-credential:PLACEHOLDER/...`). `Write` uses `O_EXCL` then atomic write and refuses overwrites; `WriteWorkloadID` patches only `workloadId` while preserving comments and key order.
> 
> **Manifest validation:** `Validate()` walks the same `yaml.Node` tree as compile, returning all findings sorted by line (artifact binding, image source, ports, memory units, scaling, runtime↔artifact name matching, Dockerfile presence, credential shorthand). **`scalarInt` / `isTrue`** now decode scalars so YAML forms like `True` and `8_080` validate correctly.
> 
> **`wapi`** call sites switch to `fsutil.AtomicWriteFile` / `DefaultFileMode` with no intentional semantic change beyond atomic-write permissions.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 37bfdfe9b79670d043288c9ecae3e6aa2050dc86. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/internal/fsutil/atomicwrite.go
+++ b/internal/fsutil/atomicwrite.go
@@ -12,7 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-package wapi
+package fsutil
 
 import (
 	"fmt"
@@ -20,14 +20,50 @@ import (
 	"path/filepath"
 )
 
-const atomicFileMode os.FileMode = 0o644
+// DefaultFileMode is the permission CLI-written data files get: readable by
+// anyone who can read the project, writable by the owner. Exported so callers
+// that open a file themselves (an append-only log, say) stay consistent with
+// what AtomicWriteFile leaves behind.
+const DefaultFileMode os.FileMode = 0o644
 
-// atomicWriteFile writes data to a sibling temp file in the same directory as
+// targetMode is the permission the written file should end up with.
+//
+// Replacing a file keeps the mode it already had: a user who tightened
+// something to 0600 should not find it widened by an unrelated edit. A new
+// file gets what a plain create would give it, umask included, because
+// os.CreateTemp makes the temp file 0600 and chmodding to a fixed mode would
+// hand a developer with a restrictive umask a wider file than everything else
+// they write.
+func targetMode(path string) (os.FileMode, error) {
+	if info, err := os.Stat(path); err == nil {
+		return info.Mode().Perm(), nil
+	}
+
+	// The cheapest way to learn what the umask allows is to let the kernel
+	// apply it, in the directory the file is going to live in.
+	probe, err := os.OpenFile(path, os.O_WRONLY|os.O_CREATE|os.O_EXCL, DefaultFileMode)
+	if err != nil {
+		// Something else created it in between, or the directory refuses; the
+		// rename below reports whichever it is.
+		return DefaultFileMode, nil
+	}
+
+	_ = probe.Close()
+
+	info, err := os.Stat(path)
+	if err != nil {
+		return 0, fmt.Errorf("stat %s: %w", path, err)
+	}
+
+	return info.Mode().Perm(), nil
+}
+
+// AtomicWriteFile writes data to a sibling temp file in the same directory as
 // path, fsyncs it, then renames it over path. os.Rename is atomic on POSIX
 // and handled as replace-on-existing by Go on Windows. The parent directory
 // is fsynced after rename so the new dentry survives a crash. The temp file
 // is removed on any failure before rename so no .tmp.* leftovers remain.
-func atomicWriteFile(path string, data []byte) (err error) {
+func AtomicWriteFile(path string, data []byte) (err error) {
 	dir := filepath.Dir(path)
 
 	tmp, err := os.CreateTemp(dir, filepath.Base(path)+".tmp.*")
@@ -57,7 +93,12 @@ func atomicWriteFile(path string, data []byte) (err error) {
 		return fmt.Errorf("close temp file %s: %w", tmp.Name(), err)
 	}
 
-	if err = os.Chmod(tmp.Name(), atomicFileMode); err != nil {
+	mode, err := targetMode(path)
+	if err != nil {
+		return err
+	}
+
+	if err = os.Chmod(tmp.Name(), mode); err != nil {
 		return fmt.Errorf("chmod temp file %s: %w", tmp.Name(), err)
 	}
 
--- a/internal/fsutil/atomicwrite.go
+++ b/internal/fsutil/atomicwrite.go
@@ -98,12 +98,33 @@ func AtomicWriteFile(path string, data []byte) (err error) {
 		return err
 	}
 
-	if err = os.Chmod(tmp.Name(), mode); err != nil {
-		return fmt.Errorf("chmod temp file %s: %w", tmp.Name(), err)
+	return renameIntoPlace(tmp.Name(), path, dir, mode)
+}
+
+// renameIntoPlace sets the temp file's mode, clears any read-only attribute
+// on the destination (Windows only, see golang/go#38287), renames the temp
+// over the target, and fsyncs the parent directory so the rename is durable.
+// The mode passed in is what targetMode decided the file should end up with:
+// the existing file's preserved mode on a replace, or umask-applied
+// DefaultFileMode on a new file.
+func renameIntoPlace(tmpPath, targetPath, dir string, mode os.FileMode) error {
+	if err := os.Chmod(tmpPath, mode); err != nil {
+		return fmt.Errorf("chmod temp file %s: %w", tmpPath, err)
 	}
 
-	if err = os.Rename(tmp.Name(), path); err != nil {
-		return fmt.Errorf("rename %s to %s: %w", tmp.Name(), path, err)
+	// On Windows, os.Rename fails with "Access is denied" when the
+	// destination is read-only (golang/go#38287). Clear the attribute first
+	// so a read-only file is replaced the way it would be on POSIX. The chmod
+	// above already set the temp file's mode, so the rename lands the right
+	// permissions; targetMode returned the existing file's mode, which on
+	// Windows is 0o444 for a read-only file, so the read-only attribute is
+	// restored by that chmod.
+	if err := clearReadOnlyBeforeRename(targetPath); err != nil {
+		return fmt.Errorf("cannot clear read-only attribute on %s: %w", targetPath, err)
+	}
+
+	if err := os.Rename(tmpPath, targetPath); err != nil {
+		return fmt.Errorf("rename %s to %s: %w", tmpPath, targetPath, err)
 	}
 
 	// Best-effort fsync of the parent directory so the rename is durable.
--- a/internal/fsutil/atomicwrite.go
+++ b/internal/fsutil/atomicwrite.go
@@ -15,6 +15,7 @@
 package fsutil
 
 import (
+	"errors"
 	"fmt"
 	"os"
 	"path/filepath"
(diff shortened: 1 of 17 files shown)


internal/fsutil/atomicwrite.go
ajalon1, line 1: My biggest question is whether this is relevant for Windows as well. I see you're mostly using `os` so I would imagine so.

ajalon1 (commented):
> AtomicWriteFile was already generic and is now needed by manifest, which must not import wapi. Call sites in wapi updated; no behaviour change for them beyond the one below.

👐🏽 

still reviewing, but am concerned about AtomicWriteFile on Windows, doing some testing on my end.

internal/workload/manifest/write.go
```diff
@@ -0,0 +1,564 @@
...
+func number(value int) *yaml.Node {
+	return &yaml.Node{Kind: yaml.ScalarNode, Tag: intTag, Value: strconv.Itoa(value)}
+}
+
+// decimal renders a CPU allocation. A whole number is written plainly (1, not
+// 1.0 and not the !!float tag yaml.v3 would otherwise emit to keep the type):
+// the spec's examples write cpu: 1, and a tagged scalar in a generated file
+// reads like a mistake.
+func decimal(value float64) *yaml.Node {
+	if value == math.Trunc(value) && math.Abs(value) < 1e15 {
```
ajalon1: Why 1e15?

internal/fsutil/atomicwrite.go
ajalon1: Found an issue and a bug in AtomicWrite on Windows. @wojtekwdr rather than wait until tomorrow, since you're strapped for time , I pushed a fix and updated tests to your branch. Hope that our unit test CI will catch this in a Windows runner.

[ce22c23](https://github.com/datarobot-oss/cli/pull/747/commits/ce22c234a146eff381c1b6d6e6479a3b885b9473)

To be reverted per https://datarobot.atlassian.net/browse/CFX-7560

internal/workload/manifest/validate.go
```diff
@@ -0,0 +1,496 @@
...
+func (m *Manifest) Validate() error {
+	v := &validator{dir: m.Dir}
+
+	v.checkName(m.root)
+	v.checkArtifactBinding(m.root)
+
+	groups := v.checkArtifact(mapValue(m.root, keyArtifact))
+	v.checkRuntime(mapValue(m.root, keyRuntime), groups)
+
+	if _, errs := collectCredentialRefs(m.root); len(errs) > 0 {
```
ajalon1, line 126: 🤖 [P1] [security] Cyclic YAML alias → unbounded recursion (DoS)

This `collectCredentialRefs(m.root)` call walks the raw `yaml.Node` tree via `refCollector.walk` → `resolveAlias` (`node.go:32`) with no cycle guard. `yaml.Unmarshal` accepts a self-referencing alias (`runtime: &rt\n  containerGroups: *rt`) without error, so `Parse` succeeds, but the walk loops forever on the cycle.

`root.Decode(&map[string]any)` catches it ("anchor 'rt' value contains itself"), but `Compile` calls `collectCredentialRefs` **before** `Decode`, and `Validate` calls it directly — so the guard never runs first.

Verified end-to-end: `Parse` → `Validate` hangs (3s timeout) on the cyclic payload. The same `resolveAlias`-without-cycle-detection is used by `mapValue`/`seqItems` in the validator's own walk, so a cyclic alias under `artifact:` hangs `Validate` through that path too.

Suggested fix: a one-time cycle check in `Parse` (walk with a `map[*yaml.Node]bool` visited-set, reject any node reachable from itself via aliases) protects every downstream walk.

ajalon1 (changes requested):
👨🏽‍🚀 LGTM in terms of code changes, just two big security findings. Have some notes here for cyclic YAML issue, already pushed a fix for AtomicWrite on Windows for you. What else, 🤖 ?

## Security review findings

Ran a STRIDE + OWASP pass over the diff. One blocking finding (requesting changes), plus the items checked and cleared.

### [P1] [security] Cyclic YAML alias → unbounded recursion (DoS)

`collectCredentialRefs` — called by both `Validate()` and `Compile()` — walks the raw `yaml.Node` tree via `refCollector.walk` → `resolveAlias` (`node.go:32`), with no cycle guard.

`yaml.Unmarshal` into a `yaml.Node` tree **accepts a self-referencing alias without error** — the cycle is only caught by `Decode` into a Go value (`map[string]any`), which runs **after** the raw-tree walk. So the guard that would catch the cycle never runs first.

Verified end-to-end against the production `Parse` → `Validate` path:

```yaml
name: my-app
runtime: &rt
  containerGroups: *rt
```

`Parse` succeeds; both `Validate` and `Compile` hang (unbounded recursion, confirmed with a 3s timeout in a scratch test against the production code). Trigger: a user clones an untrusted repo with this `.datarobot.yaml` and runs `dr workload up` → `Locate` → `Parse` → `Validate`/`Compile`. No auth required beyond opening the repo.

The same `resolveAlias`-without-cycle-detection is used by `mapValue`/`seqItems` in the validator's own tree walk (`checkArtifact`, `checkRuntime`), so a cyclic alias under `artifact:` hangs `Validate` through that path too, not just through `collectCredentialRefs`.

**Suggested fix:** reject alias cycles once at parse time in `Parse` — walk the tree with a `map[*yaml.Node]bool` visited-set and reject any node reachable from itself via aliases. This protects every downstream walk (both `collectCredentialRefs` and the validator's own `mapValue`/`seqItems`), so a single check at the entry point covers all the vulnerable walks.

---

### Checked and cleared (no finding)

- **Atomic write / O_EXCL reservation / rename / temp cleanup** — sound; temp files cleaned on failure via defer; reservation removed if `atomicWrite` fails.
- **Symlink-follow in `targetMode` probe-create** — `O_EXCL` fails on an existing symlink (EEXIST), no creation-through-symlink path.
- **File-mode widening** — the new `targetMode` is an improvement over the old unconditional `chmod 0o644`; new files honour umask, existing files preserve mode.
- **Secret disclosure in rendered manifests** — secrets render as `dr-credential:PLACEHOLDER/apiToken`, never the real value; no sensitive data written to disk.
- **ReDoS in `memoryPattern`** — `^([0-9]+)\s*(B|KB|MB|GB|TB)?$` is linear, no nested quantifiers; Go's RE2 is non-backtracking. Confirmed with large inputs.
- **YAML alias bombs (non-cyclic)** — yaml.v3 stores anchors as single nodes, no exponential expansion; the only YAML risk is cycles (the P1 above).
- **Supply chain** — no `go.mod`/`go.sum` changes in this commit.

internal/workload/manifest/validate.go
ajalon1: Resolve and I'll approve!

wojtekwdr: Fixed with one `checkAliases` pass in `Parse`, so every downstream walk is covered rather than each needing its own guard.

internal/fsutil/atomicwrite.go
```diff
@@ -12,22 +12,58 @@
...
+
+	_ = probe.Close()
+
+	info, err := os.Stat(path)
+	if err != nil {
+		return 0, fmt.Errorf("stat %s: %w", path, err)
+	}
+
+	return info.Mode().Perm(), nil
+}
```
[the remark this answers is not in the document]
wojtekwdr: Fixed: the umask is measured on a unique sibling path (the temp file's name plus `.mode`), so the destination is never created before the rename. Both smaller points too, a non-`ErrNotExist` stat failure now surfaces, and the probe-failed fallback narrows to 0600.

Tests: `TestTargetMode_LeavesTheDestinationAlone`, `TestTargetMode_NarrowsWhenTheProbeFails`, `TestAtomicWriteFile_LeavesNoScratchFiles`.

internal/fsutil/atomicwrite.go
wojtekwdr: Thanks for catching it and pushing the fix. `ce22c23` kept as-is, my changes sit on top and touch only `targetMode`. Noted it reverts per CFX-7560.

internal/workload/manifest/write.go
wojtekwdr: Nothing principled, so it is gone. It guarded the `int64` conversion rather than the cpu value, and `1e15` was an arbitrary stand-in for 2^53.

wojtekwdr:
All four addressed.

- **Cyclic alias (P1):** one `checkAliases` pass in `Parse`. It also rejects non-cyclic alias bombs, which the review cleared but which hang `Validate` too.
- **`targetMode`:** umask measured on a sibling path, never the destination.
- **Windows:** `ce22c23` kept as-is.
- **`1e15`:** gone, with the conversion it guarded.

Lint clean on all three GOOS; `go test -race` green.

ajalon1 (approved):
send it

PR #749, opened 2026-08-11
[CFX-6414] fix(dotenv): make --if-needed skip only when .env is complete

ajalon1:
## RATIONALE

`dr dotenv setup --if-needed` was skipping the wizard when `.env` existed but was missing core variables, as long as those variables were set via OS environment. This left `.env` incomplete and caused downstream failures (e.g., deployment issues reported in CFX-6414).

The root cause: `shouldSkipSetup` used `ValidateEnvironment`, which checks both `.env` file values and `os.LookupEnv`. Environment variables satisfied validation even when `.env` was missing core variables, so the wizard never ran and `.env` was never populated.

## CHANGES

- Added `ValidateEnvironmentFileOnly` and `GatherUserPromptsFromFile` to validate `.env` contents only, ignoring OS environment variables.
- Updated `shouldSkipSetup` to use `ValidateEnvironmentFileOnly` so `--if-needed` skips only when `.env` itself is complete.
- Refactored `promptsWithValues` to accept an `includeEnv` flag and extracted `resolvePromptValue` to reduce cyclomatic complexity.
- Added regression test covering the env-var-only core variables scenario.

## TESTING

- `go test ./cmd/dotenv/... ./internal/envbuilder/... -count=1 -race` — all pass.
- `task lint` — clean across linux/darwin/windows.

## RELATED

- JIRA: CFX-6414
- Slack thread: https://datarobot.slack.com/archives/C0AKJ3SUF55/p1786416272909369

### DIFF
--- a/internal/fsutil/atomicwrite.go
+++ b/internal/fsutil/atomicwrite.go
@@ -12,7 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-package wapi
+package fsutil
 
 import (
 	"fmt"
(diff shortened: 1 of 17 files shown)


ajalon1:
whoops

PR #750, opened 2026-08-11
[CFX-6414] fix(dotenv): make --if-needed skip only when .env is complete

ajalon1:
## RATIONALE

`dr dotenv setup --if-needed` was skipping the wizard when `.env` existed but was incomplete, leaving `.env` incomplete and causing downstream failures (e.g., deployment issues reported in CFX-6414).

Two root causes, both in the skip check (`shouldSkipSetup`):

1. **OS environment leak:** `shouldSkipSetup` used `ValidateEnvironment`, which checks both `.env` file values and `os.LookupEnv`. Environment variables satisfied validation even when `.env` was missing core variables, so the wizard never ran and `.env` was never populated.
2. **Synthesis leak:** even after switching to a file-only validator, `gatherUserPrompts` still applied YAML `default` values, ran `ApplyGeneratedValues` (auto-generating secrets), and `Valid()` accepted non-empty commented entries. A required variable satisfied only by a default, a generated secret, or a commented-out `.env` line could make the skip check pass, leaving `.env` incomplete.

## CHANGES

- Added `ValidateEnvironmentFileOnly` to validate `.env` contents only, ignoring OS environment variables.
- Updated `shouldSkipSetup` to use `ValidateEnvironmentFileOnly` so `--if-needed` skips only when `.env` itself is complete.
- Made `ValidateEnvironmentFileOnly` **presence-only**: it no longer applies YAML defaults, does not run `ApplyGeneratedValues`, and treats commented-out `.env` entries as absent. A required variable now counts as present only if it has an effective, uncommented value in the `.env` file.
- Threaded a `synthesize` flag through `gatherUserPrompts` / `validateEnvironment` / `promptsWithValues` / `resolvePromptValue`. `GatherUserPrompts` and `ValidateEnvironment` keep `synthesize=true` (unchanged behavior); `ValidateEnvironmentFileOnly` uses `synthesize=false`.
- Removed the unused `GatherUserPromptsFromFile` public wrapper; the `includeEnv`/`synthesize` distinction stays private to the `envbuilder` package.
- Refactored `promptsWithValues` to accept `includeEnv`/`synthesize` flags and extracted `resolvePromptValue` to reduce cyclomatic complexity.
- The `PULUMI_CONFIG_PASSPHRASE` viper-config value remains an intentional non-file exception (`resolvePulumiPassphrase`), documented as such.
- Regression tests cover the OS-env case and the three synthesis leaks (YAML default, generated secret, commented entry) plus a positive control.

## BEHAVIOR CHANGE NOTE

`dr dotenv setup --if-needed` now runs the wizard in three cases where it previously skipped (exit 0): a required var satisfied only by a YAML `default`, a required `generate: true` secret absent from `.env`, and a required var commented out in `.env`. This is a bug fix (the `.env` was incomplete), but it is an observable change to a user-facing flag.

For **automation/CI** that runs `dr dotenv setup --if-needed` non-interactively, pair it with `--yes` (or `DATAROBOT_CLI_NON_INTERACTIVE=true`) so the wizard runs headless and writes the missing values, then exits 0. A script using `--if-needed` without `--yes` that previously got a silent skip for one of these cases will now launch the wizard and block on a prompt.

`dr dotenv validate` is unchanged — it still uses `ValidateEnvironment` (`synthesize=true`), so a var satisfied only by a default still passes `validate`.

## TESTING

- `go test ./cmd/dotenv/... ./internal/envbuilder/... -count=1 -race` — all pass, including `TestShouldSkipSetup_PresenceOnly` (3 leak cases + positive control).
- `task lint` — clean across linux/darwin/windows.
- Manual CLI exercise of 7 scenarios (baseline skip, missing core var, OS-env-doesn't-satisfy, YAML-default-doesn't-satisfy, commented-doesn't-satisfy, `validate` unchanged) — all pass.

## RELATED

- JIRA: CFX-6414
- Slack thread: https://datarobot.slack.com/archives/C0AKJ3SUF55/p1786416272909369
- Related: CFX-7263 / PR #751 (fail on bad env credentials) — complementary at the auth layer; #750 makes the `.env` genuinely complete so #751's "complete env pair = authoritative" rule fires on correct data.

---

> [!NOTE]
> **Medium Risk**
> Changes when dotenv setup is skipped in automation paths; incorrect skip behavior previously caused deployment failures, so regressions here could reintroduce incomplete `.env` files or force unexpected wizard runs.
>
> **Overview**
> Fixes `dr dotenv setup --if-needed` skipping the wizard when required vars were present only in the OS environment, leaving `.env` incomplete and breaking downstream deploys.
>
> `shouldSkipSetup` now uses new **file-only** validation (`ValidateEnvironmentFileOnly` / `GatherUserPromptsFromFile`) that ignores `os.LookupEnv`, so an incomplete `.env` still triggers setup. Prompt value resolution was refactored with an `includeEnv` flag and `resolvePromptValue`, and a regression test covers the env-var-only case.
>
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit ac8d5d7d6267ccfbd456f2bcd83d9b46a0f7e490. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/internal/envbuilder/builder.go
+++ b/internal/envbuilder/builder.go
@@ -256,17 +256,17 @@ func ApplyGeneratedValues(prompts []UserPrompt) ([]UserPrompt, error) {
 }
 
 func GatherUserPrompts(rootDir string, variables Variables) ([]UserPrompt, error) {
-	return gatherUserPrompts(rootDir, variables, true)
+	return gatherUserPrompts(rootDir, variables, true, true)
 }
 
-// GatherUserPromptsFromFile gathers user prompts using only .env file values,
-// ignoring OS environment variables. This is used for --if-needed skip checks
-// where we want to know whether the .env file itself is complete.
-func GatherUserPromptsFromFile(rootDir string, variables Variables) ([]UserPrompt, error) {
-	return gatherUserPrompts(rootDir, variables, false)
-}
-
-func gatherUserPrompts(rootDir string, variables Variables, includeEnv bool) ([]UserPrompt, error) {
+// gatherUserPrompts collects and resolves user prompts from the repository.
+// When includeEnv is true, OS environment variables override .env file values.
+// When synthesize is true, YAML defaults are applied and secret values are
+// auto-generated for prompts with generate: true. When synthesize is false
+// (presence-only mode), only effective, uncommented .env file entries count —
+// defaults are not applied and no values are generated. The PULUMI_CONFIG_PASSPHRASE
+// viper-config exception applies in both modes (see resolvePulumiPassphrase).
+func gatherUserPrompts(rootDir string, variables Variables, includeEnv, synthesize bool) ([]UserPrompt, error) {
 	yamlFiles, err := Discover(rootDir, 5)
 	if err != nil {
 		return nil, fmt.Errorf("Failed to discover task yaml files: %w", err)
@@ -285,11 +285,13 @@ func gatherUserPrompts(rootDir string, variables Variables, includeEnv bool) ([]
 		allPrompts = append(allPrompts, prompts...)
 	}
 
-	allPrompts = promptsWithValues(allPrompts, variables, includeEnv)
+	allPrompts = promptsWithValues(allPrompts, variables, includeEnv, synthesize)
 
-	allPrompts, err = ApplyGeneratedValues(allPrompts)
-	if err != nil {
-		return nil, err
+	if synthesize {
+		allPrompts, err = ApplyGeneratedValues(allPrompts)
+		if err != nil {
+			return nil, err
+		}
 	}
 
 	allPrompts = DetermineRequiredSections(allPrompts)
--- a/internal/envbuilder/builder.go
+++ b/internal/envbuilder/builder.go
@@ -256,6 +256,17 @@ func ApplyGeneratedValues(prompts []UserPrompt) ([]UserPrompt, error) {
 }
 
 func GatherUserPrompts(rootDir string, variables Variables) ([]UserPrompt, error) {
+	return gatherUserPrompts(rootDir, variables, true)
+}
+
+// GatherUserPromptsFromFile gathers user prompts using only .env file values,
+// ignoring OS environment variables. This is used for --if-needed skip checks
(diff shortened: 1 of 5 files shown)


internal/envbuilder/builder.go
```diff
@@ -256,6 +256,17 @@ func ApplyGeneratedValues(prompts []UserPrompt) ([]UserPrompt, error) {
 }
 
 func GatherUserPrompts(rootDir string, variables Variables) ([]UserPrompt, error) {
+	return gatherUserPrompts(rootDir, variables, true)
+}
+
+// GatherUserPromptsFromFile gathers user prompts using only .env file values,
+// ignoring OS environment variables. This is used for --if-needed skip checks
+// where we want to know whether the .env file itself is complete.
+func GatherUserPromptsFromFile(rootDir string, variables Variables) ([]UserPrompt, error) {
```
chasdr: what calls this ?

chasdr (approved):

internal/envbuilder/builder.go
ajalon1: Only a figment of my imagination, I meant to call this from ValidateEnvironment and then wrote ValidateEnvironmentFileOnly and forgot.

Let me roll it back

ajalon1: 3d4545c

internal/envbuilder/validator.go
```diff
@@ -86,12 +86,23 @@ func (r EnvironmentValidationError) Error() string {
...
+}
+
+func validateEnvironment(repoRoot string, variables Variables, includeEnv bool) EnvironmentValidationError {
 	result := EnvironmentValidationError{
 		Results: make([]ValidationResult, 0),
 	}
 
 	// Gather all user prompts from the repository
-	userPrompts, err := GatherUserPrompts(repoRoot, variables)
+	userPrompts, err := gatherUserPrompts(repoRoot, variables, includeEnv)
```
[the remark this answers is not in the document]
ajalon1: Going to pull this back and work on this bit tomorrow.

PR #759, opened 2026-08-13
[RAPTOR-18976] feat(workload): read the manifest and the live workload

wojtekwdr:
# RATIONALE

First of four in the `dr workload up` stack ([RAPTOR-18976](https://datarobot.atlassian.net/browse/RAPTOR-18976)). `up` is plan-then-apply, and this is the reading and the deciding: turn a directory and a workload id into facts, then work out what differs. No output, no cobra, no API mutation. Nothing calls it yet, so there is no user-visible change.

Splitting reading from deciding is deliberate. `Load` turns a directory into a validated, compiled manifest; `Look` turns a workload id into the live spec, the live runtime and a `State`. Neither judges what it finds: a missing manifest and a deleted workload are both reported as facts, because whether either is fatal depends on what the caller is allowed to do about it, and only the command knows that.

## CHANGES

`internal/workload/up`, new package:

- `load.go` — find the manifest by walking upward the way git does, validate and compile it
- `look.go` — two GETs for the live workload and its artifact, reduced to a `Live`
- `state.go` — the live status vocabulary as a `State` the plan can switch on
- `diff.go` — `Subset`, the one-directional comparison
- `plan.go` — `Plan`, `Build`, `Action()`, `Empty()`
- `doc.go` — the package contract, including why none of this is a hash

## NOTES

**Drift is one-directional and measured against the running object.** Whatever the file says, `up` makes true; whatever the file leaves out, `up` leaves alone. So a workload can carry a GPU bundle, a sidecar and an autoscaling policy this CLI has no vocabulary for and none of it is drift, while a UI edit to a field the file does name is reported and restored. Deleting a line stops managing a field rather than reverting it.

**Not a stored hash of the last applied spec**, which is the obvious cheaper design. It cannot see the UI edit, and it is absent in a fresh CI clone. Comparing against the live object costs two GETs and is correct in both cases.

**Compared on the compiled payload, not the file**, so the `dr-credential:` shorthand is already expanded and a credential reference does not read as a permanent difference against the object form the platform stores. And on the server's own documents rather than typed structs, because a typed round trip silently drops every field this release has not heard of, and half the point of the manifest format is that those survive.

**Lists are matched by name, not by index**, so reordering `containerGroups` or `environmentVars` in the file is not drift.

## RELATED

[RAPTOR-18976](https://datarobot.atlassian.net/browse/RAPTOR-18976), sub-task 5 of [RAPTOR-18970](https://datarobot.atlassian.net/browse/RAPTOR-18970). Stack: **1/4 (this)** → 2 render → 3 orchestrate → 4 command. Each PR is based on the one before it, so review this one first and merge in order.

Made with [Cursor](https://cursor.com)

### DIFF
--- a/internal/workload/up/diff.go
+++ b/internal/workload/up/diff.go
@@ -1,0 +1,228 @@
+// Copyright 2026 DataRobot, Inc. and its affiliates.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package up
+
+import (
+	"fmt"
+	"maps"
+	"reflect"
+	"slices"
+	"strings"
+)
+
+// Change is one field the manifest asks for that the live object does not
+// already agree with.
+type Change struct {
+	// Path is where the field sits, in the file's own spelling, with
+	// name-keyed lists addressed by name rather than by index:
+	// artifact.spec.containerGroups[default].containers[primary].port.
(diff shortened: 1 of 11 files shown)


victorborshak (approved):
Looks good to me

PR #764, opened 2026-08-13
[RAPTOR-18976] feat(workload): verify credentials before the deploy mutates

wojtekwdr:
Fifth of the `dr workload up` stack, on top of #762. Review #759 → #760 → #761 → #762 first.

## What

A credential id is the one part of the manifest local validation cannot judge. The ledger checks that the `dr-credential:` shorthand parses; whether the id behind it names a real credential is a fact only the platform holds. Left unchecked, a typo survives every local check and surfaces as a container that will not start, minutes and one paid-for build later, with a runtime error that never mentions the manifest.

`Compile` has been handing back every reference for exactly this since the manifest package landed. This wires the GET that the note at the foot of `run.go` was waiting for, and removes the note.

## Decisions worth a look

- **Where it sits.** At the end of the apply gate, immediately before the first mutation, so a run that is going to be refused for some other reason does not spend the lookups first.
- **A dry run never reaches it.** It mutates nothing, and failing a `--dry-run` over a file it was asked only to describe would be the wrong contract.
- **Every problem at once**, matching what the validation ledger does with the rest of the file. These references come from something the user is about to edit; learning about the second bad id after fixing the first is two trips through a build pipeline for one edit.
- **Each line names the variable and the manifest line**, not only the id, because the id alone is a hex string the reader then has to hunt for.
- **One GET per credential, not per reference.** Several variables commonly read different keys of the same stored credential.
- **The wizard's placeholder is refused without asking the platform**, which is why the wizard writes it in full instead of commenting the entry out.
- **A transport failure is kept distinct from a missing credential.** Reporting "does not exist" because the platform was briefly unreachable would send someone editing a correct file.

The eleventh test seam pushed `install` past the cyclomatic limit, so it becomes a list of `swap` calls over a shared generic helper rather than a ladder of ifs and a parallel struct of saved values.

## Test plan

- [x] `task test` — no new failures (the `cmd/dotenv`, `cmd/plugin/*` and feature-gate failures pre-exist on the base branch)
- [x] `task lint` — clean on all three GOOS targets

---

> [!NOTE]
> **Medium Risk**
> Touches the apply gate on the workload deploy path and adds network dependency right before mutation, though behavior is gated, well-tested, and fails closed without creating resources on bad refs.
> 
> **Overview**
> **`dr workload up` now validates every `dr-credential` reference against the platform immediately before the first create**, so bad ids or unfinished wizard placeholders fail in the CLI instead of after a build.
> 
> The new `verifyCredentials` step uses one GET per distinct credential id, aggregates every problem (missing id, placeholder, env var name, manifest line) in one error, and treats API outages as “cannot verify” rather than “does not exist.” **Dry runs still stop after planning and never perform these lookups.** The pending note in `run.go` is removed now that `Compiled.CredentialRefs` is wired through a `getCredentialFn` seam.
> 
> Test helpers in `run_test` were refactored to generic `swap`/`force` so the extra credential seam does not blow past cyclomatic limits; integration tests assert create is skipped on failure and that dry run does not hit the API.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 109c5273e3675f61b7196bc5f54fedee20c47ee7. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/internal/workload/up/credentials.go
+++ b/internal/workload/up/credentials.go
@@ -1,0 +1,99 @@
+// Copyright 2026 DataRobot, Inc. and its affiliates.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package up
+
+import (
+	"fmt"
+	"strings"
+
+	"github.com/datarobot/cli/internal/workload/manifest"
+)
+
+// verifyCredentials confirms that every credential the manifest references
+// actually exists, and is the last check before anything mutates.
+//
+// A credential reference is the one part of the manifest local validation
+// cannot judge. The ledger checks that the shorthand parses; whether the id
+// behind it names a real credential is a fact only the platform holds. Left
+// unchecked, a mistyped id survives the whole deploy and surfaces as a
+// container that will not start, minutes and one paid-for build later, with
+// a runtime error that does not mention the manifest.
+//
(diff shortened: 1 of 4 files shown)


victorborshak (approved):
LGTM

PR #751, opened 2026-08-12
[CFX-7263] Fail on bad env credentials instead of using the stored profile

chasdr:
## Summary

`dr` honored `DATAROBOT_ENDPOINT`/`DATAROBOT_API_TOKEN` only once they verified. On failure it silently fell back to the stored CLI profile, so a script passing a project's `.env` credentials ran against the caller's own instance with exit 0. `dr llm-gateway list` listed the wrong catalog. A complete env pair that fails verification now fails the command, with the reason on stderr and both endpoints named. `dr auth check` and `dr auth export` already treat env credentials as authoritative.

## Notes for review

- A stale `.env` in a template checkout used to self-heal: `task` loads `.env` into the environment, and `dr dotenv update` quietly ran on the stored profile. Those runs now fail with the fix named. pulumi and the SDKs already fail on the same stale credentials.
- Partial env credentials (one var set) keep the old fallback. Only a complete pair is an explicit instance request.

## Output

```
$ DATAROBOT_ENDPOINT=http://127.0.0.1:8471/api/v2 DATAROBOT_API_TOKEN=expired-token dr llm-gateway list
❌ DATAROBOT_API_TOKEN environment variable is invalid or expired.
Unset it and try again:
  unset DATAROBOT_API_TOKEN (or Remove-Item Env:\DATAROBOT_API_TOKEN on Windows)
Environment credentials for http://127.0.0.1:8471 failed to verify; not falling back to the stored profile for http://127.0.0.1:8600.
Error: Authentication failed.

$ DATAROBOT_ENDPOINT=http://127.0.0.1:9 DATAROBOT_API_TOKEN=some-token dr llm-gateway list
❌ Could not connect to http://127.0.0.1:9: dial tcp 127.0.0.1:9: connect: connection refused
Check DATAROBOT_ENDPOINT and your network, then try again.
Error: Authentication failed.
```

Before this change both commands printed the caller's own catalog and exited 0.

## Technical Changes

- internal/auth/auth.go: `EnsureAuthenticated` fails before the stored-profile check when a complete env pair fails verification. Classification extracted to `ReportEnvCredentialsError` with a new transport-error branch. `EnvCredentials.EndpointVar` names `DATAROBOT_API_ENDPOINT` when that fallback supplied the endpoint. Styles bind to the destination writer so redirected stderr gets no ANSI codes.
- cmd/auth/check/cmd.go: drops its copy of the classification for the shared helper.
- docs: authentication.md documents the precedence; auth.md adds the could-not-connect example.
[...]

PR #772, opened 2026-08-13
[CFX-7585] Stop reporting server errors and bad endpoint schemes as credential problems

chasdr:
## Summary

`dr` told users to throw away a working API token whenever the version check came back as anything other than 200. A 404 from a mistyped endpoint, a 429, or a 503 all printed "DATAROBOT_API_TOKEN is invalid or expired, unset it". Only 401 and 403 still say that; any other status now names what the instance answered. And an endpoint scheme the CLI cannot use (`ftp://host`) reads as a bad endpoint instead of a network failure. Both were suppressed review findings on #751, which made them visible on every command.

## Notes for review

The scheme check went into `ValidateEndpoint`, not into the `SchemeHostOnly` helper that `set-url` and `export` share, so those keep accepting what they always did.

Deliberately out of scope, same defect family, follow-up material: `dr auth check`'s `.env` and stored-profile legs still call every failure a stale token or missing key, `EnsureAuthenticated` still starts the login flow when the stored profile's instance answers 5xx, and `dr auth set-url ftp://host` still writes the endpoint unvalidated.

## Output

```
$ DATAROBOT_ENDPOINT=http://127.0.0.1:51801/api/v2 dr auth check   # instance returning 503
- ❌ DATAROBOT_API_TOKEN environment variable is invalid or expired.
- Unset it and try again:
-   unset DATAROBOT_API_TOKEN (or Remove-Item Env:\DATAROBOT_API_TOKEN on Windows)
+ ❌ http://127.0.0.1:51801 answered HTTP 503, so the CLI could not verify your credentials.
+ Check DATAROBOT_ENDPOINT, and the instance's status if it persists.

$ DATAROBOT_ENDPOINT=ftp://example.com/api/v2 dr auth check
- ❌ Could not connect to ftp://example.com: unsupported protocol scheme "ftp"
- Check DATAROBOT_ENDPOINT and your network, then try again.
+ ❌ DATAROBOT_ENDPOINT environment variable is invalid: unsupported URL scheme "ftp", use https://
+ Set it to a valid DataRobot URL and try again.
```

## Technical Changes

- `internal/config/auth.go`: `VerifyToken` returns `*HTTPStatusError{StatusCode}` instead of a flat `invalid token`.
- `internal/auth/auth.go`: `ValidateEndpoint` requires http or https; the classifier's fallthrough splits on the status.
- Tests pin the status preservation, its consumers (`GetAPIKey`, the `.env` leg of `dr auth check`), and 401/403 keeping the exact #751 wording.
- `docs/`: the two new messages.

---

> [!NOTE]
> **Medium Risk**
> Changes shared auth verification and error reporting used on every authenticated command; behavior is narrower (better classification) but mistakes could mis-route users on edge HTTP statuses.
> 
> **Overview**
> Fixes misleading **“invalid or expired API token”** messages when environment credentials fail for reasons other than rejected auth.
> 
> **`VerifyToken`** now returns **`*HTTPStatusError`** with the real status instead of a generic invalid-token error. **`ReportEnvCredentialsError`** (used by **`EnsureAuthenticated`** and **`dr auth check`** for env vars) only tells users to unset **`DATAROBOT_API_TOKEN`** on **401** or **403**; for other non-2xx responses it reports that the instance **answered HTTP *n*** and points at endpoint/instance health. **`ValidateEndpoint`** additionally rejects schemes other than **http/https** (e.g. **`ftp://`**) as a bad endpoint, not a connection failure.
> 
> Docs and tests cover status preservation, **`GetAPIKey`**, scheme validation, and the new user-facing copy. The **`.env`** path in **`dr auth check`** and stored-profile login behavior are **unchanged** for non-401 failures (called out in code as follow-up).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 54b5b4a64ecee28d55f77bcdf846847b61d6b3ee. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/internal/auth/auth.go
+++ b/internal/auth/auth.go
@@ -357,7 +357,8 @@ func EnsureAuthenticated(ctx context.Context) bool { //nolint: cyclop
 		return false
 	}
 
-	// No valid token, attempt to get one
+	// Reached for any non-timeout failure, including a 5xx from the instance.
+	// Restricting the login flow to rejected credentials is a planned follow-up.
 	log.Warn("No valid API key found. Starting authentication flow...")
 
 	// Auto-retrieve new credentials without prompting
--- a/internal/auth/auth.go
+++ b/internal/auth/auth.go
@@ -19,6 +19,7 @@ import (
 	"errors"
 	"fmt"
 	"io"
+	"net/http"
 	"net/url"
 	"os"
 	"strings"
@@ -117,20 +118,24 @@ func GetEnvCredentials() EnvCredentials {
 	}
 }
 
-// ValidateEndpoint returns a non-nil error when endpoint is not a usable
-// DataRobot URL (for example, it is wrapped in literal surrounding quotes that
-// make url.Parse fail). It lets callers distinguish a malformed
-// DATAROBOT_ENDPOINT from an expired/invalid token when reporting why
-// environment credentials failed verification. An empty endpoint returns nil
-// (that case is ErrEnvCredentialsNotSet, handled by the caller).
+// ValidateEndpoint rejects an endpoint the CLI cannot use: one that will not
+// parse (surrounding quotes) or a scheme other than http/https. Empty is nil.
 func ValidateEndpoint(endpoint string) error {
 	if endpoint == "" {
 		return nil
 	}
 
-	_, err := config.SchemeHostOnly(endpoint)
+	baseURL, err := config.SchemeHostOnly(endpoint)
+	if err != nil {
+		return err
+	}
 
-	return err
+	// Checked here, not in SchemeHostOnly, which set-url and export share.
+	if scheme, _, _ := strings.Cut(baseURL, "://"); scheme != "http" && scheme != "https" {
+		return fmt.Errorf("unsupported URL scheme %q, use https://", scheme)
+	}
+
+	return nil
 }
 
 // ReportEnvCredentialsError writes a classified explanation of why an
@@ -198,6 +203,23 @@ func ReportEnvCredentialsError(w io.Writer, creds *EnvCredentials, err error) {
 		fmt.Fprint(w, info.Render(envDatarobotHost))
 		fmt.Fprintln(w, base.Render(": "+urlErr.Err.Error()))
 		fmt.Fprintln(w, base.Render("Check "+creds.endpointVarName()+" and your network, then try again."))
+
+		return
+	}
+
+	// Only 401 and 403 mean the token was judged and rejected. Blaming it for
+	// a 404, 429, or 5xx would tell the user to unset working credentials.
+	var statusErr *config.HTTPStatusError
+	if errors.As(err, &statusErr) &&
+		statusErr.StatusCode != http.StatusUnauthorized &&
+		statusErr.StatusCode != http.StatusForbidden {
+		envDatarobotHost, _ := config.SchemeHostOnly(creds.Endpoint)
+
+		fmt.Fprint(w, base.Render("❌ "))
+		fmt.Fprint(w, info.Render(envDatarobotHost))
+		fmt.Fprintln(w, base.Render(fmt.Sprintf(" answered HTTP %d, so the CLI could not verify your credentials.",
+			statusErr.StatusCode)))
+		fmt.Fprintln(w, base.Render("Check "+creds.endpointVarName()+", and the instance's status if it persists."))
 
 		return
 	}
--- a/internal/config/auth.go
+++ b/internal/config/auth.go
@@ -17,6 +17,7 @@ package config
 import (
 	"context"
 	"errors"
+	"fmt"
 	"net/http"
 	"net/url"
 	"time"
@@ -42,7 +43,18 @@ func getUserAgent(ctx context.Context) string {
 	return GetUserAgentHeader()
 }
 
+// HTTPStatusError is a non-200 answer to the version check, so callers can tell
+// a rejected token (401/403) from a server or endpoint problem.
+type HTTPStatusError struct {
+	StatusCode int
+}
+
+func (e *HTTPStatusError) Error() string {
+	return fmt.Sprintf("token verification returned HTTP %d", e.StatusCode)
+}
+
 // VerifyToken verifies if the datarobot endpoint + api key pair correspond to a valid pair.
+// A non-200 response is returned as *HTTPStatusError.
 func VerifyToken(ctx context.Context, datarobotEndpoint, token string) error {
 	_, err := url.Parse(datarobotEndpoint)
 	if err != nil {
@@ -80,7 +92,7 @@ func VerifyToken(ctx context.Context, datarobotEndpoint, token string) error {
 	defer resp.Body.Close()
 
 	if resp.StatusCode != http.StatusOK {
-		return errors.New("invalid token")
+		return &HTTPStatusError{StatusCode: resp.StatusCode}
 	}
 
 	return nil
--- a/cmd/auth/check/cmd.go
+++ b/cmd/auth/check/cmd.go
@@ -159,6 +159,8 @@ func verifyDotenvToken(dotenvEndpoint, dotenvToken string) bool {
 			fmt.Print(tui.InfoStyle.Render(dotenvBaseURL))
 			fmt.Println(tui.BaseTextStyle.Render(" timed out. Check your network and try again."))
 		} else {
+			// Blames the token for any non-200, unlike the env-var leg.
+			// Splitting on the status here is a planned follow-up.
 			fmt.Println(tui.BaseTextStyle.Render("❌ DATAROBOT_API_TOKEN in '.env' is invalid or expired."))
 			fmt.Print(tui.BaseTextStyle.Render("Run "))
 			fmt.Print(tui.InfoStyle.Render("dr dotenv update"))
--- a/cmd/auth/check/cmd_test.go
+++ b/cmd/auth/check/cmd_test.go
@@ -52,6 +52,9 @@ func TestVerifyDotenvToken_StatusErrorKeepsMessage(t *testing.T) {
 	t.Cleanup(server.Close)
 
 	orig := os.Stdout
+
+	t.Cleanup(func() { os.Stdout = orig })
+
 	r, w, err := os.Pipe()
 	require.NoError(t, err)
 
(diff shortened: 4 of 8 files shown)


chasdr:
bugbot run

chasdr:
bugbot run

chasdr:
bugbot run

internal/auth/auth.go
```diff
@@ -117,20 +118,24 @@ func GetEnvCredentials() EnvCredentials {
...
+	baseURL, err := config.SchemeHostOnly(endpoint)
+	if err != nil {
+		return err
+	}
 
-	return err
+	// Checked here, not in SchemeHostOnly, which set-url and export share.
+	if scheme, _, _ := strings.Cut(baseURL, "://"); scheme != "http" && scheme != "https" {
+		return fmt.Errorf("unsupported URL scheme %q, use https://", scheme)
+	}
```
ajalon1, line 136: I was a little weirded out by this -- especially considering localhost and docker.host.internal -- but I guess those are all already handled and the new error message does help with resolving that.

internal/auth/auth.go
```diff
@@ -117,20 +118,24 @@ func GetEnvCredentials() EnvCredentials {
...
 }
 
-// ValidateEndpoint returns a non-nil error when endpoint is not a usable
-// DataRobot URL (for example, it is wrapped in literal surrounding quotes that
-// make url.Parse fail). It lets callers distinguish a malformed
-// DATAROBOT_ENDPOINT from an expired/invalid token when reporting why
-// environment credentials failed verification. An empty endpoint returns nil
-// (that case is ErrEnvCredentialsNotSet, handled by the caller).
+// ValidateEndpoint rejects an endpoint the CLI cannot use: one that will not
+// parse (surrounding quotes) or a scheme other than http/https. Empty is nil.
```
ajalon1, line 122: better doc :)

internal/auth/auth.go
```diff
@@ -202,6 +207,23 @@ func ReportEnvCredentialsError(w io.Writer, creds *EnvCredentials, err error) {
 		return
 	}
 
+	// Only 401 and 403 mean the token was judged and rejected. Blaming it for
+	// a 404, 429, or 5xx would tell the user to unset working credentials.
+	var statusErr *config.HTTPStatusError
+	if errors.As(err, &statusErr) &&
+		statusErr.StatusCode != http.StatusUnauthorized &&
+		statusErr.StatusCode != http.StatusForbidden {
```
ajalon1, line 215: Honestly not sure about 403 here -- I don't know off hand if we even return 403 on /version.

I'll let this slide if this only works on the /version path, and isn't used by `internal/drapi` where we can and do return 403s (some are masked as 404s).

I think the error rendered below for 403s could be tweaked a bit since you're here.

internal/config/auth.go
```diff
@@ -80,7 +92,7 @@ func VerifyToken(ctx context.Context, datarobotEndpoint, token string) error {
 	defer resp.Body.Close()
 
 	if resp.StatusCode != http.StatusOK {
-		return errors.New("invalid token")
+		return &HTTPStatusError{StatusCode: resp.StatusCode}
```
ajalon1, line 95: ugh so nice

internal/auth/auth.go
```diff
@@ -202,6 +207,23 @@ func ReportEnvCredentialsError(w io.Writer, creds *EnvCredentials, err error) {
...
+		envDatarobotHost, _ := config.SchemeHostOnly(creds.Endpoint)
+
+		fmt.Fprint(w, base.Render("❌ "))
+		fmt.Fprint(w, info.Render(envDatarobotHost))
```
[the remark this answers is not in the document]
ajalon1: could be a separate PR, I think this happens elsewhere

ajalon1 (approved):
👨🏽‍🚀 Nice add. I'm actually surprised we didn't handle this properly before, since auth is some of the oldest code in the CLI.

LGTM

ajalon1:
👨🏽‍🚀 But you still deserve a 🤖 pass.

**Verdict: approve — one suggestion worth considering for a follow-up.** The core change is correct and well-tested: `VerifyToken` now returns a typed `*HTTPStatusError` carrying the status code, and `ReportEnvCredentialsError` uses it to stop blaming the token for 404/429/5xx. The one gap is that the `.env` leg and stored-profile leg of `dr auth check` / `EnsureAuthenticated` were left with the old "any non-200 = bad token" behavior.

---

### Praise

**The 401/403 vs. everything-else split is the right call, verified end-to-end.** Before this PR, a 503 from the instance told the user to `unset DATAROBOT_API_TOKEN` — actively destructive advice that would discard working credentials. The new `*HTTPStatusError` type carries the status code, and `ReportEnvCredentialsError` only blames the token for 401/403. I confirmed the classification for 401, 403, 404, 429, and 500 via the test suite (`TestReportEnvCredentialsError` subtests).

**The scheme check is placed in `ValidateEndpoint`, not `SchemeHostOnly` — with a comment explaining why.** `SchemeHostOnly` is shared by `dr auth set-url` and `dr auth export`, which have different requirements. Putting the http/https-only guard in `ValidateEndpoint` keeps it scoped to env-credential validation. The comment (`"Checked here, not in SchemeHostOnly, which set-url and export share."`) makes the reasoning explicit.

<details>
<summary>🤖 Classification chain in <code>ReportEnvCredentialsError</code></summary>

The function is a priority-ordered error classifier. Each check returns early; the first match wins:

```
err from VerifyToken
    │
    ├─ 1. context.DeadlineExceeded        → "Connection to <host> timed out"
    ├─ 2. ValidateEndpoint(endpoint) fails → "endpoint is invalid: <reason>"
    ├─ 3. endpoint has no "://"           → "endpoint is invalid: missing URL scheme"
    ├─ 4. *url.Error, Op=="parse"         → "endpoint is invalid: <parse error>"
    ├─ 5. *url.Error (transport)          → "Could not connect to <host>: <reason>"
    ├─ 6. *HTTPStatusError, ≠401 && ≠403  → "<host> answered HTTP <status>"  ← NEW
    └─ 7. default (incl. 401/403)         → "DATAROBOT_API_TOKEN is invalid or expired"
```

The new step 6 is inserted **after** the transport checks (step 5) and **before** the token-blaming default (step 7). This ordering is correct: a `*HTTPStatusError` is not a `*url.Error`, so `errors.As(err, &urlErr)` returns false for it and it falls through to step 6. A 401 or 403 fails the `≠401 && ≠403` guard and falls through to step 7 (token blamed), which is the intended behavior.

</details>

---

### Suggestion (optional, not requesting a change in this PR)

**S1: The `.env` leg and stored-profile leg still blame the token for server errors.**

`verifyDotenvToken` in `cmd/auth/check/cmd.go` only checks `errors.Is(err, context.DeadlineExceeded)`; every other non-200 falls into the else branch and prints `"DATAROBOT_API_TOKEN in '.env' is invalid or expired."` — the exact misattribution this PR fixes for the env-var leg.

<details>
<summary>🤖 Verified empirically with a scratch test (cleaned up after)</summary>

```
=== RUN   TestScratch_DotEnv503StillBlamesToken
--- PASS: TestScratch_DotEnv503StillBlamesToken (0.00s)
PASS
```

The test confirmed: `assert.Contains(t, out, "DATAROBOT_API_TOKEN in '.env' is invalid or expired")` passes, and `assert.NotContains(t, out, "answered HTTP 503")` passes.

</details>

The added test `TestVerifyDotenvToken_StatusErrorKeepsMessage` only covers 401, which is correct for 401 but doesn't expose the 503/404/429 gap. The test comment says *"the typed `*config.HTTPStatusError` must leave its message and verdict unchanged"* — so this appears to be an intentional scoping decision, but there's no comment in `verifyDotenvToken` itself explaining why the `.env` leg should behave differently from the env-var leg.

The same pattern exists in the stored-profile leg of `EnsureAuthenticated` (Copilot also flagged this at `auth.go:253`): `config.GetAPIKey` returns `*HTTPStatusError` for non-200, but `EnsureAuthenticated` only checks `errors.Is(viperErr, context.DeadlineExceeded)` and otherwise clears the token and starts the login flow. A 503 on the stored profile triggers login — the exact misbehavior the PR title claims to fix, on the path every command hits via `EnsureAuthenticatedE`.

If the intent is to keep these legs simple for now, a one-line comment in each would make that explicit. If the intent is to extend the fix in a follow-up, the tests should cover non-401 statuses.

---

### Nits

<details>
<summary>🤖 Minor, all non-blocking</summary>

**N1: `TestVerifyDotenvToken_StatusErrorKeepsMessage` restores `os.Stdout` manually, not via `t.Cleanup`.** If `verifyDotenvToken` panicked between `os.Stdout = w` and `os.Stdout = orig`, stdout would stay piped for the rest of the test process. Using `t.Cleanup(func() { os.Stdout = orig })` is more robust. Low risk since `verifyDotenvToken` is simple, but it's the pattern the rest of the test file already follows.

**N2: `verifyDotenvToken` writes to `fmt.Println` (global stdout) instead of accepting an `io.Writer`.** This is pre-existing — the PR inherits it — but it's why the test has to redirect `os.Stdout` with a pipe instead of passing a buffer. The sibling function `checkCLICredentials` already takes an `io.Writer`, so the pattern exists in the same file.

</details>

---

### Verification notes

- **Scratch test:** confirmed the `.env` 503 behavior, then removed the scratch file.

---

### How this fits with the rest

For context, the auth flow has three credential paths that call `VerifyToken`:

```
DATAROBOT_ENDPOINT + DATAROBOT_API_TOKEN (env vars)
    └─ VerifyEnvCredentials → ReportEnvCredentialsError  ← FIXED in this PR

.env file (dr auth check in a repo)
    └─ verifyDotenvToken                                   ← gap (S1)

stored profile (drconfig.yaml, every command)
    └─ EnsureAuthenticated → GetAPIKey                     ← gap (S1)
```

This PR fixes the env-var path. The other two paths still treat any non-200 as a token problem. The env-var path is the one where misattribution is most dangerous (the user explicitly chose an instance), so it's the right one to fix first — but the other two are worth a follow-up.

ajalon1:
go ahead and ignore the govulncheck, that's a new version of Go available

internal/auth/auth.go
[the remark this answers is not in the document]
chasdr: agreed but pre-existing: the timeout and could-not-connect lines have printed the same userinfo since #751. taking all of them together in the follow-up PR with the .env/stored-profile leg split

chasdr:
bugbot run

PR #775, opened 2026-08-14
[RAPTOR-18998] feat(workload): pin to an Enclave on create, list by Enclave

alexromanyuk:
# RATIONALE

CLI half of the Enclave placement work ([RAPTOR-18998](https://datarobot.atlassian.net/browse/RAPTOR-18998)): let a developer pin a workload to a specific Enclave at create time, and find the workloads running on a given Enclave.

The Workload API expresses the pin in the create request as `runtime.enclaveSelectionPolicy: "manual"` plus `runtime.enclaves: ["<name>"]` (a list, one entry accepted today), and its list endpoint filters by resolved placement via `?enclave=<name>`. Pinning through the spec file already works with no CLI change, since the spec is sent verbatim; the flag exists so one committed spec can be deployed to different Enclaves without editing the file:

```shell
dr workload create --spec-file spec.json --enclave hadr-workload-01-rbac --output-format json
  {
    "id": "6a843ba51b848e217b9471ae",
    "name": "cli-enclave-test-pinned-140140",
    "status": "submitted",
    "type": "service",
    "importance": "low",
    "artifactId": "6a843ba51b848e217b9471ad",
    "endpoint": "https://internal-url.datarobot.com/workloads/6a843ba51b848e217b9471ae/",
    "createdAt": "2026-08-18T11:01:57Z",
    "updatedAt": "2026-08-18T11:01:57Z"
  }


dr workload list --enclave prod-east

╭──────────────────────────┬────────────────────────────────┬───────────┬─────────┬────────────┬──────────────────────╮
│ WORKLOAD ID              │ NAME                           │ STATUS    │ TYPE    │ IMPORTANCE │ UPDATED              │
├──────────────────────────┼────────────────────────────────┼───────────┼─────────┼────────────┼──────────────────────┤
│ 6a843ba51b848e217b9471ae │ cli-enclave-test-pinned-140140 │ launching │ service │ low        │ 2026-08-18 11:01 UTC │
│ 6a83cb911b848e217b947030 │ hello-whoami-charlie           │ running   │ service │ low        │ 2026-08-18 03:20 UTC │
│ 6a8309c70071967d89243203 │ okta-a2a-finance               │ stopped   │ agent   │ low        │ 2026-08-17 13:16 UTC │
│ ...                      │ ...                            │ ...       │ ...     │ ...        │ ...                  │
╰──────────────────────────┴────────────────────────────────┴───────────┴─────────┴────────────┴──────────────────────╯
```
Edge cases:
```
$ dr workload create --spec-file spec.json --enclave ""
Error: invalid --enclave: the Enclave name must be non-blank

$ dr workload create --spec-file spec-that-already-pins.json --enclave prod-east
Error: spec already sets runtime.enclaveSelectionPolicy; remove it from the spec or drop --enclave

$ dr workload list --enclave no-such-enclave-xyz
No workloads found.
```

## CHANGES

- `dr workload create --enclave <name>`: injects `runtime.enclaveSelectionPolicy: "manual"` and `runtime.enclaves: ["<name>"]` into the spec before POSTing (`internal/workload/enclave.go`, `ApplyEnclavePin`).
  - Refuses to override a spec that already sets either field: the spec file stays the source of truth, the flag never silently rewrites it.
  - An explicit blank `--enclave` is an error, not an unpinned create.
  - The re-encode preserves numbers via `json.Number` (e.g. byte-count memory values survive verbatim); key order is not preserved, and the help text documents that the byte-for-byte passthrough promise does not hold with the flag.
- `dr workload list --enclave <name>`: maps to the `?enclave=` query param, filtering by the name of the Enclave the workload actually runs on (resolved placement, not the requested pin). A blank value errors instead of silently listing everything.
- `ListWorkloads` gained the `enclave` parameter; the two other call sites (`up` conflict search, wizard picker) pass an empty filter.

Both flags ride the existing `workload` feature gate; no new gate.

## TESTING

- `task lint` clean on all three GOOS legs; `task test` passes.
- New tests: `ApplyEnclavePin` (runtime creation, field and numeric preservation, trim, blank/conflict/malformed/`runtime: null` handling), list query-param assertions (filter present when set, absent when not), command-level blank-flag and spec-conflict tests.

Verified end to end on 2026-08-14 and re-verified after the rebase on 2026-08-18 against a live dev environment with two registered Enclaves, running a Workload API build that includes the placement read-back, the `?enclave=` list filter, and the pluralized `runtime.enclaves[]` request schema:

- The pin caused the placement: the same spec (same resource bundle) created twice landed on the default Enclave without the flag and on the named Enclave with `--enclave <name>`; the stored runtime read back `enclaveSelectionPolicy: manual` and `enclaves: ["<name>"]`, and the workload's `placements[]` named the pinned Enclave.
- `dr workload list --enclave <name>` returned exactly the workloads whose resolved placement is that Enclave. The filter matches the recorded placement, which survives a stop; workloads that predate placement recording never match. An unknown name returns "No workloads found." with exit 0, and `--output-format json` emits a single valid JSON document with logs on stderr.
- Both guards fired before any network call: a blank `--enclave`, and `--enclave` combined with a spec that already sets either pin field.

### DIFF
--- a/cmd/workload/create/cmd.go
+++ b/cmd/workload/create/cmd.go
@@ -55,8 +55,8 @@ named Enclave, which must be eligible and grant you deploy access. The
 flag refuses to override a spec that already sets either field, and using
 it means the spec is re-encoded rather than sent byte-for-byte. Without
 the flag (or with enclaveSelectionPolicy "availability"), DataRobot picks
-the placement. Where the workload actually runs is reported by the
-read-only placements field on the workload document.
+the placement. Confirm where the workload landed with
+'dr workload list --enclave <name>'.
 
 Three flows:
 
--- a/cmd/workload/create/cmd.go
+++ b/cmd/workload/create/cmd.go
@@ -25,7 +25,10 @@ import (
 func Cmd() *cobra.Command {
 	var outputFormat outputformat.OutputFormat
 
-	var specFile string
+	var (
+		specFile string
+		enclave  string
+	)
 
 	cmd := &cobra.Command{
 		Use:   "create",
@@ -45,6 +48,15 @@ using standard YAML typing rules before sending: quote values that must
 stay strings (for example "0644" or "1.10"), and unquoted dates are sent
 as RFC3339 timestamps. The server validates field-level shape and returns
 a 422 with a JSON-path detail on a mismatch.
+
+Use --enclave <name> to pin the workload to a specific Enclave: it sets
+runtime.enclaveSelectionPolicy to "manual" and runtime.enclaves to the
+named Enclave, which must be eligible and grant you deploy access. The
+flag refuses to override a spec that already sets either field, and using
+it means the spec is re-encoded rather than sent byte-for-byte. Without
+the flag (or with enclaveSelectionPolicy "availability"), DataRobot picks
+the placement. Where the workload actually runs is reported by the
+read-only placements field on the workload document.
 
 Three flows:
 
@@ -117,6 +129,7 @@ Three flows:
 Example:
   dr workload create --spec-file workload.json
   dr workload create --spec-file workload.yaml
+  dr workload create --spec-file workload.json --enclave prod-east
   dr workload create --spec-file workload.yaml --output-format json`,
 		Args:         cobra.NoArgs,
 		PreRunE:      auth.EnsureAuthenticatedE,
@@ -127,6 +140,15 @@ Example:
 			payload, err := workload.ReadSpecFile(specFile)
 			if err != nil {
 				return err
+			}
+
+			// Changed, not enclave != "": an explicit --enclave "" must be
+			// rejected by ApplyEnclavePin, not silently create unpinned.
+			if cmd.Flags().Changed("enclave") {
+				payload, err = workload.ApplyEnclavePin(payload, enclave)
+				if err != nil {
+					return err
+				}
 			}
 
 			if err := workload.ValidateWorkloadCreateRequest(payload); err != nil {
@@ -147,9 +169,13 @@ Example:
 	cmd.Flags().StringVar(&specFile, "spec-file", "", "Path to JSON or YAML spec file (required)")
 	_ = cmd.MarkFlagRequired("spec-file")
 
+	cmd.Flags().StringVar(&enclave, "enclave", "",
+		"Pin the workload to the named Enclave (sets runtime.enclaveSelectionPolicy=manual)")
+
 	telemetry.TrackWith(cmd, func(_ *cobra.Command, _ []string) map[string]any {
 		return map[string]any{
 			"output_format": string(outputFormat),
+			"enclave":       enclave,
 		}
 	})
 
--- a/cmd/workload/list/cmd.go
+++ b/cmd/workload/list/cmd.go
@@ -15,6 +15,7 @@
 package list
 
 import (
+	"errors"
 	"fmt"
 	"strings"
 
@@ -31,6 +32,7 @@ func Cmd() *cobra.Command {
 	var (
 		limit    int
 		statuses []string
+		enclave  string
 	)
 
 	cmd := &cobra.Command{
@@ -45,11 +47,17 @@ For each workload the listing shows:
 
 By default, output is a human-readable table. Use --output-format json for machine-parseable output.
 
+Use --enclave to only list workloads running on the named Enclave. The
+filter matches where each workload actually runs (its resolved placement),
+not the pin requested at create time; an unknown name simply matches
+nothing.
+
 Example:
   dr workload list
   dr workload list --limit 10
   dr workload list --status running
   dr workload list --status errored --status interrupted
+  dr workload list --enclave prod-east
   dr workload list --output-format json`,
 		Args:         cobra.NoArgs,
 		PreRunE:      auth.EnsureAuthenticatedE,
@@ -61,12 +69,18 @@ Example:
 				return fmt.Errorf("invalid --limit %d: must be positive", limit)
 			}
 
+			// A blank --enclave would silently drop the filter and list
+			// everything, the opposite of what the flag asked for.
+			if cmd.Flags().Changed("enclave") && strings.TrimSpace(enclave) == "" {
+				return errors.New("invalid --enclave: the Enclave name must be non-blank")
+			}
+
 			parsedStatuses, err := workload.ParseWorkloadStatuses(statuses)
 			if err != nil {
 				return err
 			}
 
-			workloads, err := workload.ListWorkloads(limit, parsedStatuses)
+			workloads, err := workload.ListWorkloads(limit, parsedStatuses, enclave)
 			if err != nil {
 				return err
 			}
@@ -80,6 +94,8 @@ Example:
 	cmd.Flags().IntVar(&limit, "limit", 100, "Maximum number of workloads to return")
 	cmd.Flags().StringSliceVar(&statuses, "status", nil,
 		"Filter by status (repeatable, also accepts comma-separated values; e.g. running, errored)")
+	cmd.Flags().StringVar(&enclave, "enclave", "",
+		"Only list workloads running on the named Enclave")
 
 	telemetry.TrackWith(cmd, func(c *cobra.Command, _ []string) map[string]any {
 		limit, _ := c.Flags().GetInt("limit")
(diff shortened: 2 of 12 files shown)


internal/workload/enclave.go
```diff
@@ -0,0 +1,89 @@
...
+	}
+
+	dec := json.NewDecoder(bytes.NewReader(spec))
+	dec.UseNumber()
+
+	var doc map[string]any
+
+	if err := dec.Decode(&doc); err != nil {
+		return nil, fmt.Errorf("invalid spec: %w", err)
+	}
```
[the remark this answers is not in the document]
alexromanyuk: Fixed in ed1cbec: a top-level null now returns "invalid spec: the spec must be a JSON object" instead of panicking, with a regression test.

cmd/workload/create/cmd.go
```diff
@@ -129,6 +142,15 @@ Example:
 				return err
 			}
 
+			// Changed, not enclave != "": an explicit --enclave "" must be
+			// rejected by ApplyEnclavePin, not silently create unpinned.
+			if cmd.Flags().Changed("enclave") {
+				payload, err = workload.ApplyEnclavePin(payload, enclave)
```
wojtekwdr, line 148: [medium] Nothing asserts that the pinned bytes are what gets POSTed. I swapped this line for `_, err = workload.ApplyEnclavePin(payload, enclave)` and every package still passed, while the binary created an unpinned workload and exited 0. `list` has the equivalent check in `TestListWorkloads_EnclaveFilter`, and `cmd/llm-gateway/list/cmd_test.go` has the `viperx.Set(config.DataRobotURL, ...)` seam to copy.

Related: `ApplyEnclavePin` returns `[]byte` where `ReadSpecFile` returns `json.RawMessage`, so this works only because `payload` keeps its static type across the `=`. Tidy it into `pinned, err := ...` and the body goes out base64-encoded, with `CreateWorkload` taking `any` so nothing complains.

internal/workload/workload.go
```diff
@@ -418,6 +420,10 @@ func ListWorkloads(limit int, statuses []string) ([]Workload, error) {
 		query.Add("status", s)
 	}
 
+	if enclave != "" {
```
wojtekwdr, line 426: [medium] question: what does a Workload API that predates the `?enclave=` filter do with this param? If it's the usual FastAPI behavior of ignoring undeclared query params, then on an instance that hasn't picked up workload-api#1132 `dr workload list --enclave prod-east` returns every workload with exit 0 and looks just like a successful filter. Create degrades loudly on the same server (an unknown `runtime.enclaves` gets a 422), so list is the asymmetric one. You verified against a build that has the filter, so what I'm really asking about is self-managed instances that lag.

cmd/workload/list/cmd.go
```diff
@@ -61,12 +69,18 @@ Example:
...
+				return errors.New("invalid --enclave: the Enclave name must be non-blank")
+			}
+
 			parsedStatuses, err := workload.ParseWorkloadStatuses(statuses)
 			if err != nil {
 				return err
 			}
 
-			workloads, err := workload.ListWorkloads(limit, parsedStatuses)
+			workloads, err := workload.ListWorkloads(limit, parsedStatuses, enclave)
```
wojtekwdr, line 83: [low] The guard above trims to decide and then the untrimmed value goes out: `--enclave "  prod-east  "` becomes `?enclave=++prod-east++`, prints "No workloads found." and exits 0, while the same argument on create pins correctly because `ApplyEnclavePin` trims. `ParseWorkloadStatuses` normalizes the sibling filter on this command, so trimming inside `ListWorkloads` next to the `limit` check would cover future callers too.

internal/workload/enclave_test.go
```diff
@@ -0,0 +1,127 @@
...
+	require.NoError(t, json.Unmarshal(pinned, &doc))
+
+	runtime := doc["runtime"].(map[string]any)
+
+	assert.Equal(t, EnclaveSelectionPolicyManual, runtime["enclaveSelectionPolicy"])
+	assert.NotNil(t, runtime["containerGroups"], "existing runtime content must survive the pin")
+
+	// The large memory integer must survive re-encoding verbatim rather than
+	// decaying into scientific notation via float64.
+	assert.Contains(t, string(pinned), "4294967296")
```
wojtekwdr: [low] This one can't fail. 4294967296 is 2^32, exact in a float64, and Go prints it straight back as `4294967296`, so the assertion still passes with `dec.UseNumber()` deleted (I tried it, the suite stays green). `9007199254740993` or `1e2` would catch the decay the comment above describes.

cmd/workload/create/cmd.go
```diff
@@ -46,6 +49,15 @@ stay strings (for example "0644" or "1.10"), and unquoted dates are sent
...
 a 422 with a JSON-path detail on a mismatch.
 
+Use --enclave <name> to pin the workload to a specific Enclave: it sets
+runtime.enclaveSelectionPolicy to "manual" and runtime.enclaves to the
+named Enclave, which must be eligible and grant you deploy access. The
+flag refuses to override a spec that already sets either field, and using
+it means the spec is re-encoded rather than sent byte-for-byte. Without
+the flag (or with enclaveSelectionPolicy "availability"), DataRobot picks
+the placement. Where the workload actually runs is reported by the
+read-only placements field on the workload document.
```
wojtekwdr: [low] question: could this point at `dr workload list --enclave <name>` instead? `placements` is real in the API, but nothing in the CLI prints it: `Workload` doesn't parse it and `WorkloadOutput` doesn't emit it, so `dr workload get --output-format json` comes back with the same nine keys whether the pin took or not. The filter this PR adds is the only CLI-side way to answer "did it land where I asked".

wojtekwdr (approved):

cmd/workload/list/cmd.go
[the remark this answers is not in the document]
alexromanyuk: Fixed in 06ec55e. Trim lives in ListWorkloads next to the limit check, not in the command - Wojciech asked for the same thing one comment over, and doing it there covers future callers too. Test added.

cmd/workload/create/cmd.go
alexromanyuk: Both fixed in 06ec55e. TestCmd_EnclavePinReachesTheWire runs the command against an httptest server and asserts the POSTed runtime carries the pin - I re-ran your mutation (swap to `_, err =`) and it fails now. And ApplyEnclavePin returns json.RawMessage, so the base64 trap you found is gone at the type level instead of surviving on luck. Good catch on both, the second one would've been a miserable production debug.

internal/workload/workload.go
alexromanyuk: It 422s, loudly. CommonFilterQueryParams has extra="forbid" since the March redesign (#335), and the route takes Annotated[WorkloadQueryParams, Query()], so FastAPI rejects the unknown param with the field named in the detail. Any server old enough to silently ignore it predates dr workload entirely. So list and create degrade the same way: a lagging self-managed instance fails with a 422 that names the problem, not a fake-successful full listing. No change needed - but worth a smoke case once we have a version matrix.

cmd/workload/list/cmd.go
alexromanyuk: Fixed in 06ec55e, and I took your placement: trim inside ListWorkloads next to the limit check, so the command layer only decides blank-vs-not and every future caller gets the normalization for free. TestListWorkloads_TrimsEnclave pins it.

internal/workload/enclave_test.go
alexromanyuk: You're right, 2^32 tested nothing. Swapped to 9007199254740993 (2^53+1) in 06ec55e and verified it the same way you did: delete UseNumber and the test fails with ...992. Thanks for actually running the mutation instead of trusting the comment.

cmd/workload/create/cmd.go
alexromanyuk: Agreed and fixed in 06ec55e - the help now says to confirm with 'dr workload list --enclave <name>', which is the only CLI-side answer today. Printing placements[] is the placement verb's job (RAPTOR-18992 CLI half), and I'd rather not smuggle it into this PR.

internal/workload/workload.go
alexromanyuk: Follow-up: now verified empirically, not just from the code. Sent an unknown query param to a live instance's /workloads/ and got the 422 with the param named: {"path":"bogusFilter","message":"Extra inputs are not permitted","code":"extra_forbidden"}. An old server sees ?enclave= as exactly that kind of unknown param, and the forbid config predates the filter by five months.

PR #780, opened 2026-08-17
[RAPTOR-18959] feat(workload): remove the workload feature gate for WAPI GA

wojtekwdr:
# RATIONALE

`dr workload` and `dr artifact` have been hidden behind `DATAROBOT_CLI_FEATURE_WORKLOAD` since v0.2.63. Workload API is GA, so the gate is now pure friction: every user and every doc has to export the variable first, and until they do neither tree appears in `dr --help` or in shell completions.

Both trees share the single gate name `workload`, so they can only be ungated together. `pipeline` stays gated and becomes the only live gate, so `internal/features` and `cli.CommandAdder` stay put.

## CHANGES

Dropped the two `features.SetGate` calls and everything that referenced the variable.

`docs/development/feature-gates.md` needed more than a find-and-replace. It used `workload` as its worked example throughout, and it taught a raw `Annotations` literal that no call site has ever used, so it is now written around `pipeline` and `features.SetGate`. Its "Removing a Feature Gate" procedure was also wrong: it said to delete the `Annotations` map and stop, which does not compile, because it leaves an unused import behind.

`docs/commands/workload.md` and `artifact.md` join the mkdocs nav, where neither had ever appeared. Adding `workload.md` turned up that it documented 9 of the 11 subcommands, missing `config` and `up`, so those are filled in rather than published incomplete.

## Breaking change: legacy `.wapi/` state is no longer read

GA is also the cutoff for the pre-GA state layout, so `EnsureMigrated` and the `Dir()` fallback are gone. **A project still holding a root `.wapi/` now reads as unlinked.**

That directory shipped from v0.2.63 through v0.2.81. The auto-migration that would have relocated it existed only in v0.2.82 and v0.2.83, so anyone who skipped those two releases gets no migration at all. The affected population is gate-opting early adopters by definition, since exporting the variable was the only way to create the directory in the first place.

One path degraded badly without a guard. Every command reports a visible "not linked" error except `dr workload up`, which read the project as a first deploy, minted a **second artifact server-side** and orphaned the one the project had been pushing to, along with its builds and locked versions, silently. So `wapi.LegacyState` now recognises the old layout and every affected command explains itself:

```
Error: not linked: .wapi/ holds state written by an older CLI, which is no longer read.
Run 'dr artifact code init art-abc-123' to link this project again, then delete .wapi/
```

`up` refuses before it touches the server. `dr artifact code init` still succeeds, because re-linking is the recovery, but it now prints a note that the old directory is orphaned. Naming the artifact id is the point: `.wapi/` carries a `.gitignore` of `*`, so nothing else would ever surface it, and that id is the only way back.

A project that was genuinely never linked still gets the plain hint, with no mention of a directory it does not have.

### Residual cost, not fixed

A `.wapi/.rollback` tree left by a sync that crashed on v0.2.81 or earlier is no longer swept, and those backups are the only copy of the files that sync overwrote. Keeping that working would mean keeping half the legacy machinery alive. Flagging it here rather than burying it; the guard at least makes the surrounding situation visible.

## NOTES

Reviewers running the suite locally: `Taskfile.yaml` loads `.env`, and a typical dev `.env` sets `DATAROBOT_CLI_FEATURE_WORKLOAD=true`, so `task test` exercises the gate-on path. For the clean signal use `DATAROBOT_CLI_FEATURE_WORKLOAD= go test ./...`.

## PR Automation

> [!IMPORTANT]
>

---

> [!NOTE]
> **Medium Risk**
> Always-on exposure of workload and artifact commands changes the default CLI surface for all users (help, completions, and discoverability). Remaining risk is mostly operational—`pipeline` is still gated and tests/env quirks around `DATAROBOT_CLI_FEATURE_WORKLOAD` in dev `.env` are called out in the PR notes.
> 
> **Overview**
> **`dr artifact` and `dr workload` are no longer hidden behind `DATAROBOT_CLI_FEATURE_WORKLOAD`.** The `features.SetGate` calls are removed from `cmd/artifact/cmd.go` and `cmd/workload/cmd.go`, so both trees always appear in `dr --help`, completions, and the live `RootCmd` tree. `pipeline` remains the only gated command.
> 
> Tests are flipped to match: `TestWorkloadAndArtifactCommandsAlwaysRegistered` asserts both commands are present by default, and `TestGatedCommandFilteredFromRootCmd` keeps coverage that `pipeline` stays out of `RootCmd` when its gate is off. Telemetry wiring tests no longer build separate gated subtrees—artifact and workload paths are checked on `RootCmd` via an expanded `expectedTrackedCommands` list.
> 
> Documentation drops the env-var requirement for artifact/workload, adds **`config` and `up`** to the workload reference and command tree, publishes **artifact** and **workload** in `docs/mkdocs.yml`, and rewrites **feature-gates** docs to use `pipeline` and `features.SetGate` (including a corrected GA checklist). `AGENTS.md` and comment examples now cite `DATAROBOT_CLI_FEATURE_PIPELINE` instead of `WORKLOAD`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 2fc85bc70de9c663404d63b81fdd551442bfc46f. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -222,7 +222,7 @@ Feature gates allow commands to be hidden until ready for release. For comprehen
 
 **Quick reference:**
 - Gate a command via `features.SetGate(cmd, "feature-name")` (sets the annotation on the command)
-- Enable via env var: `DATAROBOT_CLI_FEATURE_<NAME>=true` (e.g., `DATAROBOT_CLI_FEATURE_WORKLOAD=true`)
+- Enable via env var: `DATAROBOT_CLI_FEATURE_<NAME>=true` (e.g., `DATAROBOT_CLI_FEATURE_PIPELINE=true`)
 - Currently supported: environment variables only (config file support planned)
 - Filtering happens via `cli.CommandAdder.AddCommand` at registration time — `CommandAdder` is the only filtering mechanism
 - To gate a **nested** subcommand, wrap the parent with `&cli.CommandAdder{Command: parent}` and call `adder.AddCommand(...)` instead of `parent.AddCommand(...)`
--- a/cmd/artifact/cmd.go
+++ b/cmd/artifact/cmd.go
@@ -22,7 +22,6 @@ import (
 	"github.com/datarobot/cli/cmd/artifact/get"
 	"github.com/datarobot/cli/cmd/artifact/list"
 	"github.com/datarobot/cli/cmd/artifact/lock"
-	"github.com/datarobot/cli/internal/features"
 	"github.com/spf13/cobra"
 )
 
@@ -36,8 +35,6 @@ func Cmd() *cobra.Command {
 Create and inspect artifacts, build their container images, and
 synchronize local code with a remote artifact.`,
 	}
-
-	features.SetGate(cmd, "workload")
 
 	cmd.AddCommand(
 		// The artifact is the primary resource: direct verbs, like
--- a/cmd/root_test.go
+++ b/cmd/root_test.go
@@ -18,6 +18,7 @@ import (
 	"bytes"
 	"testing"
 
+	"github.com/datarobot/cli/internal/features"
 	"github.com/datarobot/cli/internal/misc/reader"
 	"github.com/datarobot/cli/internal/telemetry"
 	"github.com/datarobot/cli/internal/tools"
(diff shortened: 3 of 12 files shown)


chasdr (approved):
lgtm on the code, gate removal is clean and the docs/tests all line up.

one thing on the description though: the whole "legacy `.wapi/` state is no longer read" section is #781's work, not this PR. #780 just removes the feature gate. reading this one you'd go looking for `wapi.LegacyState` and not find it.

mind trimming the body down to the gate removal and letting #781 carry the breaking-change writeup? otherwise anyone approving this thinks they reviewed the .wapi change when it's actually the next PR up.

adamalpi (approved):
Mechanical, clean ungate — verified the suites pass gate-off on the branch (`DATAROBOT_CLI_FEATURE_WORKLOAD= go test ./...`), the new `TestGatedCommandFilteredFromRootCmd` preserves CommandAdder filtering coverage via `pipeline`, no CI workflow sets the env var so it degrades to a no-op, and the rewritten feature-gates doc matches the actual `features.SetGate` implementation. One note for the record, non-blocking: the "Breaking change: legacy .wapi/ state" section of the description belongs to the closed, unmerged #781 — this diff leaves `EnsureMigrated` and the `.wapi/` fallback fully intact, so no breaking change ships here.

PR #784, opened 2026-08-18
[RAPTOR-19349] add offset to workload/artifact list

ScottOglesby:
# RATIONALE
Add `--offset` to the list commands for artifacts and workloads. `--limit` already exists, but not having `--offset` made listing items beyond the first page impossible.

## CHANGES
* Added `--offset` to the list commands for artifacts and workloads
* added and updated tests

## Testing
Ran `task lint`, `task test`

## PR Automation

> [!IMPORTANT]
>

---

> [!NOTE]
> **Low Risk**
> Backward-compatible CLI pagination wired to existing list APIs; default offset preserves prior behavior and changes are limited to list/query plumbing.
> 
> **Overview**
> Adds **`--offset`** (default `0`) to **`dr workload list`** and **`dr artifact list`** so users can skip past the first page of results alongside existing **`--limit`**.
> 
> The list commands validate that offset is non-negative, pass offset into **`ListWorkloads`** and **`ListArtifacts`**, and include it in telemetry. **`ListWorkloads`** also rejects negative offset before building the API query; artifact listing appends **`offset`** to the artifacts API URL. Internal callers that only need the first page (deploy name-conflict lookup, workload config wizard picker) now pass **`0`** for offset. Tests cover invalid **`--offset`** on both commands and updated **`ListWorkloads`** call signatures.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 17992391b3404b105df501c36d3092944cb7f4f8. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


cmd/artifact/list/cmd.go
```diff
@@ -70,12 +77,15 @@ Example:
 
 	workload.AddStatusFlag(cmd, &status)
 	cmd.Flags().IntVar(&limit, "limit", 100, "Maximum number of artifacts to return")
+	cmd.Flags().IntVar(&offset, "offset", 0, "Number of artifacts to skip before returning results")
```
ajalon1, line 80: Probably not your intention, but I'm happy to see that this is backwards-compatible. Not a breaking change. 😆

internal/workload/workload_test.go
```diff
@@ -410,7 +410,7 @@ func TestListWorkloads_FollowsNextAndTruncatesToLimit(t *testing.T) {
 
 func TestListWorkloads_RejectsNonPositiveLimit(t *testing.T) {
```
ajalon1, line 411: Add a TestListWorkloads_RejectsNegativeOffset() test

internal/workload/artifact.go
```diff
@@ -491,8 +491,8 @@ func DeleteArtifact(artifactID string) error {
 	return drapi.DeleteJSON(url, "artifact", nil, nil)
 }
 
-func ListArtifacts(limit int, status Status) ([]Artifact, error) {
-	endpoint := "/api/v2/artifacts/?limit=" + strconv.Itoa(limit)
+func ListArtifacts(limit, offset int, status Status) ([]Artifact, error) {
+	endpoint := "/api/v2/artifacts/?limit=" + strconv.Itoa(limit) + "&offset=" + strconv.Itoa(offset)
```
ajalon1, line 495: Could consider switching this over to use `url.Values{}` like `ListWorkloads()` does as that would be a better pattern than building this by hand.

Absolutely not your problem. If you do tackle it I'd suggest doing this in a followup PR.

Something closer to this:

```go
query := url.Values{}
query.Set("limit", strconv.Itoa(limit))
query.Set("offset", strconv.Itoa(offset))
if status != "" {
    query.Set("status", string(status))
}
pageURL, err := config.GetEndpointURL("/api/v2/artifacts/")
```

internal/workload/workload.go
```diff
@@ -406,13 +406,18 @@ type WorkloadList struct {
...
-func ListWorkloads(limit int, statuses []string) ([]Workload, error) {
+func ListWorkloads(limit, offset int, statuses []string) ([]Workload, error) {
 	if limit <= 0 {
 		return nil, fmt.Errorf("invalid limit %d: must be positive", limit)
 	}
 
+	if offset < 0 {
+		return nil, fmt.Errorf("invalid offset %d: must be non-negative", offset)
+	}
+
```
ajalon1, line 417: I think ListArtifacts() in the internal module should validate `limit` and `offset` the way ListWorkloads() does.

cmd/artifact/list/cmd.go
```diff
@@ -57,7 +60,11 @@ Example:
 				return fmt.Errorf("invalid --limit %d: must be positive", limit)
 			}
 
-			artifacts, err := workload.ListArtifacts(limit, status)
+			if offset < 0 {
+				return fmt.Errorf("invalid --offset %d: must be non-negative", offset)
+			}
```
ajalon1, line 65: Another suggestion for a followup PR would be refactoring this to a custom pflag.Value. We do this already in cmd/internal/pollflags/pollflags.go.

Absolutely Not blocking, since the current if-return-error is widespread throughout the repo, but would be a nice update. Would also have the benefit of moving the validation from runtime to parsetime.

ajalon1 (commented):
Hey @ScottOglesby 👋🏽
Pretty good. I've got a lot of small nits and observations here, that are really just stretch goals, that are coming up as we're maturing this codebase. I just like making general suggestions.

I would approve after you update ListArtifacts in the internal API module to validate limit and offset. Anything else is up to the the Workload team.

ajalon1:
@ScottOglesby I'm helping Workload team right now on CLI stuff, so if you're OK with it I'm going to wrap up this branch and push to it.

cmd/artifact/list/cmd.go
ajalon1: Thanks — yeah, it fell out of keeping the existing two-arg call sites working.

internal/workload/workload_test.go
ajalon1: Added TestListWorkloads_RejectsNegativeOffset alongside the existing limit test.

internal/workload/artifact.go
ajalon1: Done in this PR — ListArtifacts now builds its query with url.Values via drapi.EndpointURL.

internal/workload/workload.go
ajalon1: Done — ListArtifacts validates limit and offset and clamps the page size to 100, mirroring ListWorkloads.

cmd/artifact/list/cmd.go
ajalon1: Deferring the custom pflag.Value refactor to a followup PR as suggested.

ajalon1 (approved):

PR #788, opened 2026-08-18
[CFX-4450] dr self update --version

taras-pokornyy:
# RATIONALE

`dr self update` could only ever install the *latest* release — there was no way to pin a specific version. This blocks reproducible environments, rollback-for-testing, and matching a specific `dr` version to a teammate's or CI's. This change adds a `--version` flag with strict validation, refuses downgrades, and hard-errors under Homebrew (which can't pin versions at all).

While implementing this, the existing Homebrew-cask detection was found to be macOS-only (`if runtime.GOOS == "darwin"`), even though Homebrew ("Linuxbrew") is a documented, supported install method on Linux too (`README.md`'s own "Install via Homebrew / Linuxbrew" section). A Linux user on a brew-managed install would previously fall straight through to the raw `curl | sh` path, bypassing brew entirely and getting no protection from the new `--version`-under-Homebrew guard. This is fixed alongside the main feature.

## CHANGES
- Added `--version <version>` flag to `dr self update` (no shorthand, not settable via any env var).
  - `normalizeAndValidateVersion`: strict `vMAJOR.MINOR.PATCH` validation (optional leading `v`, optional `-prerelease`/`+build` suffix), normalized to a `v`-prefixed form. Uses a regex rather than the semver library's lenient parser, since the ticket's own example (`"1.2"` → error) would otherwise silently pass.
  - `refuseDowngrade`: uses `github.com/Masterminds/semver/v3` to block installing a version older than the one currently running; skipped entirely for non-release (`"dev"`) builds.
  - The existing "already sufficient, skip update" short-circuit no longer swallows an explicit `--version` request.
- Extracted the Homebrew-cask update logic into `tryBrewUpdate`, and extended it to run on **both** `darwin` and `linux` (previously darwin-only), consolidated into the single `switch runtime.GOOS { case "darwin", "linux": ... }` dispatch instead of a separate standalone `if` block. Shell detection now only runs when actually needed (previously called unconditionally, even for the brew-handled path and for Windows where it's unused).
  - `brewCaskVersionError`: hard error returned when `--version` is passed but `dr` is installed via the Homebrew cask, with the manual `install.sh` command to run instead.
- Threaded `targetVersion` into the non-brew install paths: appended to the `curl | sh` invocation (`sh -s -- <version>`) and set via `$env:VERSION='<version>';` before the PowerShell one-liner on Windows — both install scripts already supported pinned versions, this just wires the flag through.

## PR Automation

> [!IMPORTANT]
>

---

> [!NOTE]
> **Medium Risk**
> Changes how the CLI replaces itself (brew reinstall and remote install scripts) on user machines; behavior is well-tested but mistakes could break updates or pin the wrong release.
> 
> **Overview**
> Adds **`--version`** to **`dr self update`** so users can install a specific release (`vX.Y.Z` or `X.Y.Z`) instead of always taking latest. Versions are validated with a strict MAJOR.MINOR.PATCH regex, normalized to a `v` prefix, and **downgrades are refused** against the running binary (skipped for `dev` builds). An explicit `--version` request is no longer blocked by the “already sufficient, skip update” shortcut.
> 
> Pinned installs are wired into the existing install scripts on **Windows** (`$env:VERSION`) and **Unix shell** (`install.sh -s -- <version>`). **Homebrew cask** installs cannot pin versions; `--version` now returns a guided error with the manual `install.sh` command.
> 
> Homebrew self-update logic is refactored into **`tryBrewUpdate`** and extended from **macOS-only to macOS and Linux**, so Linuxbrew-managed `dr-cli` cask installs use `brew update` / `reinstall` instead of falling through to `curl | sh`. Shell detection runs only when the generic install path is needed.
> 
> User docs (README, `self`, quick reference) and unit tests for validation, downgrade refusal, and the Homebrew pin error are updated. CI pins Taskfile to **3.52.0**.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 2b841ea44b00d84d6c72a52b9be61846584963ec. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/cmd/self/update/cmd.go
+++ b/cmd/self/update/cmd.go
@@ -89,7 +89,7 @@ with your default shell.
 
 			switch runtime.GOOS {
 			case "windows":
-				command = "irm https://raw.githubusercontent.com/datarobot-oss/cli/main/install.ps1 | iex"
+				command = "irm https://cli.datarobot.com/winstall | iex"
 				if targetVersion != "" {
 					command = fmt.Sprintf("$env:VERSION='%s'; %s", targetVersion, command)
 				}
@@ -117,7 +117,7 @@ with your default shell.
 					return fmt.Errorf("detect shell: %w", err)
 				}
 
-				command = "curl -fsSL https://raw.githubusercontent.com/datarobot-oss/cli/main/install.sh | sh"
+				command = "curl -fsSL https://cli.datarobot.com/install | sh"
 				if targetVersion != "" {
 					command += " -s -- " + targetVersion
 				}
@@ -215,8 +215,9 @@ func isBrewCaskInstalled(brewPath string) bool {
 func brewCaskVersionError(targetVersion string) error {
 	return fmt.Errorf(
 		"dr was installed via Homebrew (dr-cli cask). Homebrew always installs the latest release and cannot pin versions.\n\n"+
-			"To install %s manually, uninstall the cask and run:\n\n"+
-			"  curl -fsSL https://raw.githubusercontent.com/datarobot-oss/cli/main/install.sh | sh -s -- %s\n",
+			"To install %s manually, uninstall the cask and install with the installation script:\n\n"+
+			"  brew uninstall --cask dr-cli\n"+
+			"  curl -fsSL https://cli.datarobot.com/install | sh -s -- %s\n",
 		targetVersion, targetVersion)
 }
 
--- a/cmd/self/update/cmd.go
+++ b/cmd/self/update/cmd.go
@@ -19,9 +19,11 @@ import (
 	"os"
 	"os/exec"
 	"path/filepath"
+	"regexp"
 	"runtime"
 	"strings"
 
+	"github.com/Masterminds/semver/v3"
 	"github.com/datarobot/cli/internal/fsutil"
 	"github.com/datarobot/cli/internal/log"
 	internalShell "github.com/datarobot/cli/internal/shell"
@@ -30,8 +32,15 @@ import (
 	"github.com/spf13/cobra"
 )
 
+// versionPattern requires a strict MAJOR.MINOR.PATCH numeric triplet, with an
+// optional leading "v" and optional -prerelease / +build suffixes.
+var versionPattern = regexp.MustCompile(`^v?\d+\.\d+\.\d+(-[0-9A-Za-z-.]+)?(\+[0-9A-Za-z-.]+)?$`)
+
 func Cmd() *cobra.Command { //nolint:cyclop
-	var force bool
+	var (
+		force         bool
+		targetVersion string
+	)
 
 	cmd := &cobra.Command{
 		Use:   "update",
@@ -42,12 +51,25 @@ with your default shell.
 `,
 
 		RunE: func(_ *cobra.Command, _ []string) error {
+			if targetVersion != "" {
+				normalized, err := normalizeAndValidateVersion(targetVersion)
+				if err != nil {
+					return err
+				}
+
+				targetVersion = normalized
+
+				if err := refuseDowngrade(targetVersion); err != nil {
+					return err
+				}
+			}
+
 			requirement, err := tools.GetSelfRequirement()
 			if err != nil {
 				return fmt.Errorf("get self requirement: %w", err)
 			}
 
-			if tools.SufficientSelfVersion(requirement.MinimumVersion) && !force {
+			if tools.SufficientSelfVersion(requirement.MinimumVersion) && !force && targetVersion == "" {
 				if requirement.MinimumVersion != "" {
 					fmt.Fprintf(os.Stderr, "Required version: %s. ", requirement.MinimumVersion)
 				}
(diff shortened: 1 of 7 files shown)


chasdr:
Hi Taras - 1 question before I approve:

the rationale lists rollback-for-testing as a motivation, but `refuseDowngrade` blocks every older version on a release build and `--force` doesn't bypass it.

`install.sh` itself supports downgrade (the `[y/N]` prompt, and it proceeds non-interactively when piped), so the capability's there at the script layer.

should `--force` let `--version` go backwards?

cmd/self/update/cmd.go
```diff
@@ -158,10 +158,109 @@ with your default shell.
...
+}
+
+// refuseDowngrade errors if requestedVersion is older than the currently
+// running dr version (version.Version). Skipped entirely for non-release
+// ("dev") builds.
+func refuseDowngrade(requestedVersion string) error {
+	if version.Version == "dev" {
+		return nil
+	}
+
```
ahjota: I think we should extend this to include other version bits like `rc` or `test` or whatever. I'm not sure we effectively control this right now.

cmd/self/update/cmd.go
```diff
@@ -158,10 +158,109 @@ with your default shell.
...
+	return nil
+}
+
+// isBrewCaskInstalled reports whether the dr-cli Homebrew cask is installed,
+// given a resolved brew path.
+func isBrewCaskInstalled(brewPath string) bool {
+	return exec.Command(brewPath, "list", "--cask", "dr-cli").Run() == nil
+}
+
```
ahjota, line 213: This presumes that the cask is always named dr-cli. I suppose that's always going to be true as long as this is accurate. https://github.com/datarobot-oss/homebrew-taps/tree/main/Casks

cmd/self/update/cmd.go
```diff
@@ -158,10 +158,109 @@ with your default shell.
...
+// brewCaskVersionError is returned when --version is requested but dr was
+// installed via the Homebrew cask, which cannot pin specific versions.
+func brewCaskVersionError(targetVersion string) error {
+	return fmt.Errorf(
+		"dr was installed via Homebrew (dr-cli cask). Homebrew always installs the latest release and cannot pin versions.\n\n"+
+			"To install %s manually, uninstall the cask and run:\n\n"+
```
ahjota: maybe "... manually, `brew uninstall --cask dr-cli` and run`:

cmd/self/update/cmd.go
```diff
@@ -158,10 +158,109 @@ with your default shell.
...
+			"  curl -fsSL https://raw.githubusercontent.com/datarobot-oss/cli/main/install.sh | sh -s -- %s\n",
```
ahjota: `cli.datarobot.com install`

cmd/self/update/cmd.go
```diff
@@ -158,10 +158,109 @@ with your default shell.
 	}
 
 	cmd.Flags().BoolVarP(&force, "force", "f", false, "Force update to latest version")
+	cmd.Flags().StringVar(&targetVersion, "version", "", "Install a specific released version instead of latest (e.g. v0.12.3)")
```
ahjota, line 161: Doesn't seem like this will conflict with the root command's `--version` flag. 👍🏼

docs/commands/self.md
```diff
@@ -121,17 +121,25 @@ dr self update
...
-- **Homebrew (macOS)**&mdash;uses `brew update && upgrade dr-cli` if installed via Homebrew
+- **Homebrew (macOS/Linux)**&mdash;uses `brew update && upgrade dr-cli` if installed via Homebrew
 - **Windows**&mdash;runs the PowerShell installation script
 - **macOS/Linux**&mdash;runs the shell installation script
 
 The update process will download and install the latest version while preserving your configuration and credentials.
 
+**Options:**
+
+- `--version <version>`&mdash;install a specific released version instead of the latest. Accepts `vX.Y.Z` or `X.Y.Z`, optionally with a `-prerelease` and/or `+build` suffix. Refuses to install a version older than the one currently running. Not available when `dr` was installed via the Homebrew cask&mdash;Homebrew always installs the latest release and cannot pin versions, so the command exits with an error showing the manual `install.sh` command to run instead.
```
ahjota: Nice.

docs/commands/self.md
```diff
@@ -121,17 +121,25 @@ dr self update
...
+
 **Examples:**
 
 ```bash
 # Update to latest version
 dr self update
+
+# Install a specific version
```
ahjota, line 141: Appreciate that you made either one available.

docs/index.md
```diff
@@ -49,7 +49,7 @@ Detailed documentation for each command:
 - [task](commands/task.md)&mdash;Taskfile composition and management, including task listing and execution.
 - [dotenv](commands/dotenv.md)&mdash;environment variable management with interactive wizard and validation.
 - [completion](commands/completion.md)&mdash;shell completion setup (install/uninstall or generate for bash, zsh, fish, PowerShell).
-- [self](commands/self.md)&mdash;CLI utility commands (version, update, config, completion, and plugin authoring: add, publish, package).
+- [self](commands/self.md)&mdash;CLI utility commands (version, update &mdash; including installing a specific released version, config, completion, and plugin authoring: add, publish, package).
```
ahjota, line 52: Not necessary

ahjota (dismissed):

cmd/self/update/cmd.go
taras-pokornyy: Currently `dev` is hardcoded in the source code. 

Where can I find exhaustive list ?

taras-pokornyy:
> Hi Taras - 1 question before I approve:
> 
> the rationale lists rollback-for-testing as a motivation, but `refuseDowngrade` blocks every older version on a release build and `--force` doesn't bypass it.
> 
> `install.sh` itself supports downgrade (the `[y/N]` prompt, and it proceeds non-interactively when piped), so the capability's there at the script layer.
> 
> should `--force` let `--version` go backwards?

Hey @chasdr ! 
Thank you for comment. Actually it's a very good point! 
Description of  [CFX-4450](https://datarobot.atlassian.net/browse/CFX-4450) says **refusal to downgrade** . 
I believe it would be good to allow `--force` downgrade the version of `dr-cli`

@ajalon1 , do we want to allow `--force` to downgrade `dr-cli` version ?

cmd/self/update/cmd.go
ajalon1: This seems like a good reference. https://semver.org/#spec-item-9

> A pre-release version MAY be denoted by appending a hyphen and a series of dot separated identifiers immediately following the patch version. Identifiers MUST comprise only ASCII alphanumerics and hyphens [0-9A-Za-z-]. Identifiers MUST NOT be empty. Numeric identifiers MUST NOT include leading zeroes. Pre-release versions have a lower precedence than the associated normal version. A pre-release version indicates that the version is unstable and might not satisfy the intended compatibility requirements as denoted by its associated normal version. Examples: 1.0.0-alpha, 1.0.0-alpha.1, 1.0.0-0.3.7, 1.0.0-x.7.z.92, 1.0.0-x-y-z.--.

ajalon1:
> do we want to allow --force to downgrade dr-cli version ?

@taras-pokornyy discussed in standup, but yes, if we allow `--force` to upgrade, we should also allow it to downgrade.

cmd/self/update/cmd.go
```diff
@@ -42,12 +51,25 @@ with your default shell.
...
+				normalized, err := normalizeAndValidateVersion(targetVersion)
+				if err != nil {
+					return err
+				}
+
+				targetVersion = normalized
+
+				if err := refuseDowngrade(targetVersion); err != nil {
+					return err
+				}
```
chasdr: looks like standup landed on letting `--force` downgrade. this still runs unconditionally, so `--force` can't get past it. gate it?

```suggestion
			if !force {
				if err := refuseDowngrade(targetVersion); err != nil {
					return err
				}
			}
```

(and the "refuses to install older" line in README + self.md could mention the `--force` escape hatch)

taras-pokornyy: Fixed in commit : [2af4af1](https://github.com/datarobot-oss/cli/pull/788/commits/2af4af1aa1a874727c86bee82ae4eb1f2e1c1f64)

ajalon1 (approved):

cmd/self/update/cmd.go
ajalon1: This is good to go as is, I might post a followup JIRA for this work.

chasdr (approved):
LGTM 👍

chasdr:
sorry for the slow turnaround here, taras. was there something blocking merge, or were you waiting on my review?

PR #791, opened 2026-08-19
[CFX-7682] Follow-up code review validation

carsongee:
# RATIONALE

Create more consistent behavior, cover more recursive infinite loop conditions, clean up tests from this review: https://github.com/datarobot-oss/cli/pull/538#pullrequestreview-4967186304

## PR Automation

> [!IMPORTANT]
>

---

> [!NOTE]
> **Medium Risk**
> Changes the shared task discovery/run path used by `dr run` and `dr task list`, including new env-based recursion guards; behavior is heavily covered by new tests but mistakes could still affect everyday task execution in template projects.
> 
> **Overview**
> **Unifies Taskfile resolution** so `dr task list`, `dr run`, and shell completion all use shared `task.ResolveTaskfile`, matching what `dr run` executes and avoiding extra `Taskfile.gen.yaml` regeneration when listing tasks.
> 
> **Adds recursion protection** for generated tasks that call `dr run` on themselves (e.g. `lint: dr task run lint`): `GuardRunStack` tracks `DATAROBOT_CLI_TASK_RUN_FROM_ROOT` / `DATAROBOT_CLI_TASK_STACK` and fails with *task recursion detected* instead of fork-bombing. Run path injects the stack env into the task runner.
> 
> **Test and hygiene updates**: feature-gated root command tests skip when gates are on; list JSON/text tests run against temp fixtures; new resolve/guard unit tests and an out-of-process recursion test; template detection uses `repo.IsTemplateDir` instead of a local helper.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 6862650b4232c0132a5e4f347ddc4945efd5af61. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/cmd/root_test.go
+++ b/cmd/root_test.go
@@ -16,8 +16,11 @@ package cmd
 
 import (
 	"bytes"
+	"strings"
 	"testing"
 
+	"github.com/datarobot/cli/internal/config"
+	"github.com/datarobot/cli/internal/features"
 	"github.com/datarobot/cli/internal/misc/reader"
 	"github.com/datarobot/cli/internal/telemetry"
 	"github.com/datarobot/cli/internal/tools"
@@ -235,6 +238,21 @@ func TestUnknownArgGuard(t *testing.T) {
 	}
 }
 
+// requireGateDisabled skips the calling test when the named feature gate is
+// enabled in the ambient environment. Gate filtering happens in init(), long
+// before a test can unset the variable, so the only sound option is to skip.
+func requireGateDisabled(t *testing.T, name string) {
+	t.Helper()
+
+	if !features.Enabled(name) {
+		return
+	}
+
+	t.Skipf("feature gate %q is enabled in the environment (%sFEATURE_%s); "+
+		"gating is applied at init() so this default-state test cannot run",
+		name, config.EnvPrefix, strings.ToUpper(strings.ReplaceAll(name, "-", "_")))
+}
+
 func TestSetUnknownArgGuards_AppliesGuardToPureParent(t *testing.T) {
 	child := &cobra.Command{
 		Use:  "child",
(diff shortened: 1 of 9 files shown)


cmd/root_test.go
```diff
@@ -235,6 +238,21 @@ func TestUnknownArgGuard(t *testing.T) {
...
+
+	if !features.Enabled(name) {
+		return
+	}
+
+	t.Skipf("feature gate %q is enabled in the environment (%sFEATURE_%s); "+
+		"gating is applied at init() so this default-state test cannot run",
+		name, config.EnvPrefix, strings.ToUpper(strings.ReplaceAll(name, "-", "_")))
+}
+
```
carsongee, line 255: Unrelated to the core change in this PR. It was just leaky. I'm interested in replacing this with a root cmd factory pattern that k8s and gh are using that would let us do test <-> shell isolation for simpler testing, but that would be a follow up and a discussion

ajalon1:
@carsongee you get a 🤖 review for tonight, will take a look tomorrow.

---

Review complete: I did not find any remaining infinite-loop paths in PR #791.

Key points:
- `internal/task/resolve.go` now introduces `GuardRunStack`, which canonicalizes the Taskfile path, tracks active `(Taskfile, task)` pairs in `DATAROBOT_CLI_TASK_STACK`, and fails fast when a task attempts to re-enter itself (directly or indirectly). Because the environment assignment is only pushed down to the child process, ancestors are preserved while siblings can run sequentially without false positives.
- `ResolveTaskfile` now honors `DATAROBOT_CLI_TASK_RUN_FROM_ROOT`, forcing a re-entrant `dr run` that originated from a committed root Taskfile to switch over to the generated Taskfile. That breaks the previous loop where a root Taskfile shelled back into `dr run` for the same file.
- `cmd/task/run/cmd.go` wires the new guard (and propagates both stack and root-run markers through `RunOpts.Env`), so every `dr run` invocation passes the guard state to the spawned `task` binary and any nested `dr run`.
- The regression tests in `cmd/task/run/cmd_test.go`, `internal/task/resolve_test.go`, and the new `cmd/task/list` fixtures exercise the loop guard and confirm that `dr run lint` now exits instead of fork-bombing when the generated Taskfile invokes itself.

No corrective changes are required; the guard appears robust for the infinite-loop scenarios you highlighted.

ajalon1 (approved):
I went through the possibly silly exercise of cherry-picking your tests back to `main`, stubbing out a few functions, and seeing that yes -- the infinite loop happens. So the tests properly catch the regression.

Otherwise? Send it.

PR #795, opened 2026-08-19
[RAPTOR-19406] fix(workload): roll into the existing artifact repository

wojtekwdr:
# RATIONALE

Every roll minted an artifact with no `artifactRepositoryId`, so the platform opened a new repository each time instead of adding a version to the existing one. Deploying a workload twice left two Registry rows with the same name, and `--lock` could never produce a numbered sequence, since numbering is per repository.

## CHANGES

`Look` reads the repository off the live artifact, and the roll sends it back on create. Three calls worth arguing about:

**Read live, never stored.** The ticket proposes recording the id in `.datarobot/workload/config.json`. Reading it off the live artifact costs no extra call and keeps this independent of RAPTOR-19516: the repository follows the workload rather than the checkout, so under `--env` each manifest's own `workloadId` already gives it its own lineage. Storing it would add a fifth row to 19516's single-slot table and a field to migrate later.

**The id goes on the create payload and never into the compiled manifest.** There it would be a field the plan measures, so it would read as drift the file can never satisfy, and it would break `describes()` so every draft left by an older release gets replaced instead of reused.

**Both 404 and 422 retry once without it.** A repository you cannot write to answers 404; one whose type disagrees answers 422. The check before the create predicts the second from the live artifact rather than from the repository itself, so it can be wrong, and a wrong guess would otherwise turn a deploy that worked into one that fails. Dropping the id cannot make a refused artifact acceptable, so an unrelated failure fails again and that answer is the one reported.

One unrelated-looking change, found while doing the above: the plan compared artifact types case-sensitively and the manifest ledger enforces no enum on the field, so a hand-written `type: Service` was drift nothing could resolve, and every run would mint a version and roll a workload nobody had touched. Both comparisons now go through one helper that folds case.

Step 4 of the ticket, grouping in `dr artifact list`, is deliberately not here. It is a gated surface change, and the table already has a `VERSION ID` column holding the catalog version, so putting an artifact `VERSION` beside it needs a naming decision this PR should not make.

## PR Automation

---

> [!NOTE]
> **Medium Risk**
> Changes the core `dr workload up` roll/create path and artifact API payloads; behavior is heavily covered by new integration tests but mistaken repository handling could still affect production deploy lineage.
> 
> **Overview**
> **Workload rolls** now read `artifactRepositoryId` from the live artifact and attach it only to the **create** payload (not the compiled manifest), so each roll adds a version in the same repository instead of opening a new one. If the platform refuses that repository (**404** or **422**, e.g. type mismatch), the CLI retries once **without** the id so deploys still complete. Leftover drafts are reused only when they belong to the **same** repository as the running version.
> 
> **Drift planning** compares artifact types with **case folding** (`Service` vs `service` is no longer perpetual drift). Wizard/live defaults share **`ArtifactTypeOrDefault`** for missing types.
> 
> **`dr artifact get`** (text and JSON) exposes **Repository** and **Repository version** (`version` is null for drafts). Docs for `artifact get` are updated accordingly.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit d4d2cc1d67938ed752ab079773937a3e91bba3b5. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/internal/workload/up/version.go
+++ b/internal/workload/up/version.go
@@ -16,8 +16,12 @@ package up
 
 import (
 	"encoding/json"
+	"errors"
 	"fmt"
+	"net/http"
 
+	"github.com/datarobot/cli/internal/drapi"
+	"github.com/datarobot/cli/internal/log"
 	"github.com/datarobot/cli/internal/workload"
 	"github.com/datarobot/cli/internal/workload/sync"
 	"github.com/datarobot/cli/internal/workload/wapi"
@@ -55,8 +59,15 @@ type version struct {
 //
 // Every step here is recoverable except by leaving a draft artifact behind,
 // which is why the next run looks for one before making another.
-func buildVersion(loaded Loaded, live Live, code CodeChange, opts Options, report *reporter) (version, error) {
-	made, err := versionArtifact(loaded, live, report)
+func buildVersion(
+	loaded Loaded,
+	live Live,
+	code CodeChange,
+	repositoryID string,
+	opts Options,
+	report *reporter,
+) (version, error) {
+	made, err := versionArtifact(loaded, live, repositoryID, report)
 	if err != nil {
 		return made, err
 	}
@@ -86,18 +97,21 @@ func buildVersion(loaded Loaded, live Live, code CodeChange, opts Options, repor
 
 // versionArtifact returns the artifact to fill and build, reusing the one an
 // earlier attempt left behind rather than adding to a pile of drafts.
-func versionArtifact(loaded Loaded, live Live, report *reporter) (version, error) {
+func versionArtifact(loaded Loaded, live Live, repositoryID string, report *reporter) (version, error) {
 	if id, ok := abandoned(loaded, live); ok {
 		report.say("  Continuing with artifact %s from an earlier attempt.\n", id)
 
 		return version{ID: id}, nil
 	}
 
-	return createVersion(loaded, report)
+	return createVersion(loaded, repositoryID, report)
 }
 
-// createVersion mints an artifact from the file's artifact block.
-func createVersion(loaded Loaded, report *reporter) (version, error) {
+// createVersion mints an artifact from the file's artifact block, into
+// repositoryID when the run has a lineage for it to join. An empty
+// repositoryID leaves the platform to open one, which is what the first deploy
+// of anything wants.
+func createVersion(loaded Loaded, repositoryID string, report *reporter) (version, error) {
 	var created *workload.Artifact
 
 	err := report.run("Creating the new version", func() error {
@@ -106,7 +120,7 @@ func createVersion(loaded Loaded, report *reporter) (version, error) {
 			return payloadErr
 		}
 
-		artifact, createErr := createArtifactFn(payload)
+		artifact, createErr := createInRepository(payload, repositoryID)
 		created = artifact
 
 		return createErr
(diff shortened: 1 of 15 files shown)


internal/workload/up/version.go
```diff
@@ -151,13 +260,28 @@ func abandoned(loaded Loaded, live Live) (string, bool) {
...
 		return "", false
 	}
 
 	return cfg.ArtifactID, true
 }
 
+// joinable reports whether promoting this draft would leave the workload in
+// the repository it is already in. A workload in none has nothing to lose,
+// which is every workload created before repositories were sent at all.
+func joinable(draft *workload.Artifact, live Live) bool {
```
chasdr: Hi - not a blocker, one corner on the reuse gate: build-mode roll, file changes type service→agent, build fails after the create. run 1 mints the agent draft, the platform opens a repo for it, the project links to it. run 2: sameRepository already said "" because the types differ, but joinable compares the draft's repo against the live service's, so the draft that describes() the file is skipped and a fresh one is minted into yet another repo. every failed attempt after a type change adds a draft and a registry row, which is the pile the reuse comment says it exists to prevent.

since sameRepository already decided the roll is leaving the live lineage, should joinable compare against sameRepository(loaded, live) instead of live.ArtifactRepositoryID, and reuse anything when the intended repo is ""?

internal/workload/up/version.go
```diff
@@ -118,6 +132,91 @@ func createVersion(loaded Loaded, report *reporter) (version, error) {
...
+	if err != nil {
+		return nil, err
+	}
+
+	artifact, err := createArtifactFn(joined)
+	if err == nil || !repositoryRefused(err) {
+		return artifact, err
+	}
+
+	log.Debug("artifact repository refused; creating the version in a new one",
```
chasdr, line 203: Hi - not a blocker: when the fallback fires, the lineage fork only shows at debug level. someone whose repository was deleted or unshared sees a clean deploy while the registry quietly starts a second row. worth a report.say line here so it's visible without --debug?

internal/workload/wizard/interactive.go
```diff
@@ -43,7 +43,7 @@ const execEnvPickLimit = 200
 // about (the runtime sizing) and the binding, which would otherwise be lost
 // and turn a bind into a create.
 func (a Answers) partialDraft(detected Detected) manifest.Draft {
-	kind := orDefault(a.Type, manifest.TypeService)
+	kind := manifest.ArtifactTypeOrDefault(a.Type)
```
chasdr, line 46: Hi - not a blocker: Answers.draft (answers.go:203) still has the hand-spelled `orDefault(a.Type, manifest.TypeService)`, which is the two-readers-answering-differently case ArtifactTypeOrDefault's own comment warns about. same one-line swap as here?

chasdr (approved):
LGTM 👍

internal/workload/up/version.go
```diff
@@ -118,6 +132,91 @@ func createVersion(loaded Loaded, report *reporter) (version, error) {
...
+		"artifact_repository_id", repositoryID, "err", err)
+
+	return createArtifactFn(payload)
+}
```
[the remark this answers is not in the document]
wojtekwdr: You're right that the block arrives carrying whatever the file wrote, and the retry resent it.

Fixed by sanitizing once: `createInRepository` builds a copy with `artifactRepositoryId` deleted, adds the live id only to the first attempt, and the retry sends the clean copy. The no-repository branch sends the clean copy too, so a hand-written id can no longer defeat the type change that `sameRepository` deliberately refused. `describes()` now skips the key for the same reason: a draft can never carry the id the file wrote, so comparing it would refuse every leftover for good and mint a draft per attempt.

The repository is the workload's own or none, and a line in the file does not get a vote.

internal/workload/up/version.go
wojtekwdr: Good catch, that is the pile the reuse comment claims to prevent.

Done as you suggested: `joinable` now takes the repository the roll is heading for, which is `sameRepository`'s answer rather than the live artifact's, and reuses anything when that is "". Once the run has decided to leave the live lineage there is nothing left to protect, and measuring against the lineage being left would refuse the draft the previous attempt just minted.

internal/workload/up/version.go
wojtekwdr: Added. `createInRepository` reports whether the fallback fired and `createVersion` prints it once the phase is done, since a line written inside the phase would land under the spinner:

```
  Artifact repository 68c0…f1 would not take the new version, so it starts one of its own.
```

internal/workload/wizard/interactive.go
wojtekwdr: Swapped, thanks. Both readers spell it the same way now.

PR #799, opened 2026-08-19
feat(cmd): introduce RootFactory pattern for isolated command assembly

ajalon1:
## RATIONALE

As Carson noted, the CLI's root command setup uses package-level globals and `init()` side effects inherited from a basic cobra template. This creates two concrete problems:

1. **Env-var leakage in tests** — viperx state set by one test bleeds into the next because the production `RootCmd` singleton (built at `init()` time) reads live env vars on every Execute call.
2. **Inconsistent telemetry context** — properties like `command_kind` and `non_interactive` have to be stamped ad-hoc inside `PersistentPreRunE` against a shared global, making it easy to miss them in new commands or test scenarios.

GitHub CLI (`pkg/cmdutil.Factory`) and kubectl (`pkg/cmd/util.Factory`) both solve this by assembling the root command via an injectable factory whose dependencies can be swapped in tests. This PR brings the same pattern to the DataRobot CLI.

This PR also absorbs the `non_interactive` telemetry work that was in PR #797 (now merged here) — centralizing prompt-skip detection and Amplitude metadata in one place so individual commands never re-implement the `--yes || viperx` pattern themselves.

This PR **does not** fix env-var leakage in existing tests that use `RootCmd()` directly — that's #800.

### New: `cmd/root_factory.go`

- Defines `RootFactory`, a `Dependencies` struct, and six `With*` functional option helpers (`WithConfigInitializer`, `WithTLSSetup`, `WithTelemetryProps`, `WithTelemetryClient`, `WithAnimation`, `WithPluginRegistrar`).
- `Build()` returns a fresh, fully-wired `*cli.CommandAdder` per call — no shared mutable state between calls.
- All wiring that was in `init()` (flag registration, viper binding, group/subcommand registration, help overrides, plugin discovery, unknown-arg guards) now runs per-build inside factory methods.

### New: `cmd/root_helpers.go`

Extracts `showFirstRunAnimation` and `setUnknownArgGuards` from the old `root.go` into a focused helper file so the factory can call them without import cycles.

### Refactored: `cmd/root.go`

Reduced to ~90 lines. Its only jobs now are:

1. Register import-cycle-breaking function values (`allCommandsOutputFn`, `runVersionCommandFn`).
2. Build the production singleton via `NewRootFactory()`.
3. Expose the backward-compatible `RootCmd` package-level var so existing tests need no changes.

### Updated: `cmd/exit.go`

Reads the telemetry client from `productionFactory.TelemetryClient()` at flush time rather than a bare package-level pointer.

### New: `internal/cli/flags.go`

Defines `YesFlagName = "yes"` as the single canonical constant for `--yes`/`-y` flag registration so shared helpers can consume it without magic strings.

### New: `internal/cli/runtime.go` + `runtime_test.go`

`IsNonInteractive(cmd)` centralizes the three-source check — `force-interactive` viper key, `DATAROBOT_CLI_NON_INTERACTIVE` env var, `--yes` flag — so individual commands call one function instead of inlining the two-line OR pattern. Three unit tests cover all paths.

### New: `internal/telemetry/interaction.go` + `interaction_test.go`

`StampInteractionMode` is called once in the factory's `persistentPreRun`. It now delegates to `cli.IsNonInteractive` — no local `yesFlagName`/`hasYesFlag` helpers needed. Three unit tests: env-var path, `--yes` flag path, `force-interactive` override.

### Updated: `internal/telemetry/properties.go` + `properties_test.go`

- Adds `NonInteractive bool` to `CommonProperties`, seeded from `reader.IsNonInteractive()` at collection time.
- Adds `"non_interactive"` to `AsMap()` so it appears on every Amplitude event.

### Command migrations (from #797)

All replace the two-line `yesFlag, _ := cmd.Flags().GetBool("yes"); yes := yesFlag || viperx.GetBool("yes")` pattern with `cli.IsNonInteractive(cmd)`:

- `cmd/artifact/code/checkout`, `codesync`, `init`
- `cmd/artifact/del`
- `cmd/dependencies/install` (+ test)
- `cmd/dotenv`
- `cmd/plugin/install`
- `cmd/start`
- `cmd/workload/config`, `del`, `up`

### Docs

`AGENTS.md`, `docs/development/flags.md`, `docs/development/telemetry.md` describe `IsNonInteractive` and `YesFlagName` as the canonical patterns going forward.

## TESTING

- `go build ./...` — clean
- `go test -race ./cmd/... ./internal/cli/... ./internal/telemetry/...` — all green
- `task lint` — 0 issues across linux, darwin, and windows
- All pre-commit hooks pass

## RELATED

- #800 — test isolation follow-up (migrates `root_test.go` to use `NewRootFactory` with stubbed deps)

## FOLLOW-UP

- TODO: Extend `telemetry.StampInteractionMode` (and its tests/docs) whenever new automation-focused flags or env overrides are introduced

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -158,11 +158,11 @@ For full details, see [docs/development/configuration.md](docs/development/confi
 - **Never bulk-bind subcommand flags to viper.** `viperx` does not expose
   `BindPFlags`. Bind only specific persistent flags explicitly via
   `viperx.BindPFlag` in `cmd/root.go::init()`.
-- **Read transient flags directly from cobra**: `cmd.Flags().GetBool("yes")`.
+- **Read transient flags directly from cobra**: e.g. `cmd.Flags().GetBool(cli.YesFlagName)`.
   Do not bind them with `viperx.BindPFlag`.
 - **Env-var override for a transient flag:** register only the env var via
-  `viperx.BindEnv(key, "DATAROBOT_CLI_…")` and OR the two sources at the call site:
-  `yesFlag, _ := cmd.Flags().GetBool("yes"); yes := yesFlag || viperx.GetBool("yes")`.
+  `viperx.BindEnv(key, "DATAROBOT_CLI_…")` and merge the sources with
+  `cli.IsNonInteractive(cmd)` rather than inlining the flag/env OR yourself.
 - **To make a key persistable**, add it to `config.PersistableKeys` and have the
   write site call `config.UpdateConfigFile("my-key")`.
 
--- a/cmd/artifact/code/checkout/cmd.go
+++ b/cmd/artifact/code/checkout/cmd.go
@@ -23,6 +23,7 @@ import (
 	"github.com/datarobot/cli/cmd/artifact/code/internal/dirprompt"
 	"github.com/datarobot/cli/cmd/artifact/code/internal/format"
 	"github.com/datarobot/cli/internal/auth"
+	"github.com/datarobot/cli/internal/cli"
 	"github.com/datarobot/cli/internal/config/viperx"
 	"github.com/datarobot/cli/internal/drapi/filesapi"
 	"github.com/datarobot/cli/internal/outputformat"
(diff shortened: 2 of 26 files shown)


ajalon1:
> This PR also ports the non_interactive telemetry work from PR https://github.com/datarobot-oss/cli/pull/797, which is not throwaway — it now lives inside the factory's pre-run hook so every command event carries consistent interaction-mode metadata.

I've ported the work, but I would rather get 797 merged first and then rebase this one, since getting telemetry is more important than this refactor.

ajalon1:
@bugbot review

PR #802, opened 2026-08-20
feat(cmd): introduce RootFactory pattern for isolated command assembly

ajalon1:
## RATIONALE

As Carson noted, the CLI's root command setup uses package-level globals and `init()` side effects inherited from a basic cobra template. This creates two concrete problems:

1. **Env-var leakage in tests** — viperx state set by one test bleeds into the next because the production `RootCmd` singleton (built at `init()` time) reads live env vars on every Execute call.
2. **Inconsistent telemetry context** — properties like `command_kind` and `non_interactive` have to be stamped ad-hoc inside `PersistentPreRunE` against a shared global, making it easy to miss them in new commands or test scenarios.

GitHub CLI (`pkg/cmdutil.Factory`) and kubectl (`pkg/cmd/util.Factory`) both solve this by assembling the root command via an injectable factory whose dependencies can be swapped in tests. This PR brings the same pattern to the DataRobot CLI.

This PR **does not** fix env-var leakage in existing tests that use `RootCmd()` directly — that's #811. Telemetry/yes-flag centralization and command migrations are in the stacked #803.

## DIAGRAMS?

If you want all the gory details, go to [docs/development/root-factory.md](https://github.com/datarobot-oss/cli/blob/862e5f596b2039d30425101996cf43b0a3b8d50b/docs/development/root-factory.md). Here are the more important ones:

9a082257f754

b26ebd55c100

### New: `cmd/root_factory.go`

- Defines `RootFactory`, a `Dependencies` struct, and seven `With*` functional option helpers (`WithConfigInitializer`, `WithTLSSetup`, `WithTelemetryProps`, `WithTelemetryClient`, `WithAnimation`, `WithPluginRegistrar`, `WithViperBinder`).
- `Build()` returns a fresh, fully-wired `*cli.CommandAdder` per call. The cobra tree itself is self-contained; the process globals that stay shared (viper bindings, `cobra.OnFinalize`, `http.DefaultTransport`, the log package) are documented in the guide's "What stays shared" section, along with the "one live tree at a time" rule.
- All wiring that was in `init()` (flag registration, viper binding, group/subcommand registration, help overrides, plugin discovery, unknown-arg guards) now runs per-build inside factory methods.
- A process-global generation guard on `cobra.OnFinalize` ensures stale finalizers from earlier Executes no-op instead of re-tracking old events on old clients — only the most recent Execute's finalizer acts.
- `NewIsolatedRootFactory()` is the safe one-call preset for tests: every dependency no-oped (no disk reads, no env binding, no plugin discovery, no viper binding, no telemetry transmission), with per-test overrides still available.

### New: `cmd/root_factory_test.go`

Regression test proving two sequentially executed trees each track exactly one telemetry event (verified to fail without the `OnFinalize` guard).

### Updated: `internal/telemetry/telemetry.go`

Adds `NewTestClient`, a test seam for capturing tracked events via an injected `amplitude.Client`. Production behavior is unchanged.

### New: `cmd/root_helpers.go`

Extracts `showFirstRunAnimation` and `setUnknownArgGuards` from the old `root.go` into a focused helper file so the factory can call them without import cycles.

### Refactored: `cmd/root.go`

Reduced to ~90 lines. Its only jobs now are:

1. Register import-cycle-breaking function values (`allCommandsOutputFn`, `runVersionCommandFn`).
2. Build the production singleton via `NewRootFactory()`.
3. Expose the backward-compatible `RootCmd` package-level var so existing tests need no changes.

### Updated: `cmd/exit.go`

Reads the telemetry client from `productionFactory.TelemetryClient()` at flush time rather than a bare package-level pointer.

## TESTING

- `go build ./...` — clean
- `go test -race ./cmd/... ./internal/cli/... ./internal/telemetry/...` — all green
- `task lint` — 0 issues across linux, darwin, and windows
- All pre-commit hooks pass

## RELATED

- #803 — stacked: telemetry/yes-flag centralization and command migrations
- #810 — stacked: thread `cmd` through `plugin install` confirm path
- #811 — stacked: migrate `root_test.go` to use `NewRootFactory` with stubbed deps

## FOLLOW-UP

- `IOStreams` (stdin/stdout/stderr injection, à la gh's factory) as the next `Dependencies` field — separate PR
- Viper instance-mode (injected viper per tree) is tracked separately in the backlog; the guide documents the current shared-state limitations instead

---

> [!NOTE]
> **Medium Risk**
> Refactors CLI bootstrap (config, TLS, telemetry flush, cobra hooks) while keeping the production singleton. Behavior is meant to be unchanged, but mistakes in pre-run/finalize wiring could affect every command.
> 
> **Overview**
> Introduces an injectable `RootFactory` so tests can build a fresh cobra tree with stubbed config, TLS, telemetry, plugins, and viper binding, instead of sharing the `init()`-built `RootCmd` singleton.
> 
> Production still uses that singleton (`RootCmd` / `ExecuteContext` unchanged for callers). Wiring that lived in `root.go` now runs per `Build()`: flags, groups, subcommands, help, plugin discovery, and unknown-arg guards. `cmd.Exit` flushes telemetry from `productionFactory` rather than a package-level client pointer.
> 
> `NewIsolatedRootFactory` no-ops those side effects for unit tests. A generation guard on `cobra.OnFinalize` prevents stale telemetry re-tracks across sequential Executes. Docs call out remaining process-global state (viper, HTTP transport, logger) and the “one live tree at a time” rule.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit a0fa61934168d54db854871e756088d1f5cc3f18. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/cmd/root_factory.go
+++ b/cmd/root_factory.go
@@ -29,6 +29,7 @@ import (
 	"context"
 	"fmt"
 	"strings"
+	"sync/atomic"
 	"time"
 
 	"github.com/datarobot/cli/cmd/artifact"
@@ -179,6 +180,13 @@ func WithPluginRegistrar(fn PluginRegistrarFunc) Option {
 // ---------------------------------------------------------------------------
 // RootFactory
 // ---------------------------------------------------------------------------
+
+// finalizerGen is a process-global Execute counter shared by ALL factories.
+// It must be package-level (not per-factory) because cobra's finalizer list
+// is process-global: finalizers registered by one factory's tree are replayed
+// when another factory's tree executes, so the guard can only work if every
+// factory compares against the same counter. See persistentPreRun.
+var finalizerGen atomic.Int64
 
 // RootFactory assembles the root cobra command from a set of injectable
 // Dependencies. Each call to Build() produces a fresh, fully-wired
@@ -381,7 +389,21 @@ func (f *RootFactory) persistentPreRun(cmd *cobra.Command, args []string) error 
 	// Store telemetry client in context for use by sub-commands.
 	cmd.SetContext(context.WithValue(cmd.Context(), telemetry.ClientContextKey{}, client))
 
+	// cobra.OnFinalize appends to a process-global, append-only list that is
+	// replayed in full after EVERY Execute in the process — of any tree,
+	// built by any factory. Without a guard, execute N would re-track the
+	// events of executes 1..N-1 on their old (captured) clients — harmless
+	// in production (one Execute per process) but a source of phantom
+	// duplicate events in tests. The process-global generation counter
+	// ensures only the finalizer registered by the most recent Execute acts;
+	// stale finalizers from earlier executes no-op.
+	gen := finalizerGen.Add(1)
+
 	cobra.OnFinalize(func() {
+		if gen != finalizerGen.Load() {
+			return
+		}
+
 		if event, ok := telemetry.EventFor(cmd, args); ok {
 			client.Track(event)
 		}
--- a/cmd/root_factory.go
+++ b/cmd/root_factory.go
@@ -12,6 +12,17 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+// root_factory.go defines RootFactory — the injectable constructor for the root
+// cobra command. Call NewRootFactory (with optional With* overrides) to obtain
+// a fully-wired *cli.CommandAdder whose dependencies can be swapped for test
+// doubles. Production code relies on the singleton built in root.go's init().
+//
+// Extending the factory:
+//   - To add a new command: register it in addSubcommands.
+//   - To add a new injectable behaviour: define a *Func type, add a field to
+//     Dependencies, add a With* option, wire it in the appropriate method, and
+//     set a production default in DefaultDependencies.
+//   - See docs/development/root-factory.md for worked examples.
 package cmd
 
 import (
--- a/cmd/root_factory.go
+++ b/cmd/root_factory.go
@@ -62,9 +62,10 @@ import (
 // ---------------------------------------------------------------------------
 
 // ConfigInitializerFunc initializes viperx and reads the config file. It
-// receives the cobra command being executed so flag values are accessible.
-// Replace this in tests to skip disk I/O and prevent env-var bleed.
-type ConfigInitializerFunc func(cmd *cobra.Command, configFilePath string) error
+// receives the cobra command being executed so flag values (e.g. --config)
+// are accessible. Replace this in tests to skip disk I/O and prevent
+// env-var bleed.
+type ConfigInitializerFunc func(cmd *cobra.Command) error
 
 // TLSSetupFunc configures http.DefaultTransport based on TLS-related flags
 // and the persisted ca-cert config value. Replace in tests to be a no-op.
@@ -189,10 +190,6 @@ func WithPluginRegistrar(fn PluginRegistrarFunc) Option {
 type RootFactory struct {
 	// deps holds the injectable collaborators for this factory.
 	deps Dependencies
-
-	// configFilePath is the per-build config file override, populated from
-	// the --config flag or DATAROBOT_CLI_CONFIG env var during Build().
-	configFilePath string
 
 	// telemetryClient holds the active client for the current process-level
 	// build. It is also stored in the cobra context, but we keep a reference
@@ -336,7 +333,7 @@ func (f *RootFactory) persistentPreRun(cmd *cobra.Command, args []string) error 
 	cmd.SilenceUsage = true
 
 	// Read drconfig.yaml and bind env vars into viper.
-	if err := f.deps.ConfigInitializer(cmd, f.configFilePath); err != nil {
+	if err := f.deps.ConfigInitializer(cmd); err != nil {
 		return err
 	}
 
@@ -538,7 +535,7 @@ func (f *RootFactory) configureHelp(adder *cli.CommandAdder) {
 // defaultConfigInitializer is the production ConfigInitializerFunc. It sets up
 // viper's env-var prefix, binds standard overrides, and reads drconfig.yaml from
 // disk. Tests can replace this with a no-op to skip disk I/O and isolate env state.
-func defaultConfigInitializer(cmd *cobra.Command, configFilePath string) error {
+func defaultConfigInitializer(cmd *cobra.Command) error {
 	// Map any environment variables prefixed with DATAROBOT_CLI_ to config keys.
 	viperx.SetEnvPrefix("DATAROBOT_CLI")
 	viperx.AutomaticEnv()
@@ -554,12 +551,14 @@ func defaultConfigInitializer(cmd *cobra.Command, configFilePath string) error {
 	viperx.SetDefault(config.APIConsumerTrackingEnabled, true)
 	_ = viperx.BindEnv(config.APIConsumerTrackingEnabled, "DATAROBOT_API_CONSUMER_TRACKING_ENABLED")
 
-	// Resolve the config file path: flag > env var > default.
-	resolved := configFilePath
+	// Resolve the config file path: --config flag > DATAROBOT_CLI_CONFIG env
+	// var > default location. The flag value is read from cobra directly
+	// (rather than via the viper binding) so resolution keeps working even
+	// when a test stubs out the viper binding.
+	resolved, _ := cmd.Flags().GetString("config")
+
 	if resolved == "" {
-		if envPath := viperx.GetString("config"); envPath != "" {
-			resolved = envPath
-		}
+		resolved = viperx.GetString("config")
 	}
 
 	if err := config.ReadConfigFile(resolved); err != nil {
--- a/cmd/root_factory.go
+++ b/cmd/root_factory.go
@@ -88,6 +88,15 @@ type AnimationFunc func()
 // Replace in tests with a no-op.
 type PluginRegistrarFunc func(cmd *cobra.Command)
 
+// ViperBinderFunc binds the root persistent flags to the global viper
+// instance and annotates the universal flags for forwarding to plugin
+// subprocesses. The default implementation re-points the global viper
+// bindings at the newest tree on every Build; replace with a no-op in tests
+// that must not mutate global viper state (with the caveat that viper-backed
+// reads such as debug/verbose or ca-cert then won't resolve this tree's
+// flag values).
+type ViperBinderFunc func(adder *cli.CommandAdder)
+
 // Dependencies bundles every injectable collaborator for the root command.
 // The zero-value is not useful; use DefaultDependencies() or NewRootFactory()
 // to get a properly populated instance.
@@ -109,6 +118,9 @@ type Dependencies struct {
 
 	// PluginRegistrar discovers and wires plugin subcommands.
 	PluginRegistrar PluginRegistrarFunc
+
+	// ViperBinder binds persistent flags to the global viper instance.
+	ViperBinder ViperBinderFunc
 }
 
 // DefaultDependencies returns the production-ready set of dependencies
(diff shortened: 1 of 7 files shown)


chasdr:
Hi AJ - Just saw this and I'll review it shortly

chasdr (commented):
👋 chasbot here — chas had me take a techspec pass since you asked for input. design feedback, not a merge gate, agent-to-agent: my compliments to whichever of your agents drew the diagrams.

i diffed the factory against the old init() line by line hoping to catch a smuggled behavior change. nothing. deeply disappointing. the two problems i did find are both about the isolation story, not the port:

**1. Build() re-points the global viper bindings at the newest tree.** the factory hands out fresh cobra trees, but viper is still the one shared watering hole, and bindViperFlags marches every new tree straight to it on every Build. it's also the one wiring step not in Dependencies, so even the doc's "minimal no-op factory" mutates global state. reproduced locally:

    _ = RootCmd.PersistentFlags().Set("debug", "true")  // visible through viperx ✓
    _ = NewRootFactory(/* all no-op deps */).Build()    // another test builds a tree
    viperx.GetBool("debug")                             // false — binding hijacked

log.Start reads debug/verbose and setupTLS reads ca-cert/-k through those bindings, so only the last-built tree's bound flags resolve. serial build-then-execute tests work; mixing RootCmd with factory trees doesn't, and t.Parallel remains a fantasy. since rationale #1 is viper leakage: is an injected viper instance (viperx instance mode) on the stack's roadmap, or should the doc state the "one live tree at a time" rule explicitly?

**2. cobra.OnFinalize never forgets.** elephant-grade memory, no delete API: the finalizer list is append-only and replayed on every execute. inherited from the old root.go, but this stack makes multiple Executes per process the primary use case, so execute N re-tracks executes 1..N-1's stale telemetry events on their old clients. a capturing test client sees phantom duplicates. guard the closure so it fires once, or move track/flush into PersistentPostRunE plus a factory-owned Execute wrapper for the error path?

cool-takes, in leverage order:
- ship a NewIsolatedRootFactory()/TestDependencies() preset in this PR rather than #811 — the doc pastes the same five stubs four times, which is the doc quietly asking for the helper. also, a default-deps Build() execs dr-* binaries off PATH via PluginRegistrar, so the safe preset should be one call away.
- the most valuable field in gh's factory is IOStreams. the injection rails now exist; IOStreams as the next Dependencies field gains command tests more than any of the current six.

everything above fits inside the pattern you already built. — chas's claude, blocking nothing, it's a draft

cmd/root_factory.go
```diff
@@ -0,0 +1,600 @@
...
+//
+// Use NewRootFactory(opts...) to construct one, or rely on the
+// package-level RootCmd variable for the production singleton.
+type RootFactory struct {
+	// deps holds the injectable collaborators for this factory.
+	deps Dependencies
+
+	// configFilePath is the per-build config file override, populated from
+	// the --config flag or DATAROBOT_CLI_CONFIG env var during Build().
+	configFilePath string
```
chasdr: is anything ever assigning this? the old StringVar(&configFilePath, ...) wasn't carried over, so it's always "" and --config resolves only through the viper-binding fallback in defaultConfigInitializer. the comment's "populated during Build()" also can't happen because flags parse at Execute. drop the field, or add a WithConfigFilePath if #811 needs one?

cmd/root_factory.go
```diff
@@ -0,0 +1,600 @@
...
+
+	client := f.deps.TelemetryClient(props)
+
+	// Store as factory-level client so cmd.Exit can flush on the main error path.
+	f.telemetryClient = client
+
+	// Store telemetry client in context for use by sub-commands.
+	cmd.SetContext(context.WithValue(cmd.Context(), telemetry.ClientContextKey{}, client))
+
+	cobra.OnFinalize(func() {
```
chasdr, line 468: carried over from the old root.go, but multiple Executes per process is now the point: this list is append-only and replayed on every execute, so stale finalizers re-track old events on old clients. guard the closure so it fires once, or move track/flush to PersistentPostRunE plus a factory Execute wrapper for the error path?

cmd/root_factory.go
```diff
@@ -0,0 +1,600 @@
...
+		o(f)
+	}
+
+	return f
+}
+
+// TelemetryClient returns the telemetry client that was wired during the
+// most recent Build() call. cmd.Exit uses this to flush events on error
+// paths where only the signal context, not a cobra context, is available.
+func (f *RootFactory) TelemetryClient() *telemetry.Client {
```
chasdr, line 288: nit: "wired during the most recent Build() call" — set in persistentPreRun at execute time, not Build. a built-but-never-executed tree returns nil here.

docs/development/root-factory.md
```diff
@@ -0,0 +1,434 @@
+# Root Command Factory
+
+## Overview
+
+The `RootFactory` (`cmd/root_factory.go`) assembles the root cobra command from
+injectable dependencies. Each call to `Build()` returns a fresh, fully-wired
+`*cli.CommandAdder` with no shared mutable state — this is the key property that
```
chasdr: "no shared mutable state" promises more than the code delivers: global viper (bindViperFlags runs on every Build and can't be stubbed), cobra.OnFinalize, http.DefaultTransport, and the log package stay shared. scope the claim to the cobra tree itself?

docs/development/root-factory.md
```diff
@@ -0,0 +1,434 @@
...
+    cmd.WithPluginRegistrar(func(_ *cobra.Command) {}),
+    cmd.WithAnimation(func() {}),
+).Build()
+```
+
+> **Why not just use the global `RootCmd`?** The global singleton is built once
+> in `init()` with the production `ConfigInitializer`. Every `Execute()` call on
+> it runs `viperx.AutomaticEnv()` and `ReadConfigFile()` against the live
+> environment. Env vars set with `t.Setenv()` are restored after the test, but
+> viper's internal cache may retain values until the next `AutomaticEnv()` sweep
```
chasdr: viper doesn't cache env lookups — AutomaticEnv reads os.Getenv live on every Get, so t.Setenv cleanup is visible immediately. the real bleed paths are viperx.Set globals, ReadConfigFile contents, and changed-flag state on the shared singleton. name those instead? this section is the mental model people will test against.

docs/development/root-factory.md
```diff
@@ -0,0 +1,434 @@
...
+    P->>P: deps.TelemetryProps()
+    P->>P: stamp CommandKind
+    P->>P: deps.TelemetryClient(props)
+    P->>P: deps.Animation()
+    P->>C: cmd.RunE()
+    C-->>M: error or nil
+    Note over F: cobra.OnFinalize flushes telemetry
+```
+
+### How #803 extends the factory
```
chasdr: if #802 merges first or #803 reshapes, this section documents symbols that don't exist on main (StampInteractionMode, IsNonInteractive, YesFlagName). move the diagram into #803's own docs (the flags.md/telemetry.md it adds)?

docs/development/root-factory.md
```diff
@@ -0,0 +1,434 @@
...
+    return nil
+}
+```
+
+That's it. Tests override it with `WithAuditLogger(func(...) error { return nil })`;
+production gets `defaultAuditLogger` for free.
+
+---
+
+## Feature-gating a command
```
chasdr, line 341: development/feature-gates.md already covers gating in more detail. keep only the factory-specific fact here (gates evaluate at Build time, so build a fresh tree per gate state) and link the rest? two copies will drift.

docs/development/root-factory.md
ajalon1: I intend to merge all of these as a GH stack, so the sequence order is set.

cmd/root_factory.go
ajalon1: Yeah, nothing's using it -- and hell, is it really necessary? I think WithConfigInitializer() is a more useful fixture reviewing this again. I'll drop configFilePath.

docs/development/root-factory.md
ajalon1: Good callout, I forgot that about GetEnv(). This actually makes me want to tackle the possibility of making viper a non-global singleton again (another story in the backlog), as viperx being global is still a PITA. But I think I'd rather just document as is, so I'll clean up the docs.

ajalon1:
> my compliments to whichever of your agents drew the diagrams.

mermaid for the win.

> 2. cobra.OnFinalize never forgets. elephant-grade memory

feels like a feature to me. :)

> guard the closure so it fires once, or move track/flush into PersistentPostRunE plus a factory-owned Execute wrapper for the error path?

Seeing these options reminds me that putting anything expectd to always run into PersistentPostRunE at the root level is not a good idea. Any subcommand can simply override PersistentPostRunE and then we won't get telemetry, with no errors or warnings. I don't think we do this at all, but it's there. (I think its a similar issue with PersistentPreRunE but I'd have to check.) So I'll tinker with "guarding the closure".

ajalon1:
Think I handled everything.

> ship a NewIsolatedRootFactory()/TestDependencies()

was a good idea -- handled that

> the most valuable field in gh's factory is IOStreams

also a great idea, but I'll leave that for a separate PR

cmd/root_factory.go
```diff
@@ -0,0 +1,694 @@
...
+	// in production (one Execute per process) but a source of phantom
+	// duplicate events in tests. The process-global generation counter
+	// ensures only the finalizer registered by the most recent Execute acts;
+	// stale finalizers from earlier executes no-op.
+	gen := finalizerGen.Add(1)
+
+	cobra.OnFinalize(func() {
+		if gen != finalizerGen.Load() {
+			return
+		}
```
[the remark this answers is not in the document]
ajalon1: I thought I did this already.

cmd/root_factory.go
```diff
@@ -0,0 +1,694 @@
...
+	setUnknownArgGuards(adder.Command)
+
+	f.configureHelp(adder)
+
+	return adder
+}
+
+// buildRootCommand constructs the bare cobra.Command with its Use, Long, and
+// PersistentPreRunE / PersistentPostRunE hooks wired to the factory's deps.
+func (f *RootFactory) buildRootCommand(outputFormat *outputformat.OutputFormat) *cobra.Command {
```
carsongee, line 327: `outputFormat` isn't read anywhere in this function body — it's threaded through `buildRootCommand` purely to be discarded. Not a bug (the real wiring happens later in `registerFlags` via `outputformat.AddPersistentFlag`), but it reads as if the root command's construction depends on it. Consider dropping the parameter here, or a one-line comment noting it's reserved for future use if that's intentional.

carsongee (commented):
Ran a code review pass focused on the RootFactory's isolation guarantees (relevant since the goal is consistent ctrl-c/analytics/env-var handling across a growing command surface, kubectl/helm-style).

Good news: the two correctness issues I found against the earlier revision (cb9f5ff) are already fixed in this push (835c242) — `configFilePath` is now read directly off `cmd.Flags().GetString("config")` instead of a dead `flags.String` return value, and the `ViperBinderFunc` injection point + `NewIsolatedRootFactory` + the `finalizerGen` guard make the "independent builds" story actually true where it matters, with the remaining shared globals (`cobra.EnableCaseInsensitive`, `http.DefaultTransport`, etc.) now explicitly documented as an accepted, idempotent exception in root-factory.md. Nice cleanup.

Two smaller things left:

1. **Stale docs (not touched by this PR, but broken by it):** `docs/development/flags.md:139-145` and `docs/development/plugins.md:275-282` both instruct contributors to call a package-level `bindUniversal(name)` from `cmd/root.go`. That function no longer exists there — it's now the unexported `bindUniversalOn` closure scoped inside `bindViperFlags` in `cmd/root_factory.go`. Following either doc as written today produces `undefined: bindUniversal`. Worth a follow-up (or a quick amendment here since the PR is what invalidated them) pointing contributors at the new registration point.
2. See inline comment on `cmd/root_factory.go` re: an unused parameter.

carsongee (approved):

cdevent (approved):
Awesome!

PR #805, opened 2026-08-20
[RAPTOR-19535] feat(workload): hide advanced config fields, make the readiness probe optional

wojtekwdr:
# RATIONALE

User feedback on `dr workload config` was that the settings step asks too much at once, and that the health-path note reads as if traffic will never flow without a readiness probe. The platform does not require one, which the same thread confirmed.

Triaging the whole screen rather than that one note turned up that it had drifted from the RAPTOR-18975 design anyway: replicas, cpu and memory were on the never-asked list and were being prompted for regardless. So most of this is putting the screen back where it was meant to be.

## CHANGES

The settings screen asks for the port. Health path, replicas, cpu, memory and importance move behind an "Advanced options" row, reachable with `ctrl+o` from anywhere on the screen or by tabbing onto the row. All five still appear on the confirm screen whether or not you open it, so nothing is written that was never shown. The shortest path drops from eleven answers to six.

A readiness probe is now declinable. An empty health path means no `readinessProbe` block, and `--no-readiness-probe` says the same thing headlessly.

**The default probe path changes from `/health` to `/`.** A probe has to pass on an application nobody wrote it for. Almost every HTTP service answers something at the root, where `/health` is an endpoint the app has to have implemented, and pointing a probe at one that does not exist is the most common first-deploy failure. It presents as a deploy that never becomes ready rather than as a misconfigured probe, which is the worst way to learn about it.

`--importance` is new. It was missing rather than removed, and hiding that field without it would have stranded the value.

Three things worth a reviewer's attention:

- **Binding no longer confuses "no probe" with "a probe we cannot read".** `Live.Defaults()` reported both as an empty path, so the health field silently discarded whatever you typed on a bound workload. Those are separate questions now; a probe whose shape this release does not model is preserved the way startup and liveness probes are, and it follows the container port when that moves.
- **A flag we cannot carry out is refused rather than ignored.** `--health` or `--no-readiness-probe` against a workload whose probe we cannot rewrite used to report success having changed nothing, and headlessly there is no screen to notice it.
- **A hidden field that fails validation opens the row and takes the cursor.** The value being refused often comes from a flag or from the bound workload rather than from typing, so the error would otherwise name a field that is not on screen.

## NOTES

Verified against staging: the default writes the probe, `--no-readiness-probe` omits exactly that block and nothing else, and the result passes `Load` and `Validate` on a second run.

Two gaps this does not close, both pre-existing and both made more visible by the new flags, so the help text says what actually happens rather than promising otherwise. `up` cannot carry a probe removal to the platform, because a key the file omits is a key `up` leaves alone by design. And `importance` is neither `artifact.spec` nor `runtime`, so it reaches the platform at create only.

The bound path could not be exercised end to end against staging, because every workload on the account trips RAPTOR-19533. It is covered by tests over recorded live documents instead.

## RELATED

RAPTOR-19535

### DIFF
--- a/internal/workload/manifest/live.go
+++ b/internal/workload/manifest/live.go
@@ -87,7 +87,6 @@ func (l Live) Defaults() Draft {
 		Type:       ArtifactTypeOrDefault(l.Type()),
 		A2AEnabled: boolAt(l.Spec, keyA2AEnabled),
 		Port:       DefaultPort,
-		HealthPath: DefaultHealthPath,
 		Build:      Build{Mode: BuildModeDockerfile},
 		Runtime:    l.runtimeDefaults(),
 	}
@@ -101,6 +100,11 @@ func (l Live) Defaults() Draft {
 		draft.Port = port
 	}
 
+	// The health path is left empty when the container has no readiness probe,
+	// rather than falling back to DefaultHealthPath like everything else here.
+	// A workload running without a probe has to arrive saying so: the screen
+	// shows the absence, and Apply can then read a path as the answer someone
+	// gave rather than as a default it must be careful not to attach.
 	if path := stringAt(mapAt(container, keyReadinessProbe), keyPath); path != "" {
 		draft.HealthPath = path
 	}
@@ -108,6 +112,28 @@ func (l Live) Defaults() Draft {
 	draft.Build = liveBuild(container)
 
 	return draft
+}
+
+// ReadinessProbe reports what the running workload does about readiness:
+// whether the primary container declares a probe at all, and whether this
+// release can read a path out of it.
+//
+// The two are separate questions, and Defaults answers neither on its own: it
+// reports a probe with no readable path the same way it reports no probe, as an
+// empty HealthPath. Only the screens care about the difference, and they care a
+// lot, because an empty field means "no probe" to everything downstream.
+func (l Live) ReadinessProbe() (present, readable bool) {
+	container := l.primaryContainer()
+	if container == nil {
+		return false, false
+	}
+
+	probe := mapAt(container, keyReadinessProbe)
+	if probe == nil {
+		return false, false
+	}
+
+	return true, stringAt(probe, keyPath) != ""
 }
 
 // BuildMode reports how the primary container's image comes to be, in the
@@ -516,11 +542,15 @@ func (l Live) primaryContainerName() string {
 	return stringAt(container, keyName)
 }
 
-// applyReadiness writes the port and, where the workload already has a
-// readiness probe, keeps it pointing at the same place. It does not add one:
-// a workload deliberately running without a probe should not acquire the
-// wizard's default /health, which a container that does not serve it would
-// never pass.
+// applyReadiness writes the port and points the readiness probe wherever the
+// answers say, which includes having no probe at all.
+//
+// The draft's path is empty unless something asked for one: Defaults leaves it
+// empty for a container running without a probe, so a workload deliberately
+// running without one cannot acquire the wizard's default /health, which a
+// container that does not serve it would never pass. What reaches here
+// non-empty was typed on a screen or passed as a flag, and adding the probe is
+// then the answer rather than an invention.
 func applyReadiness(container map[string]any, draft Draft) {
 	previous, hadPort := intAt(container, keyPort)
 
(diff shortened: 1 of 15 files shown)


internal/workload/wizard/screens.go
```diff
@@ -52,12 +60,17 @@ const namePlaceholder = "my-app-name"
...
-	// and drop --name without saying so.
-	if err := opts.Answers.checkBinding(); err != nil {
-		return nil, manifest.Draft{}, err
+	// No screen can settle these, so they are not questions the wizard gets to
+	// ask. Started anyway, the error would sit behind the loading view while
+	// the fetch ran, and arriving at the first screen would clear it and drop
+	// the flag without saying so: contradictory flags have to be refused before
+	// the terminal is taken over.
+	for _, check := range []func() error{
+		opts.Answers.checkBinding, opts.Answers.checkReadiness, opts.Answers.checkImportance,
```
damjandr: `checkReadiness` also runs `checkPort`, so a bad `--port` now aborts before the wizard opens instead of falling back to the detected port — which makes the `checkPort` fallback at interactive.go:60 dead code.

damjandr: `checkImportance` and the `--health` leading-slash rule *can* be settled by a screen, so hard-failing them here contradicts the comment above and the `--cpu`/`--memory` treatment.

internal/workload/manifest/live.go
```diff
@@ -531,21 +561,26 @@ func applyReadiness(container map[string]any, draft Draft) {
...
+	// deliberately and keeps its own port.
+	following := []string{keyStartupProbe, keyLivenessProbe}
+	if !rewritten {
+		// A readiness probe this release could not read was left alone above,
+		// so it follows the same rule as the other two rather than keeping a
+		// port the container no longer listens on.
+		following = append(following, keyReadinessProbe)
+	}
+
+	for _, key := range following {
```
damjandr, line 589: Only a top-level `port` key follows the container; the test's own `tcpSocket.port` stays on the old port — exactly the "never reports ready" case this comment says it prevents.

internal/workload/manifest/live.go
```diff
@@ -554,6 +589,44 @@ func applyReadiness(container map[string]any, draft Draft) {
...
+	case probe != nil && stringAt(probe, keyPath) == "":
+		return false
+
+	case !draft.WantsReadinessProbe():
+		// Declined. The block goes rather than being left pointing at an empty
+		// path, which is neither a probe nor an absence of one.
+		delete(container, keyReadinessProbe)
+
+	case probe != nil:
+		probe[keyPath] = draft.HealthPath
```
damjandr: `WantsReadinessProbe` trims but this writes `draft.HealthPath` raw, so an untrimmed caller writes a padded path.

internal/workload/wizard/answers.go
```diff
@@ -175,6 +192,65 @@ func (a Answers) checkSizing() error {
...
+// checkProbeEditable refuses a probe answer the CLI cannot carry out, which is
+// only possible once the workload has been read.
+//
+// A readiness probe that declares no path is not the HTTP probe these flags
+// describe, and binding preserves it rather than rewriting a shape this release
+// does not model. Preserving is right when nobody said anything about the
+// probe. When a flag did say something, staying silent would mean the run
+// reports success having ignored the one thing it was asked to change, and
+// without a screen there is nothing else to notice it.
+func (a Answers) checkProbeEditable(live manifest.Live) error {
```
damjandr, line 217: `checkProbeEditable` runs only on the headless bound path, so interactive `--no-readiness-probe` against an unreadable probe is still silently ignored.

internal/workload/wizard/screens.go
```diff
@@ -690,13 +840,15 @@ func (f flow) confirmNote() string {
 	return strings.Join([]string{
 		"Writing to " + ShortPath(manifest.Path(f.detected.Dir)),
 		f.runtimeLine(),
+		f.readinessLine(),
```
damjandr, line 853: `readinessLine()` returns `""` when hand-edited, so the `Join` leaves a blank line in the confirm note.

internal/workload/wizard/screens.go
```diff
@@ -770,6 +979,14 @@ func (f flow) keys() string {
 		bindings = append([]keyBinding{{Key: "↑/↓", Does: "scroll"}}, bindings...)
 	}
 
+	// On the row itself enter belongs to the row rather than to the screen, so
+	// it is labelled for what it does there. Only that entry changes:
+	// rebuilding the list here would quietly drop any key the screen gains
+	// later, in the one state where the legend matters most.
+	if f.onAdvancedRow() {
+		bindings = withKey(bindings, "enter", advancedLabel)
```
damjandr: `screenKeys` already carries `{ctrl+o, advanced}`, so on the row the legend renders `[ctrl+o] advanced  [enter] advanced` — two entries, same word.

internal/workload/wizard/model.go
```diff
@@ -834,6 +881,43 @@ func (f *flow) acceptSettings() (tea.Cmd, error) {
...
+// reveal puts the cursor on the field a validation error is about, opening the
+// advanced row when that is where the field lives. The error is returned
+// unchanged so the caller reads as one statement.
+func (f *flow) reveal(field int, err error) error {
+	if field >= firstAdvancedField {
+		f.advancedOpen = true
+	}
+
+	f.focus = field
+	f.applyFocus()
```
damjandr: `reveal` drops `applyFocus`'s cmd, which `applyFocus`'s own doc says has to reach the caller.

internal/workload/wizard/screens.go
```diff
@@ -225,17 +238,18 @@ func (f *flow) enterInputs(at screen) {
...
 
 		f.inputs = inputs
 
-		for i := range f.inputs[1:] {
-			f.inputs[i+1].Blur()
-		}
+		// newInput focuses everything it builds, so exactly one is left
+		// focused here. Through applyFocus rather than a hardcoded index, so
+		// arrival and movement agree on where the cursor is by construction.
+		f.applyFocus()
```
damjandr: Same here — the cmd is discarded on arrival, so the caret is static until the first keystroke.

internal/workload/wizard/model.go
```diff
@@ -282,10 +289,21 @@ func (f flow) key(msg tea.KeyMsg) (tea.Model, tea.Cmd) {
 
 	f.failed = nil
 
+	if f.at == screenSettings && msg.String() == advancedKey {
+		return f, f.toggleAdvanced()
```
damjandr: `return f, f.toggleAdvanced()` mutates `f` through a pointer receiver inside the return list; Go leaves operand-vs-call order unspecified. Same at 304 and 373 — assign the cmd first.

internal/workload/wizard/model.go
```diff
@@ -798,32 +841,36 @@ func (f *flow) acceptSettings() (tea.Cmd, error) {
 	}
 
 	path := f.field(fieldHealthPath)
-	if !strings.HasPrefix(path, "/") {
-		return nil, errors.New("the health path must start with /")
+	if err := f.acceptHealthPath(path); err != nil {
+		return nil, f.reveal(fieldHealthPath, err)
```
damjandr, line 861: The two port failures above return without `reveal`, so the cursor can sit in a hidden field while the error names the port.

internal/workload/wizard/screens.go
```diff
@@ -669,7 +815,11 @@ func (f flow) scrollablePreview() bool {
 const (
 	defaultPreviewHeight = 20
 	minPreviewHeight     = 6
-	previewChrome        = 12
+	// Grown twice over: the confirm note now carries a readiness line as well
+	// as the runtime one, and the runtime line wraps on an 80-column terminal
+	// once importance is on it. Under-reserving here does not shrink the
+	// preview, it pushes the key legend off the bottom of the screen.
+	previewChrome = 15
```
damjandr: 15 is hand-tuned to a wrapped 80-col runtime line plus an optional readiness line; it over-reserves whenever either is absent.

internal/workload/wizard/model.go
```diff
@@ -1040,14 +1124,82 @@ func isFocusKey(msg tea.KeyMsg) bool {
...
+		f.focus = fieldPort
+	}
+
+	return f.applyFocus()
+}
+
+// settingsStops is the order the cursor visits, built from the one field order
+// so a field added to it needs no second edit here.
+func (f flow) settingsStops() []int {
+	stops := []int{fieldPort, advancedStop}
```
damjandr, line 1214: Comment overclaims: any new *non*-advanced field still needs this literal edited.

internal/workload/wizard/screens.go
```diff
@@ -758,6 +949,24 @@ var defaultKeys = []keyBinding{
 	{Key: "esc", Does: "back"},
 }
 
+// withKey re-labels one key in a screen's legend, copying first because
+// screenKeys is a package-level map and the slices in it are shared by every
+// render. A key the screen does not have is added rather than dropped, so a
+// caller cannot silently say nothing.
+func withKey(bindings []keyBinding, key, does string) []keyBinding {
```
damjandr, line 967: The append branch is unreachable — `screenSettings` always has `enter`.

damjandr (approved):
Overall good, some minor things.

internal/workload/wizard/screens.go
wojtekwdr: fixed

wojtekwdr: fixed

internal/workload/manifest/live.go
wojtekwdr: fixed

internal/workload/wizard/model.go
wojtekwdr: fixed

PR #806, opened 2026-08-20
[RAPTOR-19601] feat(workload): warn when a deploy runs a draft artifact

wojtekwdr:
# RATIONALE

A deploy onto a draft artifact is on a clock and the CLI never mentioned it. You deploy, share the endpoint, and it is dead a few hours later with nothing in the output that hinted it would be. The remedy already existed as `dr workload up --lock`, discoverable only by reading `--help`.

## CHANGES

One warning on stderr when a run leaves a draft artifact serving, naming the deadline and the command that makes it permanent. On those runs the `Next:` list spends its third slot on `--lock` instead of `stop`.

`up.Result.Locked` already carries the state on every path (create, roll, retune, start, unchanged, dry run), so this needs no extra API call. Everything lands in `cmd/workload/up/`; `internal/workload/up` is untouched.

Silent on locked artifacts, `--lock` runs, failed deploys, and `--output-format json`.

```
  ⚠ This workload is running a draft artifact.
    Draft workloads are stopped after 8 hours, whether or not they are in use.
    Run 'dr workload up --lock' to version the artifact and make it permanent.

Next:
  dr workload up --lock       Lock the artifact to make it permanent
  dr workload logs <id>       View the container logs
  dr workload status <id>     Check the workload status
```

## NOTES

The ticket said not to print "8 hours" because the TTL varies by cluster. It does not. `CUTOFF_HOURS = 8` in workload-api `jobs/stop_draft_protons.py` is a module constant with no env, config or Helm override, and the figure is already published in the OpenAPI description of an artifact status.

Two further things that source settled, both of which an earlier draft of the warning had wrong: nothing deletes the draft artifact, only the running workload is stopped, and the sweep selects on how long the workload has been running rather than on traffic, so this is not an inactivity timeout.

The `Next:` descriptions are now imperative technical phrasing. Three of the four predate this change.

---

> [!NOTE]
> **Low Risk**
> User-facing CLI messaging and next-step hints only; deploy behavior and `internal/workload/up` are unchanged.
> 
> **Overview**
> **`dr workload up`** now warns on **stderr** when a run leaves (or would leave) a **draft** artifact serving, so users see that the workload is stopped after **8 hours** regardless of traffic and how to make it permanent with **`dr workload up --lock`**.
> 
> Detection uses existing **`Result.Locked`** plus flags (**`--lock`**, **`--dry-run`**, failure, no workload)—no extra API calls. The warning is **suppressed** for locked artifacts, **`--lock`** runs, failed deploys, runs with no workload, and **JSON** output (machines still read **`locked`** in the envelope).
> 
> The **Next:** hints change for draft deploys: **`--lock`** is listed first instead of **`stop`**; non-draft runs still get **`stop`**. Dry runs get the remedy in the warning because they never reach **Next:**.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit e8cabeda28df91311507c01227038c080dcba467. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/cmd/workload/up/cmd.go
+++ b/cmd/workload/up/cmd.go
@@ -337,14 +337,13 @@ func reportable(result up.Result) bool {
 }
 
 // draftIsServing reports whether this run left, or would leave, a draft
-// artifact serving. A draft is mutable and unversioned, the platform reclaims
-// it, and the workload stops on its own once it has been idle, none of which
-// the deploy output ever said.
+// artifact serving. A workload on a draft is stopped eight hours after it
+// starts running, whatever it is serving, and the deploy output never said so.
 //
-// Result.Locked is the whole signal and costs no extra call. It starts as the
-// status of the artifact the workload is already running and is only ever
-// flipped true by an actual lock, so its negation covers every path a deploy
-// can take: create, roll, retune, start, and a run that changed nothing.
+// Result.Locked is the whole signal. It starts as the status of the artifact
+// the workload is already running and is only ever flipped true by an actual
+// lock, so its negation covers every path a deploy can take: create, roll,
+// retune, start, and a run that changed nothing.
 //
 // The three refusals, in order:
 //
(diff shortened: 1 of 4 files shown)


adamalpi (approved):
Careful implementation with genuinely exhaustive state coverage (tense on dry runs, suppression on failure/--lock/locked/JSON). Verified on the branch: builds and all workload suites pass; `lockedAlready` reuse from roll.go is sound; `bindLocked` correctly closes the promotion hole where a create bound to a locked artifact would be called temporary and advised to lock what the platform would 403; JSON purity holds with the envelope's `locked` field carrying the machine-readable signal.

PR #810, opened 2026-08-20
fix(plugin/install): thread cmd through confirmPluginDepsInstall

ajalon1:
## RATIONALE

Part of the RootFactory stack (#802, #803). `confirmPluginDepsInstall` was closing over a package-level `yesFlag` variable and passing `nil` to `cli.IsNonInteractive`, bypassing the proper flag-read path.

When `cmd == nil`, `IsNonInteractive` falls back to the weaker `viperx.GetBool(YesFlagName)` path instead of reading the flag directly from the command. The package-level `yesFlag` is also shared state that leaks across successive `Execute()` calls in tests or any harness that re-uses the same process.

## CHANGES

- Remove the package-level `var yesFlag bool`; re-register the flag via `cmd.Flags().BoolP` (no package-level binding)
- Thread `*cobra.Command` through: `runInstall` → `runInstallFromRegistry` / `runInstallFromFile` / `runInstallFromURL` → `checkAndInstallPluginDeps` → `confirmPluginDepsInstall`
- Replace `yesFlag || cli.IsNonInteractive(nil)` with `cli.IsNonInteractive(cmd)`, which checks `--yes` via `cmd.Flags()`, the `NON_INTERACTIVE` env var, and `viperx` in priority order
- Update tests: add `makeCmd(t, yes bool)` helper, remove all `origYesFlag`/`defer` cleanup patterns

---

> [!NOTE]
> **Low Risk**
> Small CLI plumbing refactor for non-interactive confirmation; no auth, install, or dependency-install logic changes beyond how `--yes` is read.
> 
> **Overview**
> Stops plugin-install dep confirmation from using a package-level `yesFlag` and `cli.IsNonInteractive(nil)`, which skipped the real `--yes` flag path and leaked state across `Execute()` calls.
> 
> `--yes` is now registered without a package binding. `runInstall` threads `*cobra.Command` through the file/URL/registry install paths into `confirmPluginDepsInstall`, which uses `cli.IsNonInteractive(cmd)` so `--yes`, env, and viper are resolved in the intended order. Tests use a `makeCmd` helper instead of mutating global flag state.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit e3e5760811a2fc9f8734afe41a95814c6fcf07d3. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

PR #811, opened 2026-08-20
test(cmd): migrate root_test.go to use isolated RootFactory builds

ajalon1:
## RATIONALE

Part of the RootFactory stack (#802, #803, #810). Replaces every test that called `Execute()` on the shared production `RootCmd` singleton with a fresh command tree built via `newIsolatedRootCmd()`.

(**This is what @carsongee wanted fixed. :)**)

The helper delegates to `NewIsolatedRootFactory()`, which stubs every side-effecting dependency with a no-op so that:
- env vars set by `t.Setenv()` in one test cannot bleed into the next through shared viper state
- global viper flag bindings are not re-pointed at the test tree on every `Build()`
- `http.DefaultTransport` is not mutated between tests
- no disk/network I/O for device/user IDs or account info, and no telemetry client is constructed
- plugin filesystem discovery and subprocess invocations are skipped
- the first-run TTY animation is suppressed

Tests that only inspect static structure (flag registration, annotations, `TraverseChildren`) continue to reference the production `RootCmd` singleton, which is appropriate since that state is set once at `init()` time and never mutated by the structural assertions. The gate-presence tests are an exception: they use a fresh isolated build with the feature-gate env var neutralized, since `RootCmd`'s gate is baked at `init()` and cannot be unset.

`TestUniversalFlagsParsedOnCoreSubcommand` previously added and removed a sentinel command on the production singleton and manually reset `--debug` in a `t.Cleanup`. It now adds the sentinel to a fresh isolated tree, so no mutation of shared state occurs and no cleanup is needed.

`requireGateDisabled` is removed: the workload/artifact gate tests now use `newIsolatedRootCmd()` with `t.Setenv("DATAROBOT_CLI_FEATURE_WORKLOAD", "")` to build a fresh tree with the gate off, so they pass regardless of the ambient environment.

## CHANGES

- Add `newIsolatedRootCmd()` helper (delegates to `NewIsolatedRootFactory`)
- Migrate all `Execute()`-calling tests to use it
- Remove `requireGateDisabled` and the `config`/`features`/`strings` imports that supported it
- Gate-presence tests use `newIsolatedRootCmd()` with `t.Setenv` so they pass even when `DATAROBOT_CLI_FEATURE_WORKLOAD` is set
- Restructure with section comments: execute tests / pure-cobra unit tests / structural singleton tests / animation test

---

> [!NOTE]
> <sup>[Cursor Bugbot](https://cursor.com/bugbot) is generating a summary for commit 12aa57413dfc6542a83f1c3a3248581802eba875. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/cmd/root_test.go
+++ b/cmd/root_test.go
@@ -31,34 +31,21 @@ import (
 // Test isolation helper
 // ---------------------------------------------------------------------------
 
-// newIsolatedRootCmd returns a fresh *cli.CommandAdder built by a new
-// RootFactory with all side-effecting dependencies replaced by safe no-ops.
-// Each call returns an independent command tree so that:
+// newIsolatedRootCmd returns a fresh *cli.CommandAdder built by
+// NewIsolatedRootFactory, which replaces every side-effecting dependency
+// with a safe no-op. Each call returns an independent command tree so that:
 //   - viperx global state set by one test cannot bleed into the next
+//   - global viper flag bindings are not re-pointed at the test tree
 //   - http.DefaultTransport is not mutated by TLS setup
 //   - Plugin discovery (filesystem + network) is skipped
 //   - The first-run animation is suppressed
+//   - No telemetry client is constructed or transmitted with
 //
 // Use this helper for any test that calls Execute() on the root command.
 // Tests that only inspect the static command tree (flags, subcommands,
 // annotations) may continue to reference the package-level RootCmd singleton.
 func newIsolatedRootCmd() *cli.CommandAdder {
-	return NewRootFactory(
-		// Skip viperx initialization and drconfig.yaml reads so env vars set
-		// by one test cannot bleed into the next via shared viper state.
-		WithConfigInitializer(func(_ *cobra.Command, _ string) error { return nil }),
-		// Skip http.DefaultTransport mutation to avoid TLS state leaking across tests.
-		WithTLSSetup(func(_ *cobra.Command) error { return nil }),
-		// Return a minimal props struct instead of hitting the network or disk
-		// for device/user IDs, account info, etc.
-		WithTelemetryProps(func() *telemetry.CommonProperties {
-			return &telemetry.CommonProperties{}
-		}),
-		// Skip plugin discovery (filesystem walk + subprocess invocations).
-		WithPluginRegistrar(func(_ *cobra.Command) {}),
-		// Suppress the first-run animation; there is no TTY in test processes.
-		WithAnimation(func() {}),
-	).Build()
+	return NewIsolatedRootFactory().Build()
 }
 
 // ---------------------------------------------------------------------------
--- a/cmd/root_test.go
+++ b/cmd/root_test.go
@@ -16,11 +16,9 @@ package cmd
 
 import (
 	"bytes"
-	"strings"
 	"testing"
 
-	"github.com/datarobot/cli/internal/config"
-	"github.com/datarobot/cli/internal/features"
+	"github.com/datarobot/cli/internal/cli"
 	"github.com/datarobot/cli/internal/misc/reader"
 	"github.com/datarobot/cli/internal/telemetry"
 	"github.com/datarobot/cli/internal/tools"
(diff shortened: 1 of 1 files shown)


cmd/root_test.go
```diff
@@ -338,16 +346,17 @@ func TestSetUnknownArgGuards_SkipsExplicitArgs(t *testing.T) {
...
-	// The feature gating happens during init(), so this tests the actual state.
-	cmd := RootCmd
+// ---------------------------------------------------------------------------
+// Structural tests against the production singleton (read-only, no Execute)
+// ---------------------------------------------------------------------------
 
+// TestWorkloadCommandNotPresentByDefault verifies that workload is absent from
+// the default command tree. The feature gate is evaluated during init() so we
+// inspect the production singleton; no execution is required.
+func TestWorkloadCommandNotPresentByDefault(t *testing.T) {
```
chasdr, line 357: these two fail when `DATAROBOT_CLI_FEATURE_WORKLOAD` is set (reproduced: red with it, green without) because `RootCmd`'s gate is baked at `init()`. since the factory can build a gate-off tree on demand, neutralize the env for the test and assert against a fresh isolated build — verified this passes even with the ambient var on:

```go
func TestWorkloadCommandNotPresentByDefault(t *testing.T) {
	t.Setenv("DATAROBOT_CLI_FEATURE_WORKLOAD", "")

	root := NewIsolatedRootFactory().Build()

	var found bool
	for _, sub := range root.Commands() {
		if sub.Name() == "workload" {
			found = true
			break
		}
	}

	assert.False(t, found, "workload command should not be present when feature gate is not enabled")
}
```

same idea for the artifact one right below it.

cmd/root_test.go
```diff
@@ -29,6 +27,31 @@ import (
...
+//   - global viper flag bindings are not re-pointed at the test tree
+//   - http.DefaultTransport is not mutated by TLS setup
+//   - Plugin discovery (filesystem + network) is skipped
+//   - The first-run animation is suppressed
+//   - No telemetry client is constructed or transmitted with
+//
+// Use this helper for any test that calls Execute() on the root command.
+// Tests that only inspect the static command tree (flags, subcommands,
+// annotations) may continue to reference the package-level RootCmd singleton.
+func newIsolatedRootCmd() *cli.CommandAdder {
```
chasdr, line 47: this helper is the whole PR in one function — every execute test getting its own no-op tree instead of fighting the shared singleton. 👍

chasdr (commented):
👋 chasbot here (chas's Claude), last one — the test-isolation PR, and my favorite of the stack.

moving the execute-path tests onto fresh `NewIsolatedRootFactory().Build()` trees and deleting the manual `--version`/`--debug` flag-reset cleanups is exactly right, and keeping the structural (no-execute) tests on the singleton is the correct split. the helper carries the whole PR.

one catch, and i think it's your droid's slip: dropping `requireGateDisabled` left the two "not present by default" tests doing a hard `assert.False` against the production singleton, whose gate state is frozen at `init()`. so anyone with `DATAROBOT_CLI_FEATURE_WORKLOAD` exported — i.e. anyone actually working on workload — gets a red suite. i reproduced it: both FAIL with the var set, pass without.

not a hard blocker (clean CI is green), but since you just built the exact tool that fixes it i couldn't resist. suggested the fix inline and verified it passes even with the ambient var on. happy to flip to approve once it's in.

leaving this as a comment rather than approve just so a known-red-under-a-real-env test doesn't get rubber-stamped. everything else here is 👍.

cmd/root_test.go
ajalon1: Fixed — both gate tests now use `newIsolatedRootCmd()` with `t.Setenv("DATAROBOT_CLI_FEATURE_WORKLOAD", "")` so they pass regardless of the ambient env var.

cmd/root_test.go
```diff
@@ -338,16 +346,22 @@ func TestSetUnknownArgGuards_SkipsExplicitArgs(t *testing.T) {
...
+// Structural tests (read-only, no Execute)
+// ---------------------------------------------------------------------------
+
+// TestWorkloadCommandNotPresentByDefault verifies that workload is absent from
+// the default command tree. Uses a fresh isolated build with the feature-gate
+// env var neutralized so the test passes even when
+// DATAROBOT_CLI_FEATURE_WORKLOAD is set in the ambient environment.
 func TestWorkloadCommandNotPresentByDefault(t *testing.T) {
-	requireGateDisabled(t, "workload")
+	t.Setenv("DATAROBOT_CLI_FEATURE_WORKLOAD", "")
```
chasdr, line 358: this is the fix, verified: passes with `DATAROBOT_CLI_FEATURE_WORKLOAD=true` in the env now (was red before). and reusing `newIsolatedRootCmd()` is cleaner than what i suggested. 👍

chasdr (approved):
👋 chasbot back for the re-review (chas's Claude). i re-pulled the whole stack after the rebase and re-ran everything, so here's the full picture, not just this PR:

**#811 (this one):** the gate fix is in and it works — `TestWorkloadCommandNotPresentByDefault` and `TestArtifactCommandNotPresentByDefault` now pass even with `DATAROBOT_CLI_FEATURE_WORKLOAD=true` exported, which is the exact case that went red before. bonus points for routing it through `newIsolatedRootCmd()` instead of hand-rolling the build, and for fixing the stale "during init()" comment to "during Build()". approving.

**#803:** the docs fence is fixed — the "Stamping universal interaction flags" section now sits after the closing fence and renders as a real section. my approval there still stands.

**#802 / #810:** pure rebases, zero content change from what i already approved. still good.

whole rebased stack builds clean and the affected tests are green (i ran them with the gate var set, worst case). that's the loop closed. 👍

chasdr:
👋 chasbot — the lone red (`Detect Changes`) is a 5m CI timeout, not your change. everything that ran is green, gate fix included. kick it and it goes green. 👍

PR #812, opened 2026-08-20
[CFX-7713] fix(task): name the real template markers in the not-in-template error

ajalon1:
Potentially a blocker before I release CLI 0.3.0!

# RATIONALE

The "not in a template directory" error tells users to create a folder they are already standing in.

#791 replaced `internal/task.IsTemplateDir` (a bare `os.Stat` on `.datarobot`) with `repo.IsTemplateDir`, which is a strictly narrower predicate: it wants `.datarobot/answers`, or a `.datarobot/cli` holding something other than `state.yaml`. That consolidation was the right call — the old check treated the stray `.datarobot/cli/state.yaml` that older CLI versions scattered outside template directories as proof of a template.

But the error text did not move with it. So the users most likely to hit the stricter gate — the ones whose `.datarobot` folder exists and is no longer sufficient — get told:

> This command requires a '.datarobot' folder to be present.

They have one. The message sends them looking for something they already did, and says nothing about the marker actually missing. It affects `dr task list`, `dr run`, and shell completion, since the gate runs first in `Discovery.Discover`, ahead of the `PreferRootTaskfile` fallback.

## CHANGES

- Build the `ErrNotInTemplate` message from the same constants the predicate reads, so it names the real markers and points at `dr template setup` instead of leaving the remedy to guesswork.
- Add `repo.TemplateDetectStateFileName`. The `state.yaml` exemption was a bare literal in `repo.IsTemplateDir`; hardcoding it a second time in the message would have rebuilt the same drift this PR is fixing, so predicate and message now read one constant.

## BEFORE / AFTER

Fixture: a directory containing a bare `.datarobot/` folder.

Before (current `main`):

```console
$ dr task list
You don't seem to be in a DataRobot Template directory.
This command requires a '.datarobot' folder to be present.
$ echo $?
1
```

After (this PR):

```console
$ dr task list
You don't seem to be in a DataRobot Template directory.
This command requires a '.datarobot/answers' folder, or a '.datarobot/cli' folder holding something other than 'state.yaml'.
Run 'dr template setup' to create one, or switch to an existing template directory.
$ echo $?
1
```

Checked against three fixtures — a bare `.datarobot/`, a `.datarobot/cli/` holding only `state.yaml`, and a bare `.datarobot/` alongside a committed root `Taskfile.yaml`. Output is identical between `main` and this PR on all three except the message text, which is the intended blast radius: no behavior change, same exit code, same gate.

## NOTE

There is a separate regression around `dr run`, discovered while testing here, but I've split that out and have moved it to CFX-7712.

## TESTING

Two tests in `internal/task/discovery_test.go` pin the text to the predicate:

- `TestFormatDiscoveryErrorNamesRealTemplateMarkers` — the message must name
  every marker `repo.IsTemplateDir` checks.
- `TestTemplateMarkersMatchAdvertisedRequirement` — the directory shapes the
  message advertises must pass the gate, and the two it excludes (a bare
  `.datarobot`, and a `.datarobot/cli` holding only `state.yaml`) must fail it.

Both fail against the previous message. Verified by temporarily restoring the old
text: it trips the `.datarobot/cli` and `state.yaml` assertions, then passes once
the fix is back.

`task lint` clean across all three GOOS. Full `task test` green, no races.

## NOTES

Related: #808 covers the other half of the release review that surfaced this (the
versioning guidance in `releasing.md`). Independent of this PR.

Worth a reviewer's judgment: naming `state.yaml` in user-facing copy leaks a
detail most people will not care about. I kept it because without it a user with
a `.datarobot/cli` folder has no way to understand why they still fail the check.
Happy to drop it if you would rather keep the message shorter.

---

> [!NOTE]
> **Low Risk**
> User-facing error copy and a shared filename constant only; template detection logic is unchanged.
> 
> **Overview**
> Fixes the **not-in-template** error so it matches `repo.IsTemplateDir` instead of telling users a bare `.datarobot` folder is enough.
> 
> `FormatDiscoveryError` now names `.datarobot/answers` or a `.datarobot/cli` folder with something other than `state.yaml`, and points users at `dr template setup`. The `state.yaml` exemption is a shared `TemplateDetectStateFileName` constant so the message cannot drift from the predicate. Tests pin the copy to those markers and to the directory shapes that pass vs fail the gate. Detection behavior is unchanged.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit fdcdada1011c2a58ef1fea41488c61ac6ff8b8e3. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/internal/task/discovery_test.go
+++ b/internal/task/discovery_test.go
@@ -19,6 +19,7 @@ import (
 	"path/filepath"
 	"testing"
 
+	"github.com/datarobot/cli/internal/repo"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/suite"
 )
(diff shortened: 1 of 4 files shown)


carsongee (approved):
I was worried that would have unintended consequences. I thought I had tracked all of them down. Good catch

c-h-russell-walker (approved):

internal/task/discovery_test.go
ajalon1, line 1: I think these tests are overkill, but hey! Tests are cheap these days, and we can always remove them or update them if template assumptions shift.

PR #814, opened 2026-08-21
[CFX-4730] Add inert plugin CLI version constraint predicate (PR 1/2)

ajalon1:
## Summary

PR 1 of 2 (stacked-to-main chain) for CFX-4730. Adds the inert, independently-testable predicate mechanism for plugin manifest CLI version constraints. Nothing in this PR changes plugin discovery behavior — the new predicate and fields exist but are not wired into any discovery call site yet. That wiring lands in PR 2.

## What changed

- `internal/plugin/types.go`: added `MaxCLIVersion` alongside the existing `MinCLIVersion` on `PluginManifest`; added a typed `PluginSkipReason` enum (`SkipReasonNameConflict` is the zero value, so every existing `PluginConflict{Name:…, Path:…}` literal keeps its exact meaning) plus `Reason`/`Detail` fields on `PluginConflict`.
- `internal/plugin/cli_version.go` (new): `compatibleCLIVersion` — a pure predicate comparing only the core (major.minor.patch) version of the running CLI against declared `minCLIVersion`/`maxCLIVersion` bounds, inclusive on both ends. A malformed declared bound always causes a skip, even when the running CLI version is `dev` or otherwise unparseable — the dev/unparseable-CLI bypass only applies once the declared bounds are confirmed well-formed. Also adds `cliVersionSkip` (builds the `PluginConflict` record) and a `currentCLIVersion` package-level seam so tests never silently no-op against the literal `"dev"` build value.
- `internal/plugin/validation.go`: comment-only update naming `MaxCLIVersion` alongside `MinCLIVersion` as fields intentionally excluded from cross-manifest comparison.
- `internal/plugin/cli_version_test.go` (new): table-driven coverage for every relevant spec scenario — no bounds, min-only, max-only inclusive boundary, both bounds, below min, above max, malformed bound (skip regardless of CLI version), dev/unparseable CLI with well-formed bounds (bypass), the joint dev+malformed case (malformed takes precedence), and prerelease CLI version (core-version comparison).

## Out of scope (PR 2)

`internal/plugin/discover.go` call sites, `cmd/plugin/discovery.go`, `LogConflicts`/`ConflictsForName` branching, and docs — this PR is self-contained and does not change plugin discovery behavior.

Ref: https://datarobot.atlassian.net/browse/CFX-4730

## Also

Thanks @yuriyhrytsyuk for doing the original work on this!

### DIFF
--- a/openspec/changes/cli-plugin-version-constraints/tasks.md
+++ b/openspec/changes/cli-plugin-version-constraints/tasks.md
@@ -1,0 +1,66 @@
+# Tasks: Plugin manifest CLI version constraints (CFX-4730)
+
+## Review Workload Forecast
+
+| Field | Value |
+|-------|-------|
+| Estimated changed lines | ~400-430 |
+| 400-line budget risk | High |
+| Chained PRs recommended | Yes |
+| Suggested split | PR 1 (mechanism, ~180 lines) → PR 2 (integration, ~230 lines) |
+| Delivery strategy | auto-chain |
+| Chain strategy | stacked-to-main |
+
+Decision needed before apply: No
+Chained PRs recommended: Yes
+Chain strategy: stacked-to-main
+400-line budget risk: High
+
+### Suggested Work Units
+
+| Unit | Goal | Likely PR | Focused test command | Runtime harness | Rollback boundary |
+|------|------|-----------|----------------------|-----------------|-------------------|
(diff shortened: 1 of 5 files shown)


openspec/changes/cli-plugin-version-constraints/tasks.md
ajalon1, line 1: This was the spec used to implement this; will remove before I merge, just thought it would be useful to share.

PR #815, opened 2026-08-21
[CFX-4730] feat(plugin): wire CLI version constraint checks into discovery

ajalon1:
## What

PR 2 of 2 (stacks on #814). Wires the inert `compatibleCLIVersion`/`cliVersionSkip` predicate from #814 into actual plugin discovery behavior.

## Why

CFX-4730: plugin manifests can declare `minCLIVersion`/`maxCLIVersion` bounds, but until now those bounds were parsed and validated (#814) without being enforced. This PR makes the CLI actually skip incompatible plugins at discovery time.

## Changes

- `internal/plugin/discover.go`: call `cliVersionSkip` inside `loadManagedPlugin` and inside `getManifestsParallel`'s merge/dedup loop, both strictly **before** the `seen[...]` name reservation — an incompatible plugin never blocks a compatible, identically-named plugin, and is reported as a version incompatibility, never a name conflict.
- `internal/plugin/discover.go`: added `ConflictsForReason`; `LogConflicts` now switches on `PluginConflict.Reason` — WARN for name conflicts (byte-identical wording to before), INFO for version incompatibility (includes the violated bound + running CLI version + upgrade guidance).
- `cmd/plugin/discovery.go`: extracted `reportDiscoveryConflicts` so routine command registration (`RegisterPluginCommands`, which runs on every `dr` invocation) only warns about name conflicts and stays silent about version-incompatibility skips. Those surface only via `dr plugin list` / `dr plugin version <name>`, which already pass the full, unfiltered conflict set through `LogConflicts` — verified this needs no code changes there via a real binary harness run (see Testing).
- Docs: documented `maxCLIVersion`/`minCLIVersion` semantics in `docs/development/plugins.md` and `docs/development/remote-plugins.md`.

## Testing

- `task lint` — clean, 0 issues on all 3 GOOS targets.
- `task test` — full repo suite green (race + coverage); `internal/plugin` package coverage 80.9%.
- New/extended unit and integration tests: managed plugin below `minCLIVersion` is skipped and not loaded; two PATH dirs sharing a manifest name where the lexicographically-first violates `maxCLIVersion` — the second still registers under that name, and the skip is reported as `SkipReasonVersionIncompatible`, never `SkipReasonNameConflict`; `LogConflicts` level-switch behavior; `ConflictsForReason` filtering; `reportDiscoveryConflicts` silence on routine discovery.
- Manual runtime harness: built the CLI with `-ldflags "-X .../version.Version=1.0.0"`, created a fake `dr-oldwidget` PATH plugin declaring `maxCLIVersion: 0.9.0`. Confirmed `dr plugin list` excludes it from the table and prints the INFO skip line; `dr plugin version oldwidget` prints the same line then reports the plugin as not found; plain `dr --help` prints neither the plugin name nor "version incompatible" (silent on ordinary discovery, per spec).

Refs CFX-4730. Stacks on #814.

---

> [!NOTE]
> **Medium Risk**
> Changes which plugins are loaded and registered on every CLI invocation. Incorrect skip ordering or logging could hide valid plugins or spam every `dr` command, but this is not auth, secrets, or data-handling.
> 
> **Overview**
> Enforces plugin `minCLIVersion`/`maxCLIVersion` at discovery so incompatible plugins never load. This is the second slice of CFX-4730: it wires the previously inert `cliVersionSkip` check into managed and PATH discovery **before** name reservation, so a version-skipped plugin cannot block a compatible plugin with the same name.
> 
> `LogConflicts` now Warns only for name collisions and Infos version skips (with bound + upgrade guidance). Routine command registration stays silent on version skips; they appear only via `dr plugin list` / `dr plugin version`. Docs cover inclusive bounds, the `dev`/unparseable-CLI bypass, and malformed-bound skips.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 5bbf0abfe00c07c6e17e216621ac11430998de93. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/openspec/changes/cli-plugin-version-constraints/tasks.md
+++ b/openspec/changes/cli-plugin-version-constraints/tasks.md
@@ -41,26 +41,32 @@ Chain strategy: stacked-to-main
 
 ## Phase 4: Discovery Wiring (PR 2, TDD)
 
-- [ ] 4.1 RED: in `internal/plugin/discover_test.go`, extend `createManagedTestPlugin` to accept min/max bounds; add case asserting an out-of-range managed plugin is absent from results and present in conflicts with `Reason: SkipReasonVersionIncompatible`. Spec: Managed plugin below minimum.
-- [ ] 4.2 RED: add `DiscoverWithContextSuite` case — two PATH dirs export the same manifest name, lexicographically-first violates `maxCLIVersion`; assert second registers under that name, conflicts hold exactly one `SkipReasonVersionIncompatible` record and zero `SkipReasonNameConflict` records. Spec: Skip Ordering Relative to Name Deduplication.
-- [ ] 4.3 RED: override `currentCLIVersion` (with `t.Cleanup` restore) in `TestDuplicatePATHEntryDoesNotWarn` and the XDG-config-dirs discovery test to a real semver so the `dev` bypass does not mask assertions.
-- [ ] 4.4 Confirm 4.1-4.3 fail against current code (`go test ./internal/plugin/... -race`).
-- [ ] 4.5 GREEN: call `cliVersionSkip` in `loadManagedPlugin` (~L304) and `getManifestsParallel` (~L441 loop), both immediately before the `seen[...]` reservation; append its non-nil result to conflicts and `continue` without setting `seen[name]`.
-- [ ] 4.6 Run `go test ./internal/plugin/... -race -v` until 4.1-4.3 are GREEN.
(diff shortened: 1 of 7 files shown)


openspec/changes/cli-plugin-version-constraints/tasks.md
ajalon1, line 1: again, will be deleted, just wanted to leave it here for reference

PR #820, opened 2026-08-21
[CFX-7754] Reject browser-forged API keys during dr auth login

chasdr:
## Summary

During `dr auth login` the CLI runs a local listener on `localhost:51164` for 5 minutes, waiting for the browser to hand back your API key. It accepted any request on that port, so a web page you had open could plant an attacker's key with `` and the CLI would store it as your credential, so every later `dr` command ran against the attacker's account. The callback now refuses the forgeable browser requests.

## Notes for review

The gate is `Sec-Fetch-Dest` present and not `document`. Absent must stay accepted: the CLI-to-CLI port handover uses a Go http client that sends no fetch metadata, and rejecting absent would deadlock two concurrent logins. It does not gate on `Sec-Fetch-Site`, since the real callback is legitimately cross-site.

Partial by design. It does not stop a page that forces a full-page navigation (a real `document` request, visible in the tab) or browsers older than Safari 16.4. The complete fix is a callback nonce, tracked separately since it needs a web app change.

Blocking manual gate, a real login must still work in the browser. DataRobot officially supports Chrome only, so Chrome is the gate:
- Chrome (macOS): pass
- Safari (macOS): pass (verified as an extra WebKit data point, not officially supported)
- Firefox, Edge: outside official support, not gating
- Chrome (Windows): to confirm post-release. Fetch metadata is per-engine not per-OS, so the same Chromium build sends identical `Sec-Fetch-*` headers on Windows. This confirms rather than gates.

## Output

```
$ curl -s -o /dev/null -w "%{http_code}\n" -H "Sec-Fetch-Dest: image" "http://localhost:51164/?key=PLANTED-BY-ATTACKER"
403
$ curl -s -o /dev/null -w "%{http_code}\n" -H "Sec-Fetch-Dest: document" "http://localhost:51164/?key=REAL-KEY"
200
```

## Technical Changes

- `browserflow.go`: refuse when `Sec-Fetch-Dest` is present and not `document`; success page only for a keyed request, 204 for the keyless handover sentinel.
- `browserflow_test.go`: image/empty refused, document accepted, headerless accepted, keyless image no longer interrupts, keyless headerless still hands over.
- auth docs: corrected the security note, added the editing rule for the gate.

### Breakdown

- code: +17 / -5
- tests: +103 / -3
- docs: +15 / -5

---

> [!NOTE]
> **High Risk**
> Touches the login callback that accepts and stores API keys. The gate is partial by design (no `state`/nonce), so mistakes here can still accept attacker credentials or break concurrent logins.
> 
> **Overview**
> Hardens `dr auth login` so a page cannot plant an API key via a forgeable request (`` / `fetch`) to the localhost callback while login is in flight.
> 
> `handleCallback` now returns **403** when `Sec-Fetch-Dest` is present and not `document`. An **absent** header is still accepted so CLI-to-CLI port handover (Go `http.Client`, no fetch metadata) does not deadlock. Keyless handover gets **204** instead of the success HTML; a keyless browser probe (e.g. favicon) is refused and no longer aborts login.
> 
> This is a partial mitigation: full-page navigation and clients that can set headers are still in scope. Docs and tests cover the gate, genuine `document` callbacks, and the handover sentinel.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 385eb4a150d2258b5720e9848bbf09f2f53fc53c. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/docs/commands/auth.md
+++ b/docs/commands/auth.md
@@ -519,10 +519,13 @@ Properties of this flow:
 - The callback listener is bound to localhost only
 
 > [!IMPORTANT]
-> The API key arrives as a URL query parameter and is stored in plaintext. There is no
-> `state` parameter, so any local process able to reach `localhost:51164` while a login
-> is in flight could deliver a key. Treat `drconfig.yaml` as a secret and prefer
-> `DATAROBOT_API_TOKEN` in shared or automated environments.
+> The API key arrives as a URL query parameter and is stored in plaintext. The callback
+> refuses browser requests that a page can forge (an `<img>` or background `fetch` carries
+> a `Sec-Fetch-Dest` other than `document`), so a web page you have open cannot plant a
+> key while a login is in flight. It still has no `state` parameter, so a local process
+> that can forge request headers, or a page that forces a full-page navigation, can deliver
+> one. Treat `drconfig.yaml` as a secret and prefer `DATAROBOT_API_TOKEN` in shared or
+> automated environments.
 
 ## Configuration file
 
--- a/docs/commands/auth.md
+++ b/docs/commands/auth.md
@@ -520,12 +520,15 @@ Properties of this flow:
 
 > [!IMPORTANT]
 > The API key arrives as a URL query parameter and is stored in plaintext. The callback
-> refuses browser requests that a page can forge (an `<img>` or background `fetch` carries
-> a `Sec-Fetch-Dest` other than `document`), so a web page you have open cannot plant a
-> key while a login is in flight. It still has no `state` parameter, so a local process
-> that can forge request headers, or a page that forces a full-page navigation, can deliver
-> one. Treat `drconfig.yaml` as a secret and prefer `DATAROBOT_API_TOKEN` in shared or
-> automated environments.
+> refuses the requests a page can make invisibly: a hidden `<img>` carries
+> `Sec-Fetch-Dest: image` and a background `fetch` or `XMLHttpRequest` carries `empty`, and
+> the listener turns both away, so a page you have open cannot silently plant a key while a
+> login is in flight. The gate accepts only `Sec-Fetch-Dest: document` and has no `state`
+> parameter, so it does not stop a real top-level navigation to the callback URL (`window.open`,
+> a link click, or setting `window.location`) or a local process that can forge headers. A
+> document navigation is at least visible to you, because the tab moves or a popup opens.
+> Treat `drconfig.yaml` as a secret and prefer `DATAROBOT_API_TOKEN` in shared or automated
+> environments.
 
 ## Configuration file
 
--- a/internal/auth/browserflow.go
+++ b/internal/auth/browserflow.go
@@ -176,15 +176,27 @@ func (f *BrowserFlow) Close() error {
 	return f.closeErr
 }
 
-// handleCallback receives the redirect from the DataRobot web app, which carries
-// the API key as the "key" query parameter.
+// handleCallback receives the API key from the web app's redirect ("key" query param).
+// Sec-Fetch-Dest present and not "document" refuses; absent accepts (port handover sends none).
 func (f *BrowserFlow) handleCallback(w http.ResponseWriter, r *http.Request) {
+	if dest := r.Header.Get("Sec-Fetch-Dest"); dest != "" && dest != "document" {
+		log.Debugf("Refusing auth callback with Sec-Fetch-Dest %q", dest)
+		http.Error(w, "forbidden", http.StatusForbidden)
+
+		return
+	}
+
 	apiKey := r.URL.Query().Get("key")
 
-	w.Header().Set("Content-Type", "text/html; charset=utf-8")
+	// The keyless interrupt sentinel is read by a Go client that ignores the body.
+	if apiKey == "" {
+		w.WriteHeader(http.StatusNoContent)
+	} else {
+		w.Header().Set("Content-Type", "text/html; charset=utf-8")
 
-	if err := assets.Write(w, "templates/success.html"); err != nil {
-		log.Debugf("Failed to render auth success page: %v", err)
+		if err := assets.Write(w, "templates/success.html"); err != nil {
+			log.Debugf("Failed to render auth success page: %v", err)
+		}
 	}
 
 	// Non-blocking: the channel holds one key and Wait may already have returned.
--- a/internal/auth/browserflow.go
+++ b/internal/auth/browserflow.go
@@ -177,11 +177,11 @@ func (f *BrowserFlow) Close() error {
 }
 
 // handleCallback receives the API key from the web app's redirect ("key" query param).
-// Sec-Fetch-Dest present and not "document" refuses; absent accepts (port handover sends none).
+// Sec-Fetch-Dest accepts "document" (real callback) and absent (port handover sends none); any other value refuses.
 func (f *BrowserFlow) handleCallback(w http.ResponseWriter, r *http.Request) {
 	if dest := r.Header.Get("Sec-Fetch-Dest"); dest != "" && dest != "document" {
 		log.Debugf("Refusing auth callback with Sec-Fetch-Dest %q", dest)
-		http.Error(w, "forbidden", http.StatusForbidden)
+		http.Error(w, "forbidden: request type not allowed", http.StatusForbidden)
 
 		return
 	}
--- a/internal/auth/browserflow_test.go
+++ b/internal/auth/browserflow_test.go
@@ -100,11 +100,104 @@ func TestBrowserFlow_EmptyKeyIsInterrupt(t *testing.T) {
 	}()
 
 	resp := waitForCallback(t, callbackURL(t, flow, ""))
+	body, readErr := io.ReadAll(resp.Body)
+	require.NoError(t, readErr)
 	require.NoError(t, resp.Body.Close())
+
+	assert.Equal(t, http.StatusNoContent, resp.StatusCode)
+	assert.Empty(t, body, "the sentinel response must not be the success page")
 
 	err := <-errCh
 	require.Error(t, err)
 	assert.ErrorIs(t, err, ErrLoginInterrupted)
+}
+
+func TestBrowserFlow_RefusesForgedBrowserCallbacks(t *testing.T) {
+	// An <img> carries Sec-Fetch-Dest "image" and a fetch()/XHR carries "empty";
+	// a page cannot forge these headers, so neither can plant a key (CFX-7754).
+	for _, dest := range []string{"image", "empty"} {
+		t.Run(dest, func(t *testing.T) {
+			flow := newTestFlow(t)
+
+			keyCh := make(chan string, 1)
+			errCh := make(chan error, 1)
+
+			go func() {
+				key, err := flow.Wait(context.Background())
+
+				keyCh <- key
+
+				errCh <- err
+			}()
+
+			forged := waitForCallback(t, callbackURL(t, flow, "?key=planted-by-attacker"),
+				"Sec-Fetch-Dest", dest)
+			require.NoError(t, forged.Body.Close())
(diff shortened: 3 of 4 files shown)


ajalon1:
I thought I had reviewed this already.

docs/commands/auth.md
```diff
@@ -519,10 +519,13 @@ Properties of this flow:
 - The callback listener is bound to localhost only
 
 > [!IMPORTANT]
-> The API key arrives as a URL query parameter and is stored in plaintext. There is no
-> `state` parameter, so any local process able to reach `localhost:51164` while a login
-> is in flight could deliver a key. Treat `drconfig.yaml` as a secret and prefer
-> `DATAROBOT_API_TOKEN` in shared or automated environments.
+> The API key arrives as a URL query parameter and is stored in plaintext. The callback
+> refuses browser requests that a page can forge (an `<img>` or background `fetch` carries
```
ajalon1: **[Medium] Lead claim is overstated and self-contradicted.** The first sentence says "a web page you have open cannot plant a key" but the very next sentence says "a page that forces a full-page navigation, can deliver one." A top-level navigation carries `Sec-Fetch-Dest: document`, which the gate explicitly accepts, so the bypass is real.

Suggest scoping the claim to the vectors actually blocked, e.g. "cannot plant a key via a hidden `` or background `fetch`." Also, the bypass list omits `<iframe>`, `<link rel="prefetch" as="document">`, and `window.open` — all set `Sec-Fetch-Dest: document` and bypass the gate. Worth enumerating so the gate's limits aren't overstated.

internal/auth/browserflow.go
```diff
@@ -176,15 +176,27 @@ func (f *BrowserFlow) Close() error {
 	return f.closeErr
 }
 
-// handleCallback receives the redirect from the DataRobot web app, which carries
-// the API key as the "key" query parameter.
+// handleCallback receives the API key from the web app's redirect ("key" query param).
```
ajalon1, line 179: **[Low] Doc comment omits the present-and-`document` accept case.** The comment covers "present and not document refuses; absent accepts" but is silent on the third branch — present *and* `document` is accepted (the genuine callback). A future editor could misread "present … refuses" as "any present value refuses."

Suggest appending: `present-and-"document" is the real callback; absent accepts (port handover sends none).`

internal/auth/browserflow.go
```diff
@@ -176,15 +176,27 @@ func (f *BrowserFlow) Close() error {
...
+// Sec-Fetch-Dest present and not "document" refuses; absent accepts (port handover sends none).
 func (f *BrowserFlow) handleCallback(w http.ResponseWriter, r *http.Request) {
+	if dest := r.Header.Get("Sec-Fetch-Dest"); dest != "" && dest != "document" {
+		log.Debugf("Refusing auth callback with Sec-Fetch-Dest %q", dest)
+		http.Error(w, "forbidden", http.StatusForbidden)
```
ajalon1: **[Low] `http.Error` body is a bit cryptic.** Per the repo's error-message rules (AGENTS.md: "specialize messages"; `.cursor/bugbot-errors.md`: "avoid cryptic error messages"), the bare `"forbidden"` is the kind of string those rules target. It's what a browser renders if the guard ever fires on a legitimate request (e.g. a browser quirk sending an unexpected dest).

Consider `"forbidden: request type not allowed"`. Minor since this is primarily an attack/blocked path.

internal/auth/browserflow_test.go
```diff
@@ -227,14 +320,21 @@ func TestBrowserFlow_ReclaimsPortFromStaleServer(t *testing.T) {
...
 	t.Helper()
 
 	deadline := time.Now().Add(5 * time.Second)
 
 	for {
-		resp, err := http.Get(url) //nolint:noctx,gosec // test-controlled localhost URL
+		req, reqErr := http.NewRequestWithContext(context.Background(), http.MethodGet, url, nil)
+		require.NoError(t, reqErr)
+
+		for i := 0; i+1 < len(headers); i += 2 {
```
ajalon1, line 404: **[Low] Odd-length `headers` is silently dropped.** If a future caller writes `waitForCallback(t, url, "Sec-Fetch-Dest")` (name without a value), `i+1 < len(headers)` is false on the first iteration, so the header is silently dropped and the request is sent with no `Sec-Fetch-Dest` — exercising the ungated path instead of the intended one, producing a false-green test.

Suggest adding at the top of the helper: `require.Equal(t, 0, len(headers)%2, "headers must be alternating name/value pairs")`.

internal/auth/browserflow_test.go
```diff
@@ -227,14 +320,21 @@ func TestBrowserFlow_ReclaimsPortFromStaleServer(t *testing.T) {
...
-func waitForCallback(t *testing.T, url string) *http.Response {
+// starts the server asynchronously. headers are alternating name/value pairs.
+func waitForCallback(t *testing.T, url string, headers ...string) *http.Response {
 	t.Helper()
 
 	deadline := time.Now().Add(5 * time.Second)
 
 	for {
-		resp, err := http.Get(url) //nolint:noctx,gosec // test-controlled localhost URL
+		req, reqErr := http.NewRequestWithContext(context.Background(), http.MethodGet, url, nil)
```
ajalon1: **[Low] `context.Background()` vs `t.Context()`.** The repo is on Go 1.26, so `t.Context()` (available since 1.24) is usable. `context.Background()` is never cancelled, so an in-flight `Do` isn't torn down on test timeout/failure. Bounded by the server's `ReadHeaderTimeout: 10s`, so not a correctness bug — just nicer to use `t.Context()` so requests are cancelled automatically.

internal/auth/browserflow_test.go
```diff
@@ -100,13 +100,106 @@ func TestBrowserFlow_EmptyKeyIsInterrupt(t *testing.T) {
...
+	assert.Empty(t, body, "the sentinel response must not be the success page")
+
 	err := <-errCh
 	require.Error(t, err)
 	assert.ErrorIs(t, err, ErrLoginInterrupted)
 }
 
+func TestBrowserFlow_RefusesForgedBrowserCallbacks(t *testing.T) {
+	// An <img> carries Sec-Fetch-Dest "image" and a fetch()/XHR carries "empty";
+	// a page cannot forge these headers, so neither can plant a key (CFX-7754).
```
ajalon1, line 118: **[Medium] Tests are fully sequential — the real attack is concurrent.** Both `RefusesForgedBrowserCallbacks` and `KeylessBrowserProbeDoesNotInterrupt` issue the forged request, wait for its response, then issue the genuine one. The real threat model is a page firing many ``/`fetch` probes *concurrently* with the legitimate top-level navigation, where ordering isn't guaranteed.

The non-blocking `select` on the buffered-1 `keyCh` (`browserflow.go` ~line 209) isn't exercised under interleaved contention, which is where a subtle ordering bug (e.g., a 403 path that accidentally touched the channel) would hide.

Suggest a subtest that fires N forged probes concurrently (e.g. 10 goroutines with `Sec-Fetch-Dest: image`), waits for all to return 403, then sends the genuine callback and asserts `<-keyCh == "real-key"` and `<-errCh` is nil.

ajalon1 (approved):
Reviewed from four angles (security/STRIDE, correctness & concurrency, test quality, docs & code style) using parallel review passes against `upstream/main`.

**Verdict: approve.** The `Sec-Fetch-Dest` gate is a correct, low-risk defense-in-depth improvement and the 204 sentinel change fixes a real latent bug (favicon probes no longer abort an in-flight login). The handler is HTTP-correct (header ordering, 204 with no body, 403 early-return before the channel send), concurrency-safe (buffered-1 `keyCh` with non-blocking `select`, `closeOnce` on `Close`), and the tests pass clean under `-race -count 5` with `golangci-lint` reporting 0 issues.

Inline comments below cover the actionable findings (2 Medium, 4 Low). A few additional optional items that didn't warrant inline comments:

- **Test coverage (optional):** `Sec-Fetch-Dest: "document"` with an empty key (the intersection of the two new behaviors) is untested — expected 204 + `ErrLoginInterrupted`. Looping the forged request 2-3x before the genuine callback would catch per-request state leaks. Adding `"script"` to the forged-dest table documents the gate's breadth.
- **Minor UX (informational):** A user manually navigating to `localhost:51164/` with no `?key=` now gets a blank 204 instead of the success page. The interrupt behavior is unchanged; only the rendered body differs. Could render a small "this URL is for the CLI login" page for that case if desired.
- **Docs (informational):** The note could parenthetically acknowledge browsers that don't send Fetch Metadata headers aren't protected, though all modern browsers do.

None of the above are blocking. Nice fix.

chasdr:
👋 chasbot here — chas handed me your review and told me to actually read it this time. all six addressed in `05ed80c`:

- **docs overstated claim** (Medium): scoped it. the "cannot plant a key" line now covers only the invisible vectors (hidden ``, background `fetch`/XHR); the residual bypass is described as what it actually is, a real `Sec-Fetch-Dest: document` top-level navigation that the user sees because the tab moves or a popup opens. one back at you on the enumeration though: `<iframe>` sends `Sec-Fetch-Dest: iframe`, not `document`, so the gate already refuses it. `window.open` and `<link rel=prefetch as=document>` are the genuine `document` bypasses, so I named the category instead of listing elements.
- **sequential vs concurrent** (Medium): added `TestBrowserFlow_RefusesConcurrentForgedProbes` — 10 forged `image` probes fired concurrently, every one must 403, then the genuine callback lands and `keyCh` yields `real-key`. exercises the non-blocking `select` on the buffered `keyCh` under contention, which the sequential tests never touched.
- **doc comment third branch**: now names all three (document + absent accept, anything else refuses).
- **cryptic 403 body**: `"forbidden: request type not allowed"`.
- **odd-length headers**: `require.Equal(t, 0, len(headers)%2, ...)` guard at the top of the helper.
- **context**: `t.Context()` in `waitForCallback`.

green under `-race`. your droid earned its keep on this one.

docs/commands/auth.md
chasdr: scoped the claim to the invisible vectors and reframed the residual bypass as a `document` navigation in 05ed80c. dropped the element list on purpose: `<iframe>` carries `Sec-Fetch-Dest: iframe`, so the gate refuses it already. `window.open` and `<link rel=prefetch as=document>` are the real `document`-dest bypasses, so I named the category rather than a list that has a refused entry in it. shout if you'd still rather enumerate.

internal/auth/browserflow.go
chasdr: done in 05ed80c. comment now names all three branches: `document` and absent accept, anything else refuses.

internal/auth/browserflow.go
chasdr: `"forbidden: request type not allowed"` in 05ed80c.

internal/auth/browserflow_test.go
chasdr: added `require.Equal(t, 0, len(headers)%2, "headers must be alternating name/value pairs")` at the top of the helper in 05ed80c.

internal/auth/browserflow_test.go
chasdr: switched to `t.Context()` in 05ed80c.

internal/auth/browserflow_test.go
chasdr: added `TestBrowserFlow_RefusesConcurrentForgedProbes` in 05ed80c: 10 forged `image` probes fired concurrently, all must 403, then the genuine callback lands and `keyCh` == `real-key` with a nil err. the non-blocking `select` on the buffered `keyCh` is now exercised under contention.

chasdr:
@ajalon1 you want to scan this again - or am i gtg ?

ajalon1:
send it, still ✅

ajalon1 (approved):

PR #821, opened 2026-08-24
[RAPTOR-19599] Rename .wapiignore to .drignore

wojtekwdr:
# RATIONALE

`wapi` is an internal codename, and `.wapiignore` at the project root was the last user-visible place it reached. Removing the workload feature gate makes that filename public, and after that a rename costs a deprecation cycle instead of a fallback branch.

## CHANGES

`dr artifact code init` now writes `.drignore`; sync reads it and falls back to a committed `.wapiignore` when none sits beside it, saying once per run that the old name is deprecated, so projects configured before this change keep filtering exactly as they did. When both files are present `.drignore` wins and sync warns that the other file's patterns are inert, which is a state a sync can produce unaided since the ignore file is itself uploaded. The filenames now live in one place, so the package that seeds a starter file and the package that reads one cannot drift apart, and the `uv.lock` hint names whichever file the project actually has.

One decision is worth a reviewer's eye: the ignore file stays part of the synced set rather than joining `.gitignore` in `systemExcludes`, which keeps this a pure rename but means a user who follows the deprecation advice propagates the rename through the catalog. Happy to close that separately if you would rather it went in here.

## PR Automation

> [!IMPORTANT]
>

---

> [!NOTE]
> **Medium Risk**
> Changes which ignore file filters the synced tree, including dual-file precedence that can drop legacy patterns. A mistake here can upload secrets or skip intended excludes, though fallback and warnings are covered by tests.
> 
> **Overview**
> Renames the user-authored sync ignore file from `.wapiignore` to `.drignore`. `init` now seeds `.drignore` and will not write a new file if either name already exists.
> 
> Sync prefers `.drignore`, still reads a lone `.wapiignore`, and prints a one-line deprecation notice on stderr (including JSON mode and `up --dry-run`). If both files are present, only `.drignore` applies and a warning is logged that the legacy patterns are inert. Filenames live in the `ignore` package (`FileName`, `Locate`) so writers and readers cannot drift; `uv.lock` exclusion hints name the file actually in use.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit da7adcee51b9c774f1ae141c43bc7c81302b7180. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


internal/workload/up/run.go
```diff
@@ -625,5 +644,9 @@ func defaultCodeChange(loaded Loaded, _ Live) (change CodeChange, err error) {
 		return CodeChange{}, fmt.Errorf("cannot compare the working tree with the last deploy: %w", err)
 	}
 
-	return CodeChange{Applies: true, Files: len(plan.Uploads) + len(plan.Deletes)}, nil
+	return CodeChange{
+		Applies:      true,
+		Files:        len(plan.Uploads) + len(plan.Deletes),
+		IgnoreNotice: engine.IgnoreFileNotice(),
```
adamalpi: **[Nit] ShadowWarning logs twice on a real deploy**

A non-dry-run `up` that deploys code runs phase 2 twice — once in this sizing engine and once in `defaultSync`'s engine (`build.go`) — so the `log.Warn(ShadowWarning)` prints twice per deploy. This mirrors the pre-existing double-print of the lockfile warning, so it's consistent rather than new, but this PR adds a second message to the duplicated set. A once-per-process guard (or surfacing the warning from only one of the two engines) would fix both. Fine to defer.

internal/workload/up/run.go
```diff
@@ -183,6 +186,22 @@ func Run(opts Options) (Result, error) {
...
 
+// noteIgnoreFile passes on the sync engine's note about a deprecated ignore
+// filename. Sizing the working tree is the whole of what a --dry-run does and
+// it is where the engine reads the file, so this is the only point on that
+// path where there is anything to say.
+//
+// Only the housekeeping note comes through here. The ignore-file problems that
+// cost the user something are logged by the phase that finds them, so they
+// survive a run that fails before reaching this line.
+func noteIgnoreFile(code CodeChange, opts Options) {
```
adamalpi, line 197: **[Nit] First `up` on an unlinked legacy project misses the notice**

When the project isn't linked yet, `defaultCodeChange` returns `FirstDeploy` before any engine exists, so a project with a committed `.wapiignore` gets no deprecation notice on that first run (the patterns still apply via `defaultSync`, and every subsequent run shows the notice). Only delays the signal by one run — noting it so the gap is a decision rather than an accident.

internal/workload/ignore/matcher.go
```diff
@@ -46,24 +58,150 @@ var systemExcludes = []string{".datarobot/workload", ".wapi", ".git", ".gitignor
...
+//
+// Printing it is the caller's job, which keeps this package doing no I/O
+// beyond finding the ignore file. Same for ShadowWarning. The two are never
+// both set: the legacy name is only ever in effect when it is the only one.
+func (m *Matcher) Notice() string {
+	if m.source != LegacyFileName {
+		return ""
+	}
+
+	return fmt.Sprintf("Using %s for sync ignore patterns. That name is deprecated: rename the file to %s.",
```
adamalpi: **[Nit] The notice advises the rename without the coordination caveat**

The notice says "rename the file to .drignore" with no hint that the ignore file is part of the synced set, so the rename propagates to collaborators (and anyone syncing with an older CLI binary reads only `.wapiignore`). The docs cover this; the notice — the thing users actually see — doesn't. One clause like "coordinate with collaborators" would close the gap.

adamalpi (approved):
Careful, well-tested rename with the right precedence call (`.drignore` wins outright, warning instead of merging). I verified the sharp edges hold: `.wapiignore` is not swallowed by the `.wapi` system exclude, the ShadowWarning goes to stderr so JSON purity is intact, `Locate`/`New` agree on directories and dangling symlinks, and no stray `.wapiignore` references remain anywhere in the repo. Full workload/artifact test suites pass locally on the branch. A few nits inline, none blocking.

internal/workload/ignore/matcher.go
wojtekwdr: fixed

internal/workload/up/run.go
wojtekwdr: fixed

internal/workload/up/run.go
```diff
@@ -625,5 +644,9 @@ func defaultCodeChange(loaded Loaded, _ Live) (change CodeChange, err error) {
 		return CodeChange{}, fmt.Errorf("cannot compare the working tree with the last deploy: %w", err)
 	}
 
-	return CodeChange{Applies: true, Files: len(plan.Uploads) + len(plan.Deletes)}, nil
```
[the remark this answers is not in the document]
wojtekwdr: fixed

PR #817, opened 2026-08-21
[RAPTOR-19604] fix(fsutil): write through symlinks in AtomicWriteFile instead of replacing them

ajalon1:
# RATIONALE

`AtomicWriteFile` wrote a temp file beside the target and renamed it into place. `os.Rename` replaces a symlink itself rather than following it, so every write to a symlinked manifest silently destroyed the link and left the real file untouched — while reporting success.

This matters beyond edge cases: `dr workload up` writes the `workloadId` binding back through the same helper on every first deploy, so a symlinked `.datarobot.yaml` (e.g. a monorepo sharing one manifest across several projects) is silently detached on the very first deploy.

### `982f327` — Write through symlinks instead of replacing them

**`internal/fsutil/atomicwrite.go`**
- At the top of `AtomicWriteFile`, call `filepath.EvalSymlinks(path)`. When the path is a symlink (or a chain of them) and resolves successfully, swap in the resolved path so the temp file is created beside the real file and renamed over it. When the path does not exist or is dangling the error is silently ignored and behaviour is unchanged (new-file path).
- No caller changes are needed; the fix is entirely inside `fsutil`.

**`internal/fsutil/atomicwrite_test.go`**
- Add `TestAtomicWriteFile_WriteThroughSymlink`

### `7b8a501` — Name the caller's symlink path in errors

**`internal/fsutil/atomicwrite.go`**
- Move symlink resolution into `resolveWriteTarget`, which returns both the resolved path and a `viaSymlink` flag. The flag records whether the caller's path itself was a symlink (vs. symlinked intermediate directories like `/tmp` on macOS, which `EvalSymlinks` also resolves but need no annotation).
- Wrap failures after resolution with `write through symlink <callerPath>: ...` so a caller who passed a symlink sees the path they know, not a resolved target they may never have seen (dotfiles setups, macOS `/private/tmp` resolution).
- `AtomicWriteFile` now captures `callerPath` before resolution and applies the wrapper in a deferred function.

**`internal/fsutil/atomicwrite_test.go`**
- Add `TestAtomicWriteFile_WriteThroughRelativeSymlink`
- Add `TestAtomicWriteFile_DanglingSymlinkIsReplaced`
- Add `TestAtomicWriteFile_ErrorNamesTheSymlinkPath`

### `edc6a95` — Set viaSymlink on resolve failure + pin symlink chain behaviors

**`internal/fsutil/atomicwrite.go`**
[...]

PR #822, opened 2026-08-24
[RAPTOR-19526] fix(workload): clear the stale workloadId on delete

wojtekwdr:
# RATIONALE

`dr workload delete` removed the workload on the server and left behind the `workloadId` it had written into `.datarobot.yaml`. The next `dr workload up` was bound to something that no longer existed, and the only way out was hand-editing a file marked `# managed by the CLI`.

It bites on the recovery path, which is the worst place for it: the realistic sequence is a first deploy going wrong, deleting the broken workload to start clean, and then finding the CLI's own state stale.

Two things in the ticket did not survive contact with the code, and they changed what got built. `up` did not silently target the dead id, it errored and named two remedies. But one of them, `dr workload config --workload-id <id>`, is refused whenever a manifest exists, so the error pointed at a command that will not act.

# CHANGES

**`dr workload delete` clears a binding it can prove is its own.** The id is compared inside the same parse that removes it, so a concurrent `up` cannot have a fresh binding deleted by a check that saw a different value. It gained `--dir`, without which it never finds the manifest of a project deployed with `up --dir site` and deleted from the repository root, which is the reported reproduction. A 404 deliberately leaves the file alone: it means the workload is not on *this* instance, which is not proof the binding is stale.

**`up` treats a binding that resolves to nothing as drift and creates.** This reverses the rule recorded in `state.go` and in RAPTOR-18979's live-state table, so it is the part most worth a reviewer's attention; 18979 has been amended to match. That rule feared leaving two workloads of one name, but a 404 means the first is gone. The hazard that is real is a 404 meaning "not on this instance", so the plan line and the JSON envelope both name the id that went missing, which is what lets CI tell a wrong-org run from a legitimate first deploy.

**A terminated workload is still refused.** It continues to exist and to hold its name and its artifact, so a create collides with it. Folding it in with "missing" also made `up --lock` silently skip locking on a one-way operation.

**`up` mints a new artifact when the linked one no longer matches the file.** An artifact's spec is fixed at creation, so reusing a diverged one deployed the frozen answer and reported success. A read that fails is not treated as a difference, so a timeout never starts a new artifact lineage.

**The shared manifest edit layer stops losing things the user owns:** comments above and below the binding, and the line endings of a CRLF checkout. It refuses, with the reason and a remedy, on four shapes it cannot make sound again: an aliased binding, disagreeing repeated keys, a file whose only recognized key is the binding, and a multi-document file.

## NOTES

Larger than the usual budget at ~1.8k lines. Roughly half is the fixes from a `/code-review max` pass that found 15 issues, and unpicking those from the code they fix would not leave two reviewable commits.

`.datarobot/workload/` is deliberately left alone. It binds the project to an *artifact*, not to a workload, and the artifact survives the deletion. `delete` names it rather than removing it.

## RELATED

- RAPTOR-18979 — live-state row amended for the reversal above
- RAPTOR-19595 — the state directory has no lifecycle, and `rm -rf` is the documented answer in three places
- RAPTOR-19602 — a refused deploy still prints a plan and a `Next:` block for work it will not do
- RAPTOR-19604 — `fsutil.AtomicWriteFile` replaces a symlink instead of writing through it

---

> [!NOTE]
> **High Risk**
> Reverses `up` so a 404 binding becomes an unconfirmed create, which can deploy to the wrong instance/org; it also rewrites user-owned YAML. Terminated workloads stay refused and the plan/JSON name the missing id.
> 
> **Overview**
> **`dr workload delete` takes back the `workloadId` it wrote** when the located `.datarobot.yaml` names that exact id, so the next `up` is not bound to a gone workload. Adds `--dir` so a project deployed from a subdirectory is found. A 404 does not rewrite the file. The `.datarobot/` artifact link is left in place and named on stderr.
> 
> **`up` now treats a missing binding as drift and creates**, instead of refusing and asking for a hand edit. The plan and JSON `priorWorkloadId` name the id that 404’d so a wrong-instance run is visible. Terminated workloads still exist and are refused, with a remedy that points at `delete` (including `--dir` when needed).
> 
> **Linked artifacts are reused only while their spec still matches the file.** A mismatch mints a new artifact and relinks without wiping the sync baseline; a failed read is not treated as a mismatch, so a timeout does not start a new lineage.
> 
> Manifest edits share a safer read-modify-write: comments and CRLF are preserved; aliased bindings, disagreeing duplicate keys, emptying the file, and multi-document YAML are refused rather than rewritten.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 4c2dcd4ded2af24f9e39418893f6637ca2bbbbf4. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/cmd/workload/del/cmd.go
+++ b/cmd/workload/del/cmd.go
@@ -20,15 +20,20 @@ package del
 import (
 	"errors"
 	"fmt"
+	"io"
 	"net/http"
+	"path/filepath"
 
 	"github.com/datarobot/cli/cmd/helpers"
 	"github.com/datarobot/cli/internal/auth"
 	"github.com/datarobot/cli/internal/config/viperx"
 	"github.com/datarobot/cli/internal/drapi"
+	"github.com/datarobot/cli/internal/fsutil"
 	"github.com/datarobot/cli/internal/misc/reader"
 	"github.com/datarobot/cli/internal/telemetry"
 	"github.com/datarobot/cli/internal/workload"
+	"github.com/datarobot/cli/internal/workload/manifest"
+	"github.com/datarobot/cli/internal/workload/wapi"
 	"github.com/datarobot/cli/tui"
 	"github.com/spf13/cobra"
 )
@@ -44,6 +49,13 @@ replicas first, then removes the workload. The artifact it was created from
 is not deleted with it; remove that separately with
 'dr artifact delete <artifact-id>' once no workload references it.
 
+If the .datarobot.yaml found from --dir (the current directory by default,
+searched upward from there) is bound to the workload being deleted, its
+workloadId is removed too, so the next 'dr workload up' creates a new
+workload rather than pointing at one that is gone. Only a manifest naming
+that exact id is touched. Pass the same --dir you deployed with: a manifest
+in a subdirectory is not visible from its parent.
+
 Without --yes the command asks for confirmation.
 
 Example:
@@ -58,17 +70,35 @@ Example:
 				return err
 			}
 
+			// Not discarded: a lookup failure means the flag was renamed or
+			// dropped, and the silent fallback is a search from the working
+			// directory, which is the behaviour --dir exists to replace.
+			dir, err := cmd.Flags().GetString("dir")
+			if err != nil {
+				return fmt.Errorf("cannot read --dir: %w", err)
+			}
+
+			// Checked before the delete, not after: a typo is worth catching
+			// while it still costs nothing, and the cleanup that reports it
+			// runs on the far side of an irreversible call.
+			if dir != "" && !fsutil.DirExists(dir) {
+				return fmt.Errorf("--dir %s: %w", dir, manifest.ErrNotADirectory)
+			}
+
 			if err := workload.DeleteWorkload(args[0]); err != nil {
 				return handleDeleteError(err, args[0])
 			}
 
 			fmt.Println(tui.BaseTextStyle.Render("Deleted workload: " + args[0]))
 
+			clearStaleBinding(cmd.ErrOrStderr(), dir, args[0])
+
 			return nil
 		},
 	}
 
 	cmd.Flags().BoolP("yes", "y", false, "Skip the confirmation prompt.")
+	cmd.Flags().String("dir", "", "Project directory whose manifest holds the binding, searched upward from here.")
 
 	// Bind only the env var (DATAROBOT_CLI_NON_INTERACTIVE) to viper. The --yes
 	// flag itself is read directly from cmd.Flags() so an explicit --yes does
@@ -117,6 +147,15 @@ func confirmDelete(cmd *cobra.Command, workloadID string) (bool, error) {
 // handleDeleteError converts a 404 into a friendly informational message
 // (returns nil) so the user does not see a stack-trace-style HTTP error
 // for what is effectively a no-op.
+//
+// The binding is deliberately left alone here. A 404 says the workload is not
+// on this instance, which is not the same as saying it is gone: the same
+// manifest read against the wrong endpoint, organisation or token answers
+// exactly this way. `up` treats that ambiguity by naming the id and letting
+// the user look; rewriting a committed file on the strength of it, in the one
+// branch where nothing was actually deleted, would be the stronger claim made
+// on the weaker evidence. A binding that really is dead is cleared by the
+// deploy that recreates it.
 func handleDeleteError(err error, workloadID string) error {
 	var httpErr *drapi.HTTPError
 
@@ -127,4 +166,106 @@ func handleDeleteError(err error, workloadID string) error {
 	}
 
 	return err
+}
+
+// clearStaleBinding takes back the workloadId the CLI wrote into a manifest,
+// now that the workload it names has been deleted.
+//
+// The id is matched inside the edit itself, so this only ever touches a file
+// that is demonstrably about the workload just deleted, and the value it
+// checks is the value it removes. A manifest naming some other workload is
+// none of its business.
+//
+// dir is where to start looking, and it exists because Locate only walks
+// upward. A project deployed with `up --dir site` and then deleted from the
+// repository root is invisible from there, which is the reported bug's own
+// reproduction; --dir is how the two commands agree on which project this is.
+// A path that is not a directory is a typo worth naming, because Locate would
+// otherwise walk past it to an ancestor and quietly edit the wrong project.
+//
+// Nothing here can fail the command. The workload is already gone by the time
+// this runs, so a manifest that cannot be found, read or written is stepped
+// over rather than turned into a failure for an operation that succeeded. The
+// unwritable case still says so, because the user has to finish it by hand.
+func clearStaleBinding(w io.Writer, dir, workloadID string) {
+	if dir == "" {
+		dir = "."
+	}
+
+	path, err := manifest.Locate(dir)
+	if err != nil {
+		// Locate is the one place that decides what counts as a directory, so
+		// its answer is used rather than re-derived here. Reaching this needs
+		// the directory to have gone away between the check that runs before
+		// the delete and this line, which is rare and still worth a word: the
+		// binding was not looked at, so it may well still be there.
+		if errors.Is(err, manifest.ErrNotADirectory) {
+			fmt.Fprintln(w, tui.DimStyle.Render("No manifest was checked: "+err.Error()+"."))
+		}
+
+		return
+	}
+
+	cleared, err := manifest.ClearWorkloadID(path, workloadID)
+
+	// A file that cannot be parsed cannot be checked, so there is nothing to
+	// say: it may not be this project's manifest at all, and `up` reports an
+	// unreadable one in its own words. A failure past that point is on a file
+	// whose binding was matched, so it is this command's business to report.
+	if errors.Is(err, manifest.ErrUnreadable) {
+		return
+	}
+
+	// The manifest package's refusals each carry the remedy that fits them,
+	// and they differ: one says delete the file, another says edit it by hand.
+	// Appending a single generic "remove that line" walked the user into the
+	// empty manifest the only-key refusal exists to prevent.
+	if err != nil {
+		fmt.Fprintln(w, tui.DimStyle.Render("Could not remove workloadId from "+path+": "+err.Error()))
+
+		return
+	}
+
+	if !cleared {
+		return
+	}
+
+	fmt.Fprintln(w, tui.DimStyle.Render(
+		"Removed workloadId from "+path+"; the next 'dr workload up"+manifest.DirFlag(dir)+"' creates a new workload."))
+
+	noteLinkedArtifact(w, filepath.Dir(path))
+}
+
+// noteLinkedArtifact names the artifact this project is linked to, and how to
+// stop being linked to it.
+//
+// Deleting a workload leaves its artifact alone, by design and as this
+// command's help promises, so the link under .datarobot/ is still accurate and
+// is deliberately left in place. It is only surprising when it is invisible,
+// which is what this line fixes.
+//
+// It states that the artifact survived and stops there. Saying the next deploy
+// reuses it would be false twice over: a locked artifact makes the next deploy
+// refuse instead, and a manifest naming a published image or an artifactId
+// never consults the link at all.
+//
+// The remedy is the state directory, not `dr artifact delete`. Deleting the
+// artifact is refused outright while it is locked, and for an unlocked one it
+// leaves the link pointing at something gone, which the next deploy reports as
+// a bare 404 naming no fix. Removing the directory is what `up` itself already
+// tells the user to do when a locked artifact blocks a deploy, so this says
+// the same thing rather than inventing a second answer.
+func noteLinkedArtifact(w io.Writer, projectDir string) {
+	if !wapi.Exists(projectDir) {
+		return
+	}
+
+	cfg, err := wapi.LoadConfig(projectDir)
+	if err != nil || cfg.ArtifactID == "" {
+		return
+	}
+
+	fmt.Fprintln(w, tui.DimStyle.Render(
+		"This project is still linked to artifact "+cfg.ArtifactID+", which was not deleted with the workload. "+
+			"To unlink it, delete "+wapi.Dir(projectDir)+"."))
 }
--- a/internal/workload/manifest/write.go
+++ b/internal/workload/manifest/write.go
@@ -18,6 +18,7 @@ import (
 	"bytes"
 	"errors"
 	"fmt"
+	"io"
 	"math"
 	"os"
 	"path/filepath"
@@ -110,6 +111,13 @@ const (
 // atomicWrite is the durable write, a variable so a test can make it fail and
 // check that the reservation is taken back.
 var atomicWrite = fsutil.AtomicWriteFile
+
+// ErrUnreadable reports that a manifest could not be read or parsed at all, as
+// distinct from an edit that failed after the file was understood. A caller
+// editing a file it merely found, rather than one it was handed, needs to tell
+// those apart: an unparseable file may not be about its business, while a
+// write that fails on a file whose contents it matched certainly is.
+var ErrUnreadable = errors.New("manifest could not be read")
 
 // ErrExists reports that a manifest is already present at the target path.
 // Setup never rewrites one: the file is the interface after the first write,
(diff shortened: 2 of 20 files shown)


ajalon1:
One thing up front -- I think this will hit symlink problems as well, I have https://github.com/datarobot-oss/cli/pull/817 open for that

ajalon1 (commented):
## Review summary

Ran a full review of this branch. One confirmed correctness bug (see inline comment on `build.go:216`) plus two related correctness gaps in how the plan preview and the create message stay in sync with the new relink/reuse logic. The rest are consistency/maintainability nits.

Leaving this as **COMMENT** rather than approving because of the confirmed bug in `relinkProject`.

One more finding that doesn't land as an inline comment because the file it's in isn't touched by this diff:

**`internal/workload/up/load.go` — `Load()` doesn't special-case the new `manifest.ErrNotADirectory`**
This diff adds `manifest.ErrNotADirectory` to `manifest.Locate` (`internal/workload/manifest/manifest.go:39`) and gives `dr workload delete` a friendly message for it (`cmd/workload/del/cmd.go:84`: `"--dir %s is not a directory"`). But `up.Load()` (`internal/workload/up/load.go:128-139`) only special-cases `manifest.ErrNotFound` and falls through to the raw, unwrapped error for anything else — so `dr workload up --dir <typo>` still surfaces a bare `not a directory: <path>`-style error while the sibling `delete` command (touched by this same diff) gives a clear, actionable one. Worth the same treatment for consistency, even though it's a pre-existing file this PR didn't need to touch.

internal/workload/up/build.go
```diff
@@ -180,6 +194,64 @@ func ensureArtifact(loaded Loaded, report *reporter) (string, bool, error) {
...
+	}
+
+	// The catalog and the last synced version belong to the artifact being
+	// left behind, so they are cleared: keeping them would have the next sync
+	// diff this project's tree against a baseline from another lineage.
+	cfg.ArtifactID = artifactID
+	cfg.CatalogID = nil
+	cfg.LastSyncedVersionID = nil
+
+	if err := saveProjectFn(projectDir, cfg); err != nil {
```
ajalon1: **Confirmed bug**: `relinkProject` performs two separate, non-atomic writes with no rollback between them — `saveProjectFn` (here, line 216) links the project to the new artifact, then `resetBaselineFn` (line 225) resets the sync baseline as a second, independent step.

If `saveProjectFn` succeeds but `resetBaselineFn` then fails (disk full, killed process, permission race), `config.json` is already pointing at the new artifact — but the error `ensureArtifact` wraps this in (`build.go:188-191`) says *"artifact %s was created but %s could not be linked to it, so the next deploy would create another"*. That's wrong for this specific partial-failure path: the link **was** saved, so the next run reuses this already-relinked project (via `reusableArtifact`) with the **old** sync baseline, which claims every file is already uploaded. Nothing gets pushed to the new artifact, and the build fails with an opaque "nothing to build" — exactly the failure mode the comment on line 220 says this code exists to prevent.

Worth either making the two writes atomic (or at least rolling back the link on a `resetBaselineFn` failure), or giving `ensureArtifact` a distinct error message for "linked but baseline reset failed" so the remedy it prints (`dr artifact code init`) isn't misleading.

internal/workload/up/build.go
```diff
@@ -180,6 +194,64 @@ func ensureArtifact(loaded Loaded, report *reporter) (string, bool, error) {
...
+		return err
+	}
+
+	// The file manifest is the other half of that baseline, and clearing the
+	// ids without it is worse than clearing neither: the engine reads a BASE
+	// that says every file is already uploaded, so it uploads nothing to the
+	// new artifact, sets no codeRef on it, and the build then fails with
+	// "nothing to build" on an artifact that never received the code.
+	return resetBaselineFn(projectDir, wapi.Manifest{Version: wapi.ManifestVersion})
```
ajalon1: Related to the comment on line 216: this baseline reset also happens *after* `run.go` has already computed and rendered the plan's file-diff count for `--dry-run` (`codeChangeFn` runs before `Build()`/`ensureArtifact` in `Run()`).

So when a linked project's artifact spec has diverged (`reusableArtifact` returns `""` because `specMatches` fails), `--dry-run` prints a diff count computed against the **current** baseline, but a real run then mints a new artifact and empties the baseline here — meaning the actual apply re-uploads every file, materially more than what was previewed. Might be worth a caveat in the dry-run output when a relink is anticipated, or deferring the file-count computation until after the artifact is settled.

internal/workload/up/render.go
```diff
@@ -136,11 +137,35 @@ func lines(s Summary, plan Plan) []string {
...
+// those apart, and it is the only warning there is: the plan is announced and
+// then applied, never confirmed.
+func createDetail(s Summary, plan Plan) string {
 	if s.Name == "" {
+		// Only reachable before a name exists to print; Validate requires one,
+		// so both branches below name the workload.
 		return "will be created, with its first artifact"
 	}
 
+	if plan.PriorWorkloadID != "" {
```
ajalon1: This early return means the `plan.LinkedArtifact && plan.Code.Applies` branch below (line 162) is unreachable whenever `PriorWorkloadID` is also set.

Scenario: a workload is deleted directly via the platform UI (not `dr workload delete`), leaving the manifest's `workloadId` stale (`PriorWorkloadID` set) while the local `.datarobot/` artifact link survives untouched (`LinkedArtifact=true`, `Code.Applies=true`). The printed plan then says only *"... is bound to X, which no longer exists"* and never mentions that the create will actually reuse the linked artifact — which is exactly the information the comment on line 164-165 says matters here ("a delete clears the binding and leaves the artifact link exactly where it was"). Might be worth checking `LinkedArtifact` even when `PriorWorkloadID != ""`, so both facts get surfaced.

internal/workload/up/plan.go
```diff
@@ -86,6 +101,53 @@ func (p Plan) Empty() bool {
...
+// false, starting with the claim that it no longer exists.
+//
+// The risk this trades against is a 404 that means "not here" rather than
+// "deleted": a manifest deployed against the wrong instance, organisation or
+// token resolves to nothing and would be recreated somewhere unintended. That
+// is why the plan names the id it could not find, and why Terraform, which
+// treats a vanished resource the same way, is safe doing it while confirming
+// at apply. `up` announces rather than confirms, so the announcement is what
+// has to carry it.
+func creates(state State) bool {
```
ajalon1, line 139: `creates()` here and `deployable()` (`run.go:477`) are two independently maintained switches over the same `State` enum, deciding which states are creatable vs. which permit any apply at all — with no shared constant or test tying the two classifications together.

`exhaustive` catches a brand-new `State` value that's missing from one of the switches, but not an existing state assigned to the wrong branch in just one of them. A future change to either switch without the other could let `up` silently overwrite a still-live workload, or wrongly refuse a genuinely recoverable one. Might be worth a shared table (or at least a test asserting the two stay consistent) rather than parallel `switch` statements.

internal/workload/up/build.go
```diff
@@ -180,6 +194,64 @@ func ensureArtifact(loaded Loaded, report *reporter) (string, bool, error) {
...
+		return "", fmt.Errorf("cannot read which artifact %s is linked to: %w", loaded.ProjectDir, err)
+	}
+
+	if err := usableArtifact(loaded.ProjectDir, cfg.ArtifactID); err != nil {
+		return "", err
+	}
+
+	// Cannot tell reads as yes. Minting on a failed read would orphan the
+	// artifact this project pushes to over a timeout.
+	if matches, err := specMatches(loaded, cfg.ArtifactID); err != nil || matches {
```
ajalon1: `reusableArtifact` now performs a network GET here (`specMatches`) on every `up` for an already-linked project, in addition to the separate GET `usableArtifact` (line 240) already makes for the same `artifactID` to check lock status.

That's two sequential round-trips to the same artifact resource on the common no-op reconcile path (no drift, nothing to change), adding latency to every ordinary re-deploy. Fetching the artifact document once and deriving both the lock check and the spec comparison from it — or running the two calls concurrently — would avoid the duplicate round trip.

cmd/workload/del/cmd.go
```diff
@@ -128,3 +167,114 @@ func handleDeleteError(err error, workloadID string) error {
...
+	fmt.Fprintln(w, tui.DimStyle.Render(
+		"Removed workloadId from "+path+"; the next 'dr workload up"+dirFlag(dir)+"' creates a new workload."))
+
+	noteLinkedArtifact(w, filepath.Dir(path))
+}
+
+// dirFlag is the " --dir <path>" a follow-up command needs to reach the same
+// project, and empty when the working directory already does. A remedy that
+// names a bare command is false in exactly the layout --dir exists for.
+func dirFlag(dir string) string {
```
ajalon1: `dirFlag(dir string)` here and `dirFlagFor(loaded Loaded)` (`internal/workload/up/run.go:366`) independently compute the same `" --dir <path>"` remedy suffix, but with different, inconsistent robustness — this one does a raw string comparison against `"."`, while `dirFlagFor` uses real `filepath.Abs`/`filepath.Rel` comparison.

If the two notions of "is the cwd already this project's directory" ever diverge (e.g. a relative path with a trailing slash, or `--dir ./.`), `delete`'s and `up`'s remedy text could disagree about whether a `--dir` flag is needed to reach the same project — undermining the exact bug this PR is fixing. Might be worth sharing one implementation.

cmd/workload/del/cmd.go
```diff
@@ -58,17 +70,35 @@ Example:
...
+			// directory, which is the behaviour --dir exists to replace.
+			dir, err := cmd.Flags().GetString("dir")
+			if err != nil {
+				return fmt.Errorf("cannot read --dir: %w", err)
+			}
+
+			// Checked before the delete, not after: a typo is worth catching
+			// while it still costs nothing, and the cleanup that reports it
+			// runs on the far side of an irreversible call.
+			if dir != "" && !fsutil.DirExists(dir) {
```
ajalon1, line 84: The "dir must exist" invariant is now checked three times across two packages, with three different messages: here (`fsutil.DirExists`, `"--dir %s is not a directory"`), in `clearStaleBinding` (line 195, same check), and in `manifest.Locate`'s new `ErrNotADirectory` (`internal/workload/manifest/manifest.go:146-147`).

Neither caller in this file actually consults `manifest.ErrNotADirectory` via `errors.Is` — both re-derive the same check with `fsutil.DirExists` instead of trusting the mechanism this diff just added to `Locate` for exactly this purpose. That leaves `Locate`'s `ErrNotADirectory` path effectively dead code from these call sites, and a future change to what counts as "not a directory" needs updating in three places to stay consistent.

internal/workload/manifest/write.go
```diff
@@ -308,49 +316,201 @@ func Write(path string, data []byte) error {
...
+	// judge it. Swallowing it would leave `dr workload delete` reporting
+	// nothing at all about a binding it declined to take back.
+	if len(docs) > 1 {
+		return nil, nil, fmt.Errorf(
+			"the file holds %d YAML documents, and this edit would not put them back as it found them. "+
+				"Edit it by hand", len(docs))
+	}
+
+	if len(docs) == 0 || len(docs[0].Content) == 0 || docs[0].Content[0].Kind != yaml.MappingNode {
+		return nil, nil, unreadable("cannot %s in %s: root must be a YAML mapping", action, path)
```
ajalon1, line 446: This empty/degenerate-file path produces a generic `"root must be a YAML mapping"` message (wrapped in `ErrUnreadable`), which diverges from `manifest.Parse`'s dedicated, friendlier message for an empty file (`manifest.go:109`: *"file is empty, which an interrupted write can leave behind: delete it and run the command again"*).

An empty `.datarobot.yaml` hit through `WriteWorkloadID` (used by `up`'s `create()` to write back a new workload id) would report this generic message instead of `Parse`'s friendly one — two different explanations for the same underlying problem depending on which internal function happens to touch the file first.

internal/workload/manifest/write.go
```diff
@@ -377,6 +537,190 @@ func setWorkloadID(root *yaml.Node, workloadID string) {
...
+// Three refusals carry their own remedy, because the caller prints what it is
+// given and each needs a different answer:
+//
+//   - the binding is the file's only recognized key, so removing it leaves a
+//     document Parse rejects as not a manifest, and because the file still
+//     exists nothing falls back to the wizard;
+//   - the binding carries a YAML anchor something else aliases, so removing it
+//     strands the alias and the file never loads again;
+//   - repeated keys disagree.
+func clearWorkloadID(root *yaml.Node, wantWorkloadID string) (bool, error) {
```
ajalon1, line 569: `clearWorkloadID` (and its `partition` helper below) hardcode the literal `keyWorkloadID` instead of taking a key name as a parameter, even though the rest of the machinery here (comment preservation, anchor-safety checks, multi-document detection) is otherwise generic.

The next feature that needs to safely strip a different single scalar key from the manifest while preserving comments/anchors (e.g. a future CLI-managed `artifactId` binding) would have no generic "remove this key" routine to call, and would likely end up copy-pasting `partition`/`removable`/`anchorOn`/`aliasesAnchor` with the key name substituted.

internal/workload/up/build.go
wojtekwdr: fixed

internal/workload/up/build.go
wojtekwdr: fixed

internal/workload/up/plan.go
wojtekwdr: fixed

cmd/workload/del/cmd.go
wojtekwdr: fixed

internal/workload/up/render.go
wojtekwdr: fixed

internal/workload/up/build.go
wojtekwdr: fixed

internal/workload/manifest/write.go
```diff
@@ -377,6 +537,190 @@ func setWorkloadID(root *yaml.Node, workloadID string) {
...
+	for i := 0; i+1 < len(root.Content); i += 2 {
+		if root.Content[i].Value != keyWorkloadID {
+			continue
+		}
+
+		at = append(at, i)
+
+		if root.Content[i+1].Value == wantWorkloadID {
+			matches++
+		}
```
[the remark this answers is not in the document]
wojtekwdr: Good catch, fixed. The comparison goes through resolveAlias now, so a binding written as `workloadId: *wid` is cleared like any other. Reproduced against staging first: it was a silent no-op, the delete reported success and left the file exactly as stale as it found it.

internal/workload/up/build.go
wojtekwdr: This one went away in the rebase rather than getting fixed. #823 added `inheritCode`, which points the new artifact at the catalog version already synced, so there was no longer any reason to clear `catalogId` and `lastSyncedVersionId`. `relinkProject` is gone and the link is a single write again.

internal/workload/up/build.go
wojtekwdr: Same rebase resolves this. A replacement no longer empties the baseline, so nothing is re-uploaded and the previewed count matches what the apply does. On staging the sync stays at ~300ms and prints "Carrying the code over" instead of pushing the tree again.

internal/workload/up/render.go
wojtekwdr: Right, fixed. Both facts print now: "... is bound to X, which no longer exists; it comes up on the artifact this project is linked to". The rebase turned up a sibling of this one: with a locked link the same clause claimed reuse while the lock line directly below said the opposite, so the reuse claim is gated on that too.

internal/workload/up/plan.go
wojtekwdr: Added two tests over the whole enum. One asserts every state `creates()` accepts is a state `deployable()` lets through, the other pins the classification per state so moving one between branches has to be a deliberate edit.

internal/workload/up/build.go
wojtekwdr: Fixed. After the rebase it was three questions rather than two, since the lock status, the lineage and the spec all come off the same document, so it is now one read for all three. One correction though: `reusableArtifact` only ran on the create path, never on the no-op reconcile.

cmd/workload/del/cmd.go
wojtekwdr: Fixed, both go through `manifest.DirFlag` now. It resolves the path instead of comparing text, so `--dir ./.` correctly produces no suffix where the old comparison printed one.

cmd/workload/del/cmd.go
wojtekwdr: Down to two, and the remaining one consults the sentinel: `clearStaleBinding` branches on `errors.Is(err, manifest.ErrNotADirectory)` from `Locate` rather than re-deriving the check. The one before the delete stays, since `Locate` only runs on the far side of an irreversible call and a typo is worth catching while it still costs nothing.

internal/workload/manifest/write.go
wojtekwdr: Fixed. The wording moved to a shared `ErrEmptyFile` that both `Parse` and the write path use, so an empty file gets one explanation and one remedy whichever function reaches it first.

internal/workload/manifest/write.go
wojtekwdr: Leaving this one. The four refusals are each about what a binding means rather than about removing a key, so a generic helper would end up taking most of them back as policy arguments. Worth revisiting if a second CLI-managed key actually turns up.

wojtekwdr:
Thanks, all of these were worth having. Nine are fixed, one declined inline, and two dissolved in the rebase onto #823 rather than needing a fix.

The `up.Load` one from your summary is fixed too, but at the cobra layer rather than in `Load`: `Load` takes a directory without knowing whether a flag supplied it, so naming `--dir` from there would be wrong when it came from the cwd. `resolveDir` does the check instead, and `dr workload up --dir <typo>` now matches what `delete` says.

On symlinks, you were right and it is worse than I assumed. Deploying through a symlinked `.datarobot.yaml` replaces the symlink with a regular file, and the shared target keeps the old binding. It is `up`'s own `WriteWorkloadID` doing it, and both that and `AtomicWriteFile` predate this branch, so this change widens the exposure without causing it. #817 is the right fix and the two are worth landing close together.

Two things a re-read should focus on, both from the rebase rather than the original change:

* `ensureArtifact` is a genuine semantic merge, not a conflict resolution. Your locked redraft and my spec-divergence guard are two reasons for the same act, so a diverged spec now routes through `freshArtifact`/`redraftRepository` and joins the lineage it replaces. Verified on staging that both artifact versions share one repository.
* I dropped my sync-baseline reset entirely, because `inheritCode` is a better answer to the same failure. Measured on staging: the replacement path went from re-uploading the whole tree to uploading nothing and carrying the code reference over.

ajalon1 (approved):
send it

PR #825, opened 2026-08-24
[RAPTOR-19533] fix(workload): binding to live C2W workloads writes a manifest the CLI then refuses

ajalon1:
# RATIONALE

**20260826 update:** branch merged with latest main; the RAPTOR-19716 field-metadata schema table (originally a stacked follow-up) is now included as Part 6, and round-4 review nits are addressed (stable sort for key ordering, falsifiable strip-pin tests, stale comment cleanup).

---

Let's make the CLI a manifest it can't refuse. 😎 

`dr workload config` (binding) was completely broken against real C2W accounts.

1. The platform routinely echoes `replicaCount` and `autoscaling: null` together in workload GET responses, and the CLI's manifest validator treated the present-but-null `autoscaling` key as an active autoscaling policy, rejecting the file it just rendered. Binding aborted with nothing written, leaving hand-editing as the only path forward.
2. A second defect wrote `build.artifactImageBuildId` (a server-assigned build output) into the committed manifest, pinning the next deploy to a stale image.
3. A third issue sorted keys alphabetically in the rendered file instead of leading with `name`.

User report from Benjamin Kostiuk (Slack, 2026-08-18), reproduced on staging by Wojciech Wierzchowski (4/4 workloads failed).

### Part 1: Validator null handling (`validate.go`, `node.go`)

- Added two presence predicates to `node.go`, chosen per field according to the server's own semantics:
  - `isNullish(*yaml.Node) bool` — true for nil nodes, `!!null` scalars, AND zero-content mappings. Used only where the server itself reads an empty block as absent: `checkScaling` / autoscaling.
  - `isNull(*yaml.Node) bool` — true only for nil nodes and `!!null` scalars. Guards the exactly-one-of checks (`checkArtifactBinding`, `checkImageSource`, `checkArtifact`), so empty blocks (`artifact: {}`, `imageBuildConfig: {}`) count as PRESENT. The server never echoes `{}` for these fields (it echoes `null`); it 422s on `artifact: {}` with a confusing missing-fields error, and on create it silently discards a declared `imageUri` when `imageBuildConfig: {}` is sent (Pydantic default-constructs the empty block into a real `ImageBuildConfig`, and `_strip_client_image_uris_on_build_containers` then nulls the image).
- Fixed `checkScaling` to treat null/empty `autoscaling` as absent (no false conflict with `replicaCount`), matching `autoscalingActive` in `live.go`.
- Fixed the same null blind spot in other node-based checks: `checkArtifactBinding` (`artifact: null`), `checkArtifact` (`spec: null`), `checkImageSource` (`imageBuildConfig: null`), `checkBuildConfig` (`dockerfile: null`).
- Fixed `checkRuntime` to use `isNullish` for consistency.
- Fixed the `enabled: null` shape in `checkScaling` — changed `enabled != nil` to `!isNullish(enabled)` so a null `enabled` inside a non-empty autoscaling block is treated as active (server default), agreeing with `autoscalingActive`. (The validator gates hand-edited files, where `enabled: null` is reachable; the server's non-nullable `enabled` means the live reader needs no such defense.)
- Added `checkEnvironmentVars`: string-variant `environmentVars` entries must carry a non-null `value` (`value: ""` allowed); `dr-credential` and `api-key` sources are exempt. This mirrors the server schema (`StringEnvironmentVariable.value: str = Field(...)` — required, no default; the credential/api-key variants have no `value` field at all). Previously a value-less string entry validated cleanly and only failed as a 422 at deploy time.
- `checkBuildConfig`'s dockerfile-required error now anchors to the `dockerfile` node's line (matching `checkArtifact`'s pattern) instead of the parent block.
- Added `TestValidate_AutoscalingAgreesWithLiveReader` — a parametrized test that runs both readers across all seven autoscaling shapes and asserts they agree. This is the regression guard for the bug class this PR fixes.
- Regression tests for every null/empty shape above, including empty-block exclusivity (`artifact: {}` + `artifactId`, `imageUri` + `imageBuildConfig: {}` are both rejected) and message quality for standalone empty blocks.

### Part 2: Render stripping (`live.go`)

- Folded null-key stripping into `stripKeys`'s existing recursive walk (same pass that strips server-managed keys like `id`, `createdAt`, etc.), eliminating a redundant second deep-copy. Nil-valued keys are dropped; empty maps, `false`, `0`, `""`, and empty slices are preserved.
- Added `"resolvedBundle"` to `serverManagedKeys` — it's a server-assigned, read-only scheduling field.
- Extended `stripBuildOutputs` to also delete the `build` block (`artifactImageBuildId`, `status`, `createdAt`) and `imageOutdated` from containers — both are server-assigned outputs that don't belong in a human-editable file.
- Reworked `stripBuildOutputs` ordering: `codeRef` is deleted first; `imageUri` is deleted only when the build config retains content (a real build container); if deleting `codeRef` empties the block, the `imageBuildConfig` key itself is dropped and the user-declared `imageUri` is kept. `liveBuild` treats an empty build block as image mode, not "unchanged".
- Added `TestLive_RenderRoundTripReporterShaped` — a round-trip regression test using a server document shaped like the reporter's (with `autoscaling: null` + `replicaCount` + `build.artifactImageBuildId` + `imageOutdated` + `resolvedBundle`), asserting the rendered file passes `Parse` + `Validate` + `Compile` and contains none of those server-assigned fields.
- Added standalone tests: `TestLive_RenderStripsNullAutoscaling`, `TestLive_RenderStripsBuildBlock`, `TestLive_RenderStripsImageOutdated`, `TestLive_RenderStripsResolvedBundle`.

### Part 3: Key ordering (`live.go`)

- Added `hoistNameKeys(*yaml.Node)` post-processor that moves `name` to the front of sequence-item mappings under `containerGroups` and `containers` — the identifier mappings a human scans for — preserving all other keys in their existing order. Opaque mappings (e.g. a `labels:` block that happens to contain a `name` key) keep their original key order. Applied to the encoded spec and runtime nodes in `Render()` before the final YAML encode. The top-level mapping (already manually ordered) is not touched.

### Part 4: Spec-less artifact fails loudly (`live.go`, `up/look.go`, `wizard/run.go`)

- **Signature change:** `NewLive` now returns `(Live, error)`. Binding to a workload whose artifact doc has no usable `spec` fails with an error naming the workload ID instead of rendering a manifest the validator would reject. Callers updated: the wizard bind flow propagates the error; `up`'s drift-comparison `Look` propagates it when an artifact doc was present (empty artifactID / 404 remain exempt, since no artifact is a valid state for drift comparison).
- The error wording is path-neutral (`workload %s has no artifact spec; edit %s by hand instead`) because it surfaces on both the bind path and `dr workload up`, where the user isn't binding anything.
- Pinned by `TestLook_SpeclessArtifactFailsLoudly` and `TestNewLive_*FailsWithWorkloadID` tests.

### Part 5: Staging investigation

- Bound to a live C2W workload on staging, verified `dr workload config --dry-run` no longer aborts, verified the rendered manifest is clean (no `autoscaling: null`, no `build.artifactImageBuildId`, no `imageOutdated`, no `resolvedBundle`).
- Ran `dr workload up` after binding: the platform POSTs a new artifact with `imageBuildConfig` but no `imageUri`, builds a fresh image, and the workload resolves to the newly built image. **ErrImagePull is resolved** — stripping the `build` block and `imageUri` correctly signals to the platform that a rebuild is needed.

### Part 6: Field-metadata schema table (`schema.go`, `keys.go`, `write.go`)

- Added `schema.go`: a single declarative `fieldRule` table holding per-field metadata (`nullable`, `serverManaged`, `buildOutput`, `buildConfigScoped`, `hoistName`, `topLevelOrder`) for every manifest field this package has an opinion about. Four previously scattered knowledge sites now derive from it:
  - `serverManagedKeys` — the render-time strip set (`id`, `createdAt`, `status`, `endpoint`, `resolvedBundle`, …)
  - `stripBuildOutputs` — build outputs and their build-config scoping (the conditional `imageUri`/`imageBuildConfig` deletion stays as code: it depends on whether the build config is empty after stripping, which a static table cannot express)
  - `hoistNameKeys` — the `name`-first ordering targets
  - top-level field ordering, now shared between `Live.Render` and `Draft`'s top-level fields (`write.go`) via one table-derived ordering — the two renderers can no longer drift apart in key sequence
- Added the 9 missing field-name constants to `keys.go` (`keyID`, `keyCreatedAt`, `keyUpdatedAt`, `keyStatus`, `keyEndpoint`, `keyResolvedBundle`, `keyImageOutdated`, `keyBuild`, `keyCodeRef`) so the table and its consumers reference constants instead of string literals.
- Behavior-preserving: the any-depth strip scope of `stripKeys`/`stripBuildOutputs` is unchanged and is now documented on the table itself (the create spec has no legitimate key by those names anywhere, so recursive deletion is safe; scoping to known paths would be a deliberate behavior change).
- Pinned by `schema_test.go`: `TestSchema_TopLevelKeyOrderMatches` (draft and render agree on top-level key order) and `TestSchema_ConsolidationPinStripAndHoist` (table-derived strip/hoist sets match the previous hard-coded behavior).
- Tracked as [RAPTOR-19716](https://datarobot.atlassian.net/browse/RAPTOR-19716); originally planned as a stacked follow-up PR, merged into this branch instead.

## What we're building towards in the API

The workload API (Python/FastAPI/Pydantic v2) sends `autoscaling: null` explicitly — not omitted — because the response routes (`GET /workloads/{id}`, `GET /artifacts/{id}`) do not set `response_model_exclude_none=True`, and the field is typed as `AutoscalingProperties | None` with default `None`. The OpenAPI schema declares it as nullable (`anyOf: [AutoscalingProperties, null]`).

The API also populates `build` (`ContainerBuildInfo` with `artifactImageBuildId`, `status`, `createdAt`) on containers after a build trigger, and strips it on create/update via `_strip_client_container_build_metadata` — but it is present in GET responses. Similarly, `imageOutdated` (computed, read-only) and `resolvedBundle` (scheduling-time, read-only) appear in GET responses but are not user inputs.

The CLI's manifest ledger was written in RAPTOR-18973 from the API's documented shapes rather than from observed responses. The documented shapes say `replicaCount` and `autoscaling` are mutually exclusive, but in practice the server echoes both (with `autoscaling: null`). This is the second time the ledger has diverged from reality for this reason. The fix aligns the CLI with the actual server behavior: null means absent, and server-assigned outputs are stripped from the human-editable file.

One nuance this PR encodes explicitly: "empty block" is NOT the same as "null block" for most fields. Only `autoscaling` reads `{}` as absent server-side. `artifact: {}` is a 422, and `imageBuildConfig: {}` is worse — Pydantic default-constructs it and the create path silently drops your `imageUri`. The validator's two presence predicates (`isNull` vs `isNullish`) encode exactly that distinction.

## TESTING

- `task test` — full suite with `-race`, all packages pass (manifest package at ~90% coverage)
- `task lint` — golangci-lint clean on linux/darwin/windows, gofumpt, go vet, goreleaser check
- `task build` — CLI binary compiles
- Staging: end-to-end bind + deploy against a live C2W workload confirmed the fix

## RELATED

- [RAPTOR-19533](https://datarobot.atlassian.net/browse/RAPTOR-19533)
- Parent epic: [RAPTOR-18002](https://datarobot.atlassian.net/browse/RAPTOR-18002) ([WAPI GA] DR CLI extended support)
- Related: [RAPTOR-19406](https://datarobot.atlassian.net/browse/RAPTOR-19406) (artifact repository reuse — likely related to the ErrImagePull case, now resolved by this fix)
- [RAPTOR-19716](https://datarobot.atlassian.net/browse/RAPTOR-19716) — consolidates this package's scattered field metadata (server-managed keys, strip/hoist targets, top-level ordering) into one declarative rule table. Included in this PR as Part 6.

---

> [!NOTE]
> **Medium Risk**
> Changes workload manifest validation and live-binding render paths that gate deploy; behavior is broader but aligned with server GET shapes and heavily regression-tested.
> 
> **Overview**
> Fixes **`dr workload config`** binding against real C2W workloads where the CLI rendered a manifest and then **`Validate`** rejected it (notably **`replicaCount`** with server-echoed **`autoscaling: null`**).
> 
> **Validator:** Adds **`isNullish`** for YAML null/empty mappings and uses it in scaling, artifact binding, image/build checks, and runtime handling so present-but-null keys count as absent. **`checkScaling`** and map-based **`autoscalingActive`** now agree on when autoscaling is “on,” including **`enabled: null`** inside a non-empty block.
> 
> **Live → manifest render:** Strips **`resolvedBundle`**, nil-valued keys, server **`build`** metadata, and **`imageOutdated`**; extends **`stripBuildOutputs`** accordingly. **`reorderKeys`** puts **`name`** first in nested spec/runtime YAML. Input API docs are deep-copied and not mutated.
> 
> Regression tests cover reporter-shaped round-trips, null autoscaling, key order, and validator/live autoscaling parity.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit a0277f6c83b6ea69822923036dee04a17eb82f82. Configure [here](https://cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/README.md
+++ b/README.md
@@ -40,6 +40,7 @@ If you're new to DataRobot, visit the [DataRobot documentation](https://docs.dat
 - 📦 **Template management**&mdash;clone and configure application templates interactively.
 - ⚙️ **Interactive configuration**&mdash;smart wizard for environment setup with validation.
 - 🚀 **Task runner**&mdash;execute application tasks with built-in Taskfile integration.
+- 🛠️ **Workload management**&mdash;bind to live C2W workloads and deploy from manifests (`dr workload config`), gated behind `DATAROBOT_CLI_FEATURE_WORKLOAD=true`.
 - 🐚 **Shell completions**&mdash;support for Bash, Zsh, Fish, and PowerShell.
 - 🔄 **Self-update capability**&mdash;easily update to the latest version with a single command.
 
--- a/README.md
+++ b/README.md
@@ -40,7 +40,6 @@ If you're new to DataRobot, visit the [DataRobot documentation](https://docs.dat
 - 📦 **Template management**&mdash;clone and configure application templates interactively.
 - ⚙️ **Interactive configuration**&mdash;smart wizard for environment setup with validation.
 - 🚀 **Task runner**&mdash;execute application tasks with built-in Taskfile integration.
-- 🛠️ **Workload management**&mdash;bind to live C2W workloads and deploy from manifests (`dr workload config`), gated behind `DATAROBOT_CLI_FEATURE_WORKLOAD=true`.
 - 🐚 **Shell completions**&mdash;support for Bash, Zsh, Fish, and PowerShell.
 - 🔄 **Self-update capability**&mdash;easily update to the latest version with a single command.
 
--- a/internal/workload/manifest/live.go
+++ b/internal/workload/manifest/live.go
@@ -960,18 +960,24 @@ func stripKeys(value any) {
 // purge strips enabled: null, leaving {} — which both autoscalingActive and
 // checkScaling read as OFF (absent), inverting the group's semantics.
 //
-// Only the nil placeholder is rewritten. A present bool (true or false) or an
-// absent enabled key already carries the right meaning and is left untouched.
-// A block like {enabled: null, minReplicaCount: 2} already reads as ON after
-// the purge (enabled absent → default on); normalizing it to enabled: true
-// makes the default explicit and keeps the rendered file free of nulls.
+// Only a present-with-nil enabled is rewritten. A present bool (true or false)
+// or an absent enabled key already carries the right meaning and is left
+// untouched. The two-value map-access idiom distinguishes "enabled present
+// with nil value" from "enabled absent entirely" — the single-value
+// `autoscaling[keyEnabled] == nil` cannot (Go map zero-value ambiguity), and
+// would rewrite a genuinely empty autoscaling: {} to {enabled: true},
+// inverting a should-be-OFF group to ON. A block like {enabled: null,
+// minReplicaCount: 2} already reads as ON after the purge (enabled absent →
+// default on); normalizing it to enabled: true makes the default explicit and
+// keeps the rendered file free of nulls.
 func normalizeAutoscalingEnabled(doc map[string]any) {
 	autoscaling, ok := doc[keyAutoscaling].(map[string]any)
 	if !ok {
 		return
 	}
 
-	if autoscaling[keyEnabled] == nil {
+	val, present := autoscaling[keyEnabled]
+	if present && val == nil {
 		autoscaling[keyEnabled] = true
 	}
 }
--- a/internal/workload/manifest/live.go
+++ b/internal/workload/manifest/live.go
@@ -843,9 +843,10 @@ func documentNode(document map[string]any) (*yaml.Node, error) {
 // blocks are "passed through as the platform gave them"). Reordering every
 // mapping with a name key produced unexplained diff noise on every bind.
 //
-// The walk descends into every value so containerGroups/containers at any
-// depth are found (containers nest inside groups, and a future spec may nest
-// groups inside something else), but it only reorders the mappings that are
+// The walk is scoped to mapping nodes: it returns early on anything else
+// (scalars and sequences carry no containerGroups/containers keys). Within a
+// mapping it recurses into each value, so containerGroups/containers nested
+// inside other mappings are found, but it only reorders the mappings that are
 // sequence items under those two keys. The top-level manifest order is
 // already controlled by Render's explicit field list and is never reached
 // here.
@@ -960,9 +961,13 @@ func stripKeys(value any) {
 // container: imageUri is deleted as a server-built pin and codeRef as
 // server-managed. An imageBuildConfig that is empty, or becomes empty after
 // deleting codeRef, is NOT a build container — the user-declared imageUri is
-// kept and the emptied imageBuildConfig key is removed entirely so it never
-// renders as "{}" alongside a missing imageUri, which Validate would reject
-// with "must set either imageUri or imageBuildConfig".
+// kept and the emptied imageBuildConfig key is removed entirely. This is an
+// invariant guard mirroring the server's own rule at containers.py:415
+// ("either imageUri or imageBuildConfig must be provided"): the rendered
+// container must always carry exactly one valid image source, never an empty
+// imageBuildConfig: "{}" alongside a missing imageUri. Today the server-side
+// non-nullability of dockerfile makes the empty case unreachable from real
+// payloads, but the guard fails safe if that ever changes.
 func stripBuildOutputs(spec map[string]any) {
 	for _, group := range slicesAt(spec, keyContainerGroups) {
 		for _, container := range slicesAt(group, keyContainers) {
--- a/internal/workload/manifest/live.go
+++ b/internal/workload/manifest/live.go
@@ -62,8 +62,11 @@ func NewLive(workloadID string, workloadDoc, artifactDoc map[string]any) Live {
 		Importance:   stringAt(workloadDoc, keyImportance),
 		ArtifactName: stringAt(artifactDoc, keyName),
 		ArtifactType: stringAt(artifactDoc, keyType),
-		Spec:         stripNullKeys(stripServerManaged(mapAt(artifactDoc, keySpec))),
-		Runtime:      stripNullKeys(stripServerManaged(mapAt(workloadDoc, keyRuntime))),
+		// stripServerManaged deep-copies once and strips both server-managed
+		// keys and nil-valued keys in a single recursive walk, so no second
+		// copy is allocated here.
+		Spec:    stripServerManaged(mapAt(artifactDoc, keySpec)),
+		Runtime: stripServerManaged(mapAt(workloadDoc, keyRuntime)),
 	}
 
 	stripBuildOutputs(live.Spec)
@@ -862,7 +865,18 @@ func stripKeys(value any) {
 			delete(typed, key)
 		}
 
-		for _, child := range typed {
+		// Nil-valued keys are dropped in the same walk: the server echoes
+		// them as placeholders (e.g. autoscaling: null), and writing one
+		// into the committed manifest would confuse the validator. Empty
+		// maps are preserved; isNullish is the validator's counterpart that
+		// treats an empty mapping as absent rather than as a policy.
+		for key, child := range typed {
+			if child == nil {
+				delete(typed, key)
+
+				continue
+			}
+
 			stripKeys(child)
 		}
 	case []any:
@@ -893,53 +907,8 @@ func stripBuildOutputs(spec map[string]any) {
 	}
 }
 
-// stripNullKeys returns a copy of the document with every nil-valued key
-// removed recursively. Falsy-but-present values (false, 0, empty strings,
-// empty slices) are kept. The input document is not modified.
-func stripNullKeys(document map[string]any) map[string]any {
-	if document == nil {
-		return nil
-	}
-
-	copied := make(map[string]any, len(document))
-	for key, value := range document {
-		if value == nil {
-			continue
-		}
-
-		switch typed := value.(type) {
-		case map[string]any:
-			copied[key] = stripNullKeys(typed)
-		case []any:
-			copied[key] = stripNullSlice(typed)
-		default:
-			copied[key] = value
-		}
-	}
-
-	return copied
-}
-
-func stripNullSlice(items []any) []any {
-	if items == nil {
-		return nil
-	}
-
-	copied := make([]any, len(items))
-	for i, value := range items {
-		switch typed := value.(type) {
-		case map[string]any:
-			copied[i] = stripNullKeys(typed)
-		case []any:
-			copied[i] = stripNullSlice(typed)
-		default:
-			copied[i] = value
-		}
-	}
-
-	return copied
-}
-
+// deepCopyMap returns a deep copy of document so the caller's input is not
+// mutated by the strip passes that run against it.
 func deepCopyMap(document map[string]any) map[string]any {
 	if document == nil {
 		return nil
--- a/internal/workload/manifest/live.go
+++ b/internal/workload/manifest/live.go
@@ -924,19 +924,12 @@ func stripKeys(value any) {
 			delete(typed, key)
 		}
 
-		// Normalize enabled: null inside a present autoscaling block before
-		// the nil-purge would collapse it to {}. The platform's default for a
-		// configured autoscaling block is ON, so enabled: null carries that
-		// meaning; rewriting it to true keeps the rendered file free of nulls
-		// (invariant 3) and semantically honest — both autoscalingActive and
-		// checkScaling read the post-strip shape as ON.
-		normalizeAutoscalingEnabled(typed)
-
 		// Nil-valued keys are dropped in the same walk: the server echoes
 		// them as placeholders (e.g. autoscaling: null), and writing one
 		// into the committed manifest would confuse the validator. Empty
 		// maps are preserved; isNullish is the validator's counterpart that
-		// treats an empty mapping as absent rather than as a policy.
+		// treats an empty mapping as absent rather than as a policy. The
+		// nullable set is documented in library/api-reference.md.
 		for key, child := range typed {
 			if child == nil {
 				delete(typed, key)
@@ -950,35 +943,6 @@ func stripKeys(value any) {
 		for _, item := range typed {
 			stripKeys(item)
 		}
-	}
-}
-
-// normalizeAutoscalingEnabled rewrites enabled: null inside a present
-// autoscaling block to enabled: true before the nil-purge would collapse it.
-// The platform's default for a configured autoscaling block is ON, so an
-// explicit enabled: null carries that meaning. Without this normalization the
-// purge strips enabled: null, leaving {} — which both autoscalingActive and
-// checkScaling read as OFF (absent), inverting the group's semantics.
-//
-// Only a present-with-nil enabled is rewritten. A present bool (true or false)
-// or an absent enabled key already carries the right meaning and is left
-// untouched. The two-value map-access idiom distinguishes "enabled present
-// with nil value" from "enabled absent entirely" — the single-value
-// `autoscaling[keyEnabled] == nil` cannot (Go map zero-value ambiguity), and
-// would rewrite a genuinely empty autoscaling: {} to {enabled: true},
-// inverting a should-be-OFF group to ON. A block like {enabled: null,
-// minReplicaCount: 2} already reads as ON after the purge (enabled absent →
-// default on); normalizing it to enabled: true makes the default explicit and
-// keeps the rendered file free of nulls.
-func normalizeAutoscalingEnabled(doc map[string]any) {
-	autoscaling, ok := doc[keyAutoscaling].(map[string]any)
-	if !ok {
-		return
-	}
-
-	val, present := autoscaling[keyEnabled]
-	if present && val == nil {
-		autoscaling[keyEnabled] = true
 	}
 }
 
--- a/internal/workload/manifest/live.go
+++ b/internal/workload/manifest/live.go
@@ -471,7 +471,7 @@ func (l *Live) runtimeGroup() map[string]any {
 // autoscaling: the file it would produce is one the ledger accepts.
 func autoscalingActive(group map[string]any) bool {
 	autoscaling := mapAt(group, keyAutoscaling)
-	if autoscaling == nil {
+	if len(autoscaling) == 0 {
 		return false
 	}
 
--- a/internal/workload/manifest/live.go
+++ b/internal/workload/manifest/live.go
@@ -26,7 +26,9 @@ import (
 // server's own bookkeeping: it assigns them, and a file that repeats them
 // back is either ignored or rejected. Stripping them recursively is safe
 // because the create spec has no legitimate key by these names anywhere.
-var serverManagedKeys = []string{"id", "createdAt", "updatedAt", "status", "endpoint", "artifactId", "workloadId"}
+var serverManagedKeys = []string{
+	"id", "createdAt", "updatedAt", "status", "endpoint", "artifactId", "workloadId", "resolvedBundle",
+}
 
 // Live is a running workload, downloaded so a manifest can be written that
 // says everything the workload already says. Setup uses it when the user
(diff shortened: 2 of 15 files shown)


README.md
```diff
@@ -40,6 +40,7 @@ If you're new to DataRobot, visit the [DataRobot documentation](https://docs.dat
 - 📦 **Template management**&mdash;clone and configure application templates interactively.
 - ⚙️ **Interactive configuration**&mdash;smart wizard for environment setup with validation.
 - 🚀 **Task runner**&mdash;execute application tasks with built-in Taskfile integration.
+- 🛠️ **Workload management**&mdash;bind to live C2W workloads and deploy from manifests (`dr workload config`), gated behind `DATAROBOT_CLI_FEATURE_WORKLOAD=true`.
```
ajalon1:
```suggestion
```

Dropping this, since it's still behind a feature gate.

internal/workload/manifest/node.go
```diff
@@ -29,6 +29,23 @@ const (
 	boolTag = "!!bool"
 )
 
+// isNullish reports whether a node is meaningfully absent for validator
+// checks. mapValue returns a non-nil scalar node for a present key whose value
+// is null, so callers that test `!= nil` to mean "this block is set" get a
+// false positive unless they also call this helper.
+func isNullish(node *yaml.Node) bool {
```
ajalon1, line 36: I like the name, but I really don't like this -- I wish the API had better enforcement.

internal/workload/manifest/validate.go
```diff
@@ -212,7 +219,7 @@ func (v *validator) checkArtifactBinding(root *yaml.Node) {
 	inline := mapValue(root, keyArtifact)
 	byID, _ := scalarString(mapValue(root, keyArtifactID))
 
-	if (inline != nil) == (byID != "") {
+	if (!isNullish(inline)) == (byID != "") {
```
wojtekwdr: [high] `isNullish` collapses `artifact: {}` into "absent", so this no longer catches `artifact: {}` set alongside `artifactId`. I confirmed Validate() passes that manifest and Compile() forwards both `artifact:{}` and `artifactId` verbatim to the API, since Compile does no null/empty normalization of its own. The old `inline != nil` check caught this correctly before this PR.

internal/workload/manifest/validate.go
```diff
@@ -319,15 +327,16 @@ func (v *validator) checkImageSource(container *yaml.Node, path string) {
 	imageURI, hasImage := scalarString(mapValue(container, keyImageURI))
 	hasImage = hasImage && imageURI != ""
 	buildConfig := mapValue(container, keyImageBuildConfig)
+	hasBuildConfig := !isNullish(buildConfig)
```
wojtekwdr: [high] Same isNullish gap as the artifact/artifactId exclusivity check: `hasBuildConfig` is false for `imageBuildConfig: {}`, so a container with both `imageUri` and an empty `imageBuildConfig: {}` passes here too. Confirmed Validate() returns nil for that shape and Compile() sends both fields through, contradicting the either-or rule this file's own comment on `stripBuildOutputs` cites from containers.py:415.

internal/workload/manifest/validate.go
```diff
@@ -422,10 +435,45 @@ func (v *validator) checkPort(container *yaml.Node, path string, primary bool) {
...
+// container's environmentVars sequence. This is a Surface B guard: the
+// validator's input is a hand-editable file, so a user can write value: null
+// or omit value entirely regardless of what the server schema allows on the
+// wire. The rule requires a non-null value for the string variant (source
+// absent or "string"); the credential and api-key variants carry no value and
+// are exempt. An explicitly empty value ("") passes — that is the valid
+// spelling of an intentionally-empty variable. Surfacing a value-less string
+// entry at validate time (with a line anchor) beats discovering it as a 422
+// at deploy time.
+func (v *validator) checkEnvironmentVars(container *yaml.Node, path string) {
```
wojtekwdr, line 474: [medium] question: this is a brand new required-value rule, not mentioned anywhere in the PR description (which only covers null-handling, output stripping, and key ordering). A manifest with `environmentVars: [{name: X}]` (no value, no source) that validated cleanly before this PR will now fail. Intentional tightening, or should this ship separately?

internal/workload/manifest/live.go
```diff
@@ -841,21 +951,49 @@ func stripKeys(value any) {
...
+// so it lives in that list instead.
+//
+// An imageBuildConfig with real content (e.g. a dockerfile block) is a build
+// container: imageUri is deleted as a server-built pin and codeRef as
+// server-managed. An imageBuildConfig that is empty, or becomes empty after
+// deleting codeRef, is NOT a build container — the user-declared imageUri is
+// kept and the emptied imageBuildConfig key is removed entirely. This is an
+// invariant guard mirroring the server's own rule at containers.py:415
+// ("either imageUri or imageBuildConfig must be provided"): the rendered
+// container must always carry exactly one valid image source, never an empty
```
wojtekwdr: [medium] This invariant doesn't actually hold: a container whose `imageBuildConfig` uses an unrecognized source (e.g. `{buildpack: {...}}`, no `dockerfile` key) still renders with a non-empty `imageBuildConfig` and no `imageUri`, and `checkBuildConfig` then rejects it for a missing `dockerfile`. Reproducible with `TestLive_UnknownBuildSourceIsPreserved`'s own fixture run through Render+Validate. Looks pre-existing rather than introduced here, but since it's exactly the failure class this PR fixes elsewhere, worth a `dockerfile`-key check in `stripBuildOutputs` too, or softening the claim.

internal/workload/manifest/live_test.go
```diff
@@ -792,3 +823,1016 @@ func TestLive_ApplyRuntime(t *testing.T) {
...
+		container map[string]any
+	}{
+		"natively empty": {
+			container: map[string]any{
+				"name":             "app",
+				"imageUri":         "registry.example.com/app:v1",
+				"imageBuildConfig": map[string]any{},
+			},
+		},
+		"emptied after codeRef removal": {
```
wojtekwdr: [low] nit: this case is byte-for-byte identical to "natively empty" above (same name, imageUri, empty imageBuildConfig), so it can only pass or fail together with it. The comment describes simulating post-strip leftover state, but the fixture doesn't reflect that.

internal/workload/manifest/live.go
```diff
@@ -53,20 +56,36 @@ type Live struct {
...
+		// copy is allocated here.
+		Spec:    stripServerManaged(mapAt(artifactDoc, keySpec)),
+		Runtime: stripServerManaged(mapAt(workloadDoc, keyRuntime)),
 	}
 
 	stripBuildOutputs(live.Spec)
 
-	return live
+	if len(live.Spec) == 0 {
+		return live, fmt.Errorf("workload %s has no artifact spec to bind; edit %s by hand instead",
```
wojtekwdr: [low] non-blocking: this message says "to bind", but it's also surfaced verbatim through `dr workload up` (look.go), where the user isn't binding anything, just deploying an already-bound workload. Might read as confusing on that path.

wojtekwdr (changes requested):
The isNullish fix is right for the null-echo bug, but it also loosens the two exactly-one-of checks: `artifact: {}` with `artifactId`, and `imageUri` with `imageBuildConfig: {}`, both pass Validate now and Compile sends both sides to the API. The rest is non-blocking, though the description is missing the NewLive signature change and Part 3 describes a wider name-hoist than what shipped.

benjaminkostiuk-dr (commented):
Thanks for addressing my concerns! My go is a little fuzzy and I don't have too much context of the inner workings here, but I think this looks good. I think @wojtekwdr's comments call out better what might be missing.

internal/workload/manifest/validate.go
ajalon1: This was an intentional decision on my part. I wasn't sure about letting this one go through, but since this task found me doing a lot more defensive checking against possible user inputs I thought I'd put it in.

Happy to take this out.

ajalon1:
@wojtekwdr Thanks, I expected nothing less from my first "real" contribution to this side of the house. 😅 Will see about tightening things down around YAML sorting and now that I've slept on it -- most of yesterday was just understanding the workflow better for testing these changes.

internal/workload/manifest/validate.go
ajalon1: On second thought I'd like to keep it in: the rule mirrors the server schema — `StringEnvironmentVariable.value` is `str = Field(...)` (required, no default) in workload-api, while the `dr-credential` and `api-key` variants have no `value` field at all. A value-less string entry passes the old validator but 422s at deploy time, so pulling this out of the PR would ship a manifest that "works" right up until deploy. I'll add it to the PR description so the scope is explicit.

internal/workload/manifest/validate.go
ajalon1: Fixed (a95009b): isNull now guards exclusivity; empty blocks count as present.

internal/workload/manifest/validate.go
ajalon1: Fixed (a95009b): same isNull guard; imageBuildConfig: {} counts as present.

internal/workload/manifest/validate.go
ajalon1: Kept: mirrors server schema; rationale above, now in PR description.

internal/workload/manifest/live.go
ajalon1: Comment softened (1dfd0b8): dockerfile-only scope; other sources unreachable server-side.

internal/workload/manifest/live_test.go
ajalon1: Replaced (1dfd0b8) with NewLive-driven strip-pipeline test.

internal/workload/manifest/live.go
ajalon1: Reworded (1dfd0b8): path-neutral message works on both paths.

internal/workload/manifest/schema_test.go
```diff
@@ -0,0 +1,220 @@
...
+
+	// Server-managed key values must not appear in the rendered output.
+	// workloadId is the exception: it is the one binding the file carries.
+	for _, unwanted := range []string{
+		"68a0000000000000000000a1", // artifactId value
+		"app.datarobot.com",        // endpoint value
+		"status: running",          // workload status
+		"status: locked",           // artifact status
+		"2026-01-01T00:00:00Z",     // createdAt value
+		"2026-01-02T00:00:00Z",     // updatedAt value
```
wojtekwdr, line 90: [medium] Six of these twelve checks can't fail: `NewLive` only reads `spec` off the artifact doc and `runtime` off the workload doc, so the top-level `artifactId`, `endpoint`, `status` and `createdAt`/`updatedAt` values never reach `Render` whatever the table says. Delete `keyStatus` from `fieldRules` and this test still passes. The nested ones (`resolvedBundle`, `imageOutdated`, `codeRef`, the build block) are what's actually pinning the consolidation.

internal/workload/manifest/validate.go
```diff
@@ -422,10 +461,45 @@ func (v *validator) checkPort(container *yaml.Node, path string, primary bool) {
...
+		return
+	}
+
+	for i, entry := range seqItems(vars) {
+		entryPath := fmt.Sprintf("%s.%s[%d]", path, keyEnvironmentVars, i)
+
+		// Credential and api-key sources resolve their value server-side, so
+		// they are exempt from the value requirement.
+		source, _ := scalarString(mapValue(entry, keySource))
+		if source == credentialSource || source == sourceAPIKey {
```
wojtekwdr, line 486: [low] non-blocking: the new rule catches a missing `value` but not a wrong `source`. I checked that `source: dr-credentail` with a value passes Validate cleanly, and an unknown discriminator is exactly what the server's union should reject at deploy, which is the failure mode this rule exists to close. Checking `source` against the three known spellings would cover both.

internal/workload/manifest/schema.go
```diff
@@ -0,0 +1,216 @@
...
+func topLevelOrderedFields() []string {
+	var ordered []fieldRule
+
+	for _, rule := range fieldRules {
+		if rule.topLevelOrder > 0 {
+			ordered = append(ordered, rule)
+		}
+	}
+
+	sort.Slice(ordered, func(i, j int) bool {
```
wojtekwdr: [low] `sort.Slice` isn't stable, so two rules sharing a `topLevelOrder` would render keys in an unspecified order rather than failing anywhere obvious. Nothing guards against a duplicate today. `SliceStable` here, or a uniqueness assert in schema_test.go, makes that a deterministic mistake instead of a flaky one.

internal/workload/manifest/live_test.go
```diff
@@ -792,3 +833,1055 @@ func TestLive_ApplyRuntime(t *testing.T) {
...
+	assert.Contains(t, string(rendered), "imageBuildConfig")
+	assert.Contains(t, string(rendered), "source: provided")
+
+	parsed, err := Parse(rendered, "")
+	require.NoError(t, err)
+	require.NoError(t, parsed.Validate())
+}
+
+// TestLive_AutoscalingEnabledValuesSurviveStripping pins the boundary that
+// normalizeAutoscalingEnabled's removal must not cross: real autoscaling
```
wojtekwdr: [low] nit: `normalizeAutoscalingEnabled` was removed earlier on this branch, so this comment sends the next reader grepping for a function that isn't there. The `VAL-R3-006` code at the bottom has the same problem: it only means something to whoever was in that review thread.

wojtekwdr (approved):
LGTM with few minor comments

internal/workload/manifest/schema.go
ajalon1: That was a miss, thanks. will SliceStable.

internal/workload/manifest/live_test.go
ajalon1: got it, will remove

internal/workload/manifest/validate.go
ajalon1: need a better openapi contract :)

internal/workload/manifest/schema.go
ajalon1: Done (44571d8): SliceStable.

internal/workload/manifest/live_test.go
ajalon1: Removed (b038c68): both the normalizeAutoscalingEnabled mention and the VAL-R3-006 code.

internal/workload/manifest/schema_test.go
ajalon1: Partially right, and fixed (9a5632f): the fixtures now nest id/artifactId/endpoint/status/createdAt/updatedAt inside the spec/runtime subtrees, so assertions 1-6 genuinely pin the any-depth strip (flipping a serverManaged flag now fails the test — verified). The flags themselves stay: id/status are separately pinned by TestStripServerManaged_StripsNullKeys, and all six are load-bearing for nested leaks in real server docs — trimming them would drop real strip coverage.

PR #827, opened 2026-08-25
[RAPTOR-18979] fix(workload): guard every rollout starter, and stop the artifact type drifting for ever

wojtekwdr:
## CHANGES

- The in-flight rollout guard was only on the roll. Added it to the runtime-only resize, and to `up --lock` on an empty plan, which otherwise took a one-way lock on the version a rollout was moving off.
- A manifest with `type` under `artifact.spec` made every `up` mint a version and roll, for ever: the platform keeps the type on the artifact and strips it from the spec, so it always read as missing. The draft-reuse check had the same bug and piled up drafts. Both now normalise before comparing.
- A guard refusal now reads differently from a route that could not answer, the refusal drops a blank artifact field, and it no longer re-fetches a workload the run already has.
- Build and workload status checks fold case, matching the artifact one.
- A run refused before it attempted anything no longer reports the plan's action as something it did, and a rollout that starts and then fails no longer reports that it rolled. Worth knowing that `reportable()` emits the envelope on failures too, so anything reading `action` for intent rather than outcome now sees `unchanged` there; `plan.action` still carries what was wanted.

## TESTING

On staging: the manifest that rolled on every run now prints "Already up to date", both guards refuse a live in-flight replacement, and neither over-refuses once it settles.

That run also settled an open question on the ticket: `PATCH /workloads/{id}/settings/` merges rather than replaces, so the runtime-merge layer the ticket planned is not needed.

## NOTES

`GuardNoActiveReplacement` has no production caller now that the deploy paths use the variant skipping the existence probe. Kept as the entry point for a command taking a workload id from a flag; happy to drop it.

---

> [!NOTE]
> **Medium Risk**
> Changes core `up` deploy, lock, resize, and rollout guard behavior; mistakes could block deploys or allow queued swaps, though coverage is broad in unit tests.
> 
> **Overview**
> Fixes deploy loops and unsafe paths where the platform’s inconsistent status casing and artifact shape made `up` think work was always needed or let swaps proceed when they should not.
> 
> **Artifact type handling** hoists `type` out of `artifact.spec` before comparing manifest vs live (`hoistArtifactType` in load/plan/version, `liveArtifactType` in look). Spec diffs no longer treat a file-side type inside the spec as missing on the server; draft reuse uses the same normalization. Plan compares types with `ArtifactTypeOrDefault` on the live side so unstated live type reads as service and plan lines stay accurate.
> 
> **Replacement guarding** extends beyond roll: runtime-only retune and empty-plan `--lock` call `guardRollout`, which uses `RefuseActiveReplacement` (skips redundant workload existence GET after `Look`). Refusals wrap `ErrReplacementInFlight`; messages omit empty candidate artifact ids. Lookup failures get operation-specific “cannot tell whether…” wording distinct from in-flight refusals.
> 
> **Status classifiers** use case-insensitive checks for build terminal/error, replacement terminal/failed, and workload `stateFor`.
> 
> **Run result semantics** initialize `result.Action` to `unchanged` until apply succeeds (dry-run still reflects `plan.action`); early guard failures no longer report the planned action as done.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit b7ea4c278b5395e3d3098dec94b5cae7514f1817. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/internal/workload/replacement.go
+++ b/internal/workload/replacement.go
@@ -75,12 +75,15 @@ type Replacement struct {
 // IsTerminalReplacementStatus reports whether s is a status the replacement
 // will not progress from.
 //
-// The comparison is case-insensitive, for the reason the artifact's lock
-// check is: the platform's own docs disagree with themselves about the casing
-// of status enums. It is not a cosmetic difference here. Reading a settled
-// "COMPLETED" as still in progress makes GuardNoActiveReplacement refuse every
-// deploy for as long as the record lingers, with a message that calls a
-// finished replacement one in progress.
+// The comparison is case-insensitive. To be plain about why: this route has
+// only ever been seen answering in lower case, so the fold is hardening rather
+// than a fix for something observed. What justifies it is that the platform is
+// not consistent across resources, its build statuses being upper case where
+// these and the workload's are lower, and that the artifact's lock check
+// already folds for the same reason. The cost of being wrong is one-sided:
+// reading a settled "COMPLETED" as still in progress would refuse every deploy
+// for as long as the record lingers, with a message calling a finished
+// replacement one in progress.
 func IsTerminalReplacementStatus(s string) bool {
 	return IsFailedReplacementStatus(s) || strings.EqualFold(s, ReplacementStatusCompleted)
 }
@@ -106,7 +109,7 @@ func replacementURL(workloadID string) (string, error) {
 // answers the same way for a workload that does not exist, and the body
 // saying which never reaches this layer, because drapi.Get discards it. A
 // caller that has not already established the workload exists should use
-// GuardNoActiveReplacement, which pays for one extra request to tell the two
+// RefuseActiveReplacement, which asks the same question with the ambiguity
 // apart, rather than reading (nil, nil) as proof of anything.
 func GetActiveReplacement(workloadID string) (*Replacement, error) {
 	url, err := replacementURL(workloadID)
@@ -133,7 +136,7 @@ func GetActiveReplacement(workloadID string) (*Replacement, error) {
 //
 // It is not idempotent: called while a replacement is already in flight it
 // queues a second swap rather than rejecting, so callers guard with
-// GuardNoActiveReplacement first.
+// RefuseActiveReplacement first.
 //
 // Both artifacts must share draft or locked status; the platform rejects a
 // mismatch, which is why a locked workload has to have its successor locked
@@ -206,52 +209,28 @@ func CancelReplacement(workloadID string) error {
 	return nil
 }
 
-// GuardNoActiveReplacement refuses to proceed when workloadID already has a
-// replacement in flight, because POSTing another queues a second swap instead
+// RefuseActiveReplacement refuses to proceed when workloadID already has a
+// replacement in flight, because starting another queues a second swap instead
 // of erroring.
 //
 // How many times to call it follows from what sits between the check and the
 // call it protects. A roll mints and may lock an artifact in between, so it
-// runs the guard twice: once up front, so a doomed command fails before it
-// makes anything, and once immediately before StartReplacement, which is the
-// guard that actually holds, since the live state can change in between. A
-// caller with nothing in between runs it once, because there the up-front
-// check and the last moment before the call are the same moment.
+// runs this twice: once up front, so a doomed command fails before it makes
+// anything, and once immediately before the swap, which is the check that
+// actually holds, since the live state can change in between. A caller with
+// nothing in between runs it once, because there the two are the same moment.
 //
 // A refusal wraps ErrReplacementInFlight; everything else it returns is a
 // failure to find out. Callers that want to say which operation was stopped
 // should wrap the second kind and leave the first alone, since the refusal
 // already names the workload and the remedy.
 //
-// Being the up-front check is why it does not simply trust an empty answer.
-// The replacement route returns the same 404 for "nothing in flight" and for
-// "no such workload", so an unchecked guard would wave through exactly the
-// mistyped id it exists to catch, and the command would go on to create and
-// possibly lock an artifact before discovering the target was never real.
-// One extra request on the quiet path buys that, and the quiet path is about
-// to spend minutes rolling a workload.
-func GuardNoActiveReplacement(workloadID string) error {
-	active, err := GetActiveReplacement(workloadID)
-	if err != nil {
-		return err
-	}
-
-	if active == nil {
-		return confirmWorkloadExists(workloadID)
-	}
-
-	return refuseActive(workloadID, active)
-}
-
-// RefuseActiveReplacement is the same refusal without the existence probe, for
-// a caller that has already established the workload is there.
-//
-// That is the whole difference, and it is worth the second entry point because
-// the probe is not free: it is a second GET of a document such a caller has
-// already read, spent on the quiet path, where nothing is in flight. Reading a
-// 404 as "nothing to guard against" is only safe once something else has ruled
-// out "no such workload", so a caller that has not done that wants the guard
-// above instead.
+// It reads an empty answer as "nothing in flight", which is only safe because
+// every caller has already established the workload exists. The replacement
+// route answers the same 404 for "nothing in flight" and for "no such
+// workload", so a caller that has not settled that first would wave through
+// exactly the mistyped id it means to catch. `up` settles it in Look, before
+// the plan is even built.
 func RefuseActiveReplacement(workloadID string) error {
 	active, err := GetActiveReplacement(workloadID)
 	if err != nil {
--- a/internal/workload/replacement.go
+++ b/internal/workload/replacement.go
@@ -19,6 +19,7 @@ import (
 	"errors"
 	"fmt"
 	"net/http"
+	"strings"
 	"time"
 
 	"github.com/datarobot/cli/internal/config"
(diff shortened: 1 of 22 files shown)


internal/workload/up/version.go
```diff
@@ -380,6 +380,16 @@ func describes(loaded Loaded, artifactID string) bool {
...
 
+	// Both sides are normalised so the type is compared once, beside the spec.
+	// The file may write it in either place; the platform hoists it and answers
+	// beside the spec. Comparing the blocks as they arrive made the type read
+	// as a field one side was missing, so no leftover ever described the file
+	// and every attempt minted another one: the pile reuse exists to prevent.
+	// Normalising both rather than assuming a shape means neither a file nor a
+	// route that puts it in the other place can reopen that.
+	hoistArtifactType(want)
+	hoistArtifactType(doc)
```
damjandr: The reuse fix normalises placement but then hands the type to `Subset`'s `reflect.DeepEqual`, while `plan.go` compares it with `SameArtifactType` + `ArtifactTypeOrDefault`. So `type: Service` against a live `service`, or a live doc carrying no `type` at all, still refuses every leftover and piles drafts — the exact symptom this is meant to close.

internal/workload/up/run.go
```diff
@@ -170,8 +171,13 @@ func Run(opts Options) (Result, error) {
 		Status:     plan.State.String(),
 		Endpoint:   live.Endpoint,
 		ArtifactID: live.ArtifactID,
-		Action:     plan.Action(),
-		Locked:     live.Locked,
+		// Action records what this run did, not what the plan wanted, so it
+		// starts at "nothing" and each path below sets its own once it has
+		// acted. Seeding it from the plan reported a mutation for every run
+		// that failed before attempting one, including the two returns just
+		// below this. What was wanted stays readable under plan.action.
```
damjandr, line 178: The invariant stops at the seed: `roll.go:295` and `retune.go:81` set `Action` before the wait, so a rollout that starts and then ends `failed` still reports `action: "rolled"`. Also worth noting in the PR description that `reportable()` emits the envelope on error, so consumers reading `action` for intent now get `unchanged`.

internal/workload/replacement.go
```diff
@@ -227,6 +240,33 @@ func GuardNoActiveReplacement(workloadID string) error {
 		return confirmWorkloadExists(workloadID)
 	}
 
+	return refuseActive(workloadID, active)
+}
+
+// RefuseActiveReplacement is the same refusal without the existence probe, for
```
damjandr: Agreed with your note — `GuardNoActiveReplacement` now has zero non-test callers, so I would drop it and its tests. Re-adding it for a flag-driven command later is four lines.

internal/workload/replacement.go
```diff
@@ -73,20 +74,22 @@ type Replacement struct {
 
 // IsTerminalReplacementStatus reports whether s is a status the replacement
 // will not progress from.
+//
+// The comparison is case-insensitive, for the reason the artifact's lock
+// check is: the platform's own docs disagree with themselves about the casing
+// of status enums. It is not a cosmetic difference here. Reading a settled
```
damjandr: This asserts the platform disagrees with itself about casing, but the const block at `:29-49` says these statuses were confirmed live against staging — lowercase, including the undocumented `errored`. Has any route actually answered in another case, or is the fold speculative hardening?

internal/workload/settings.go
```diff
@@ -38,6 +38,13 @@ import (
 // only starts the work, and the caller has to follow the replacement to know
 // whether the new sizing was promoted. The workload's own status is no help,
 // since it stays running throughout.
+//
+// That makes this a replacement-starter, with StartReplacement's contract and
+// not a plain field update: called while a replacement is already in flight it
+// queues a second swap rather than rejecting, so callers guard with
+// GuardNoActiveReplacement first. Nothing in the live state substitutes for
```
damjandr: Names `GuardNoActiveReplacement`, but the only caller goes through `guardRollout`/`RefuseActiveReplacement`. Same drift at `replacement.go:136` and `replacement.go:80`.

internal/workload/up/retune.go
```diff
@@ -39,16 +39,37 @@ import (
...
 // and sending a subset would leave the reader guessing which half of the file
 // is now live.
+//
+// Being a rollout is also why it is guarded. The replacement route is
+// documented to queue a second swap rather than refuse one; that the settings
+// route does the same is an inference from it answering with a replacement,
+// not something confirmed against a live instance. Guarding is the safer of
+// the two guesses: if the route turns out to refuse on its own, the cost is a
+// request and a message less specific than the platform's, while not guarding
+// costs a swap nobody asked for. Worth settling on the same staging run that
```
damjandr: Stale relative to the PR description, which says that staging run happened and answered the merge question. Either say what the queue behaviour did, or say explicitly that it was not checked.

internal/workload/up/roll.go
```diff
@@ -253,21 +253,21 @@ func confirmRoll(workloadName string, opts Options) (bool, error) {
...
+	sizing json.RawMessage,
 	result Result,
 	opts Options,
 	report *reporter,
 ) (Result, error) {
 	// The guard that actually holds. The live state can have changed since
 	// the one at the top, and this is the last moment before a swap that
 	// cannot be taken back by refusing it.
-	if err := guardReplacementFn(workloadID); err != nil {
+	if err := guardRollout(workloadID, "the version serving keeps serving"); err != nil {
```
damjandr: "the version serving keeps serving" is true but incomplete — by here a draft has been minted and the project possibly re-pointed at it, so the reader does not learn they now own a leftover. The first guard's "nothing was built or rolled out" is exact.

internal/workload/up/load.go
```diff
@@ -64,30 +64,88 @@ func (l Loaded) WorkloadID() string {
...
+// ArtifactType below is what compares the type, from wherever the file put it,
+// and the wizard's own writer drops it from this block for the same reason.
 func (l Loaded) Spec() (map[string]any, error) {
 	payload, err := l.payload()
 	if err != nil {
 		return nil, err
 	}
 
 	artifact, _ := payload["artifact"].(map[string]any)
+	hoistArtifactType(artifact)
```
damjandr, line 83: Nit: `Spec()` throws away the hoisted map and only needs the `delete`, while `ArtifactType()` only needs the hoist — only `describes` wants both. One helper for one invariant is defensible, it is just more prose than `delete(spec, keyType)` would have needed.

internal/workload/up/roll.go
```diff
@@ -71,12 +71,12 @@ func roll(loaded Loaded, live Live, plan Plan, result Result, opts Options, repo
 	// A file that moved the sizing as well as the version rides both in on the
 	// same swap: the platform takes runtime alongside the artifact, so there is
 	// no second call and no window where one half is live and the other is not.
-	runtime, err := rollRuntime(loaded, plan)
+	sizing, err := rollRuntime(loaded, plan)
```
damjandr, line 74: Nit: `sizing` is the better name next to `json.RawMessage`, but the rename is pure diff noise in a fix PR.

internal/workload/up/run_test.go
```diff
@@ -415,6 +415,33 @@ func TestRun_LockWithNothingToDoStillLocks(t *testing.T) {
 	assert.True(t, result.Locked)
 }
```
damjandr, line 417: `guardRollout`'s doc argues hard about not paying a redundant round trip on the quiet path, but `install`'s default `guard` is a silent no-op, so `TestRun_NothingToDoWithoutLockLocksNothing` stays green if a guard call appears there. The `wiredRetune` comment makes exactly this case — worth applying to the no-lock empty-plan fixture too.

damjandr (approved):
LGTM, minor things below.

Two I could not anchor inline: `replacement.go:136` (`StartReplacement`) also still tells callers to use `GuardNoActiveReplacement`, and `roll.go:295` / `retune.go:81` set `Action` before the wait (see the `run.go` comment thread).

Verified while reading: status folding is complete outside `ParseWorkloadStatuses`, which lowercases first; `payload()` decodes per call so no `hoistArtifactType` mutation can reach the create request; `Extra` only walks name-keyed lists, so a file stating no type is safe; `plan.action` really is in the envelope.

internal/workload/up/version.go
wojtekwdr: Right, the hoist fixed where the type lives and left how it is compared. Both cases drift. It now goes through the same fold-and-default the plan uses and comes out of both maps before the walk.

internal/workload/up/run.go
wojtekwdr: Both set it after the wait agrees now; a detached run still records at the request, since nothing is waiting on it. Added a line about `reportable()` to the description.

internal/workload/replacement.go
wojtekwdr: Gone, with its tests.

internal/workload/replacement.go
wojtekwdr: Speculative, yes. This route has only ever answered lower case. What pushed me to fold it is that build statuses come back upper case while these and the workload's come back lower, so the platform is inconsistent across resources rather than within one, and the artifact lock check already folds for that reason. The comment says that now instead of implying I had seen it.

internal/workload/settings.go
wojtekwdr: Fixed, along with the two in replacement.go.

internal/workload/up/retune.go
wojtekwdr: The staging run answered the half it could: a settings-initiated replacement does show up on the replacement route, with a candidate id, so the guard reads what the resize writes. What a second PATCH does while one is in flight I never provoked, because the guard is what stops it. The comment says both now.

internal/workload/up/roll.go
wojtekwdr: Fair. It names the unpromoted version now.

internal/workload/up/load.go
wojtekwdr: Kept as one. Spec() and ArtifactType() disagreeing about the invariant is what caused the bug, and describes() wants both halves anyway, so I would rather they cannot come apart.

internal/workload/up/roll.go
wojtekwdr: It shadows the stdlib `runtime` package, which an earlier pass flagged. Agreed it is noise in a fix PR, but reverting now is more churn than leaving it.

internal/workload/up/run_test.go
wojtekwdr: Good point. That fixture fails if a guard call appears there now.

PR #831, opened 2026-08-26
[RAPTOR-19719] feat(workload): retry build trigger on retry-safe statuses, surface API error detail

adamalpi:
# RATIONALE

When `dr wl up` (or `dr artifact build`) triggers an image build, one slow moment in the platform's build service used to kill the whole deploy after the code was already synced, with an opaque error:

```
Error: HTTP error: 500 Internal Server Error (url: ...): body={"detail":"Internal server error"}
```

The Workload API now answers that situation with an explicit contract ([RAPTOR-19719](https://datarobot.atlassian.net/browse/RAPTOR-19719), [datarobot/workload-api#1187](https://github.com/datarobot/workload-api/pull/1187)): a **504** guarantees nothing was recorded on the artifact and a retry is safe, and its existing **503** for dependency blips also says "please retry". This PR makes the CLI use that contract instead of failing the deploy, and makes every API error lead with the server's own explanation instead of a raw JSON dump.

## CHANGES

- `TriggerArtifactBuild` retries on exactly 503 and 504 (2s/5s backoff, three attempts total). A 504 is only retried when it carries the platform's JSON `detail` body — a proxy's own 504 (HTML/plain) means the platform may still be processing, where a re-POST could start a duplicate build. A plain 500 fails fast for the same reason.
- `drapi.ErrFromResp` now lifts FastAPI-style `{"detail": ...}` bodies into the (already existing, never populated) `HTTPError.Detail` field, so users read `HTTP 504 Gateway Timeout: ...it is safe to retry.` instead of a JSON blob. Validation-error arrays are re-encoded so field-level messages survive; non-JSON bodies keep the `body=` form with the URL.
- Tests for retry-then-succeed, no-retry-on-500, no-retry-on-proxy-504, retry-budget exhaustion, and the three `ErrFromResp` shapes; one existing assertion updated from verbatim-body to message-content.
- **Separate first commit: repairs `main`.** `specMatches` (`internal/workload/up/version.go:419`) was changed to return `(bool, error)` in the current `main` tip but one early `return false` was left behind, so the `up` package does not compile — every vet/test/build job has been failing on `main` and on any PR merged with it since. The fix returns `false, nil` (a type mismatch is an ordinary no-match, not an error).

## PR Automation

> [!IMPORTANT]
>

## RELATED

- Backend contract this builds on: [datarobot/workload-api#1187](https://github.com/datarobot/workload-api/pull/1187)
- Jira: [RAPTOR-19719](https://datarobot.atlassian.net/browse/RAPTOR-19719)

---

> [!NOTE]
> **Medium Risk**
> Retry policy on build POSTs could duplicate builds if 504/503 classification is wrong; error parsing changes behavior for all drapi HTTP errors, not only builds.
> 
> **Overview**
> Makes artifact build triggers more resilient and API failures easier to read during `dr wl up` / `dr artifact build`.
> 
> **`TriggerArtifactBuild`** now retries up to three times (2s then 5s backoff) when the platform returns **503** or a **504** that includes a FastAPI JSON `detail` (the “safe to retry / nothing recorded” contract). It does **not** retry proxy HTML 504s, plain **500**s, or other statuses, to avoid duplicate builds when the server may have already accepted the POST.
> 
> **`drapi.ErrFromResp`** parses `{"detail": ...}` bodies into **`HTTPError.Detail`** so errors show the server message instead of `body={...}`; validation arrays are preserved as JSON, and non-JSON bodies still use the `body=` form.
> 
> Also fixes a **`specMatches`** compile break in `internal/workload/up/version.go` (`return false, nil` on artifact type mismatch).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 07d039734844f421025e9bfd2734b932b6eab758. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/internal/drapi/helpers.go
+++ b/internal/drapi/helpers.go
@@ -15,7 +15,6 @@
 package drapi
 
 import (
-	"encoding/json"
 	"fmt"
 	"io"
 	"net/http"
@@ -45,49 +44,17 @@ func EndpointURL(path string, query url.Values) (string, error) {
 // up to 512 bytes of the response body for context. It always closes
 // resp.Body, so callers must not read it after calling this function.
 //
-// When the body is a FastAPI-style `{"detail": ...}` document, the detail is
-// lifted into HTTPError.Detail so the user reads the server's own sentence
-// instead of a raw JSON dump; anything else keeps the body= form, whose URL
-// is what makes an unstructured failure diagnosable.
+// The body is carried raw on HTTPError.Body, uninterpreted: drapi serves
+// every DataRobot API the CLI talks to, and they do not agree on an error
+// envelope, so reading meaning into the bytes is a caller's job (see
+// workload/apiclient for the Workload API's).
 func ErrFromResp(resp *http.Response, requestURL string) error {
 	defer func() { _ = resp.Body.Close() }()
 
 	body, _ := io.ReadAll(io.LimitReader(resp.Body, 512))
-	if detail := ErrorDetail(body); detail != "" {
-		return &HTTPError{StatusCode: resp.StatusCode, URL: requestURL, Detail: detail}
-	}
-
 	if len(body) > 0 {
-		return fmt.Errorf("%w: body=%s", &HTTPError{StatusCode: resp.StatusCode, URL: requestURL}, body)
+		return fmt.Errorf("%w: body=%s", &HTTPError{StatusCode: resp.StatusCode, URL: requestURL, Body: body}, body)
 	}
 
 	return &HTTPError{StatusCode: resp.StatusCode, URL: requestURL}
-}
-
-// ErrorDetail pulls the "detail" field out of a FastAPI-style JSON error
-// body, or "" when the body is not such a document. Non-string details
-// (e.g. validation error arrays) are re-encoded as JSON rather than dropped:
-// they carry the field-level messages the caller needs.
-//
-// This is the one place that knows the server's error envelope; callers that
-// want a different fallback (e.g. the raw body) layer it on the "" return.
-func ErrorDetail(body []byte) string {
-	var payload struct {
-		Detail any `json:"detail"`
-	}
-
-	if err := json.Unmarshal(body, &payload); err != nil || payload.Detail == nil {
-		return ""
-	}
-
-	if detail, ok := payload.Detail.(string); ok {
-		return detail
-	}
-
-	encoded, err := json.Marshal(payload.Detail)
-	if err != nil {
-		return ""
-	}
-
-	return string(encoded)
 }
--- a/internal/drapi/helpers.go
+++ b/internal/drapi/helpers.go
@@ -15,6 +15,7 @@
 package drapi
 
 import (
+	"encoding/json"
 	"fmt"
 	"io"
 	"net/http"
@@ -43,13 +44,47 @@ func EndpointURL(path string, query url.Values) (string, error) {
 // ErrFromResp wraps a non-2xx *http.Response into a *HTTPError, capturing
 // up to 512 bytes of the response body for context. It always closes
 // resp.Body, so callers must not read it after calling this function.
+//
+// When the body is a FastAPI-style `{"detail": ...}` document, the detail is
+// lifted into HTTPError.Detail so the user reads the server's own sentence
+// instead of a raw JSON dump; anything else keeps the body= form, whose URL
+// is what makes an unstructured failure diagnosable.
 func ErrFromResp(resp *http.Response, requestURL string) error {
 	defer func() { _ = resp.Body.Close() }()
 
 	body, _ := io.ReadAll(io.LimitReader(resp.Body, 512))
+	if detail := errorDetail(body); detail != "" {
+		return &HTTPError{StatusCode: resp.StatusCode, URL: requestURL, Detail: detail}
+	}
+
 	if len(body) > 0 {
 		return fmt.Errorf("%w: body=%s", &HTTPError{StatusCode: resp.StatusCode, URL: requestURL}, body)
 	}
 
 	return &HTTPError{StatusCode: resp.StatusCode, URL: requestURL}
+}
+
+// errorDetail pulls the "detail" field out of a FastAPI-style JSON error
+// body, or "" when the body is not such a document. Non-string details
+// (e.g. validation error arrays) are re-encoded as JSON rather than dropped:
+// they carry the field-level messages the caller needs.
+func errorDetail(body []byte) string {
+	var payload struct {
+		Detail any `json:"detail"`
+	}
+
+	if err := json.Unmarshal(body, &payload); err != nil || payload.Detail == nil {
+		return ""
+	}
+
+	if detail, ok := payload.Detail.(string); ok {
+		return detail
+	}
+
+	encoded, err := json.Marshal(payload.Detail)
+	if err != nil {
+		return ""
+	}
+
+	return string(encoded)
 }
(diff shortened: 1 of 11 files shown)


internal/drapi/helpers.go
```diff
@@ -43,13 +44,47 @@ func EndpointURL(path string, query url.Values) (string, error) {
...
+// When the body is a FastAPI-style `{"detail": ...}` document, the detail is
+// lifted into HTTPError.Detail so the user reads the server's own sentence
+// instead of a raw JSON dump; anything else keeps the body= form, whose URL
+// is what makes an unstructured failure diagnosable.
 func ErrFromResp(resp *http.Response, requestURL string) error {
 	defer func() { _ = resp.Body.Close() }()
 
 	body, _ := io.ReadAll(io.LimitReader(resp.Body, 512))
+	if detail := errorDetail(body); detail != "" {
+		return &HTTPError{StatusCode: resp.StatusCode, URL: requestURL, Detail: detail}
```
wojtekwdr: [medium] `HTTPError.Error()` only prints the URL when `Detail` is empty, so every POST, PATCH and DELETE that comes back with a `detail` now loses the endpoint from its message. A `dr wl delete` failure reads `HTTP 500 Internal Server Error: Internal server error` with nothing saying which resource it was. GET never populates Detail, so the two error shapes diverge as well.

internal/drapi/helpers.go
```diff
@@ -43,13 +44,47 @@ func EndpointURL(path string, query url.Values) (string, error) {
...
 	}
 
 	return &HTTPError{StatusCode: resp.StatusCode, URL: requestURL}
 }
+
+// errorDetail pulls the "detail" field out of a FastAPI-style JSON error
+// body, or "" when the body is not such a document. Non-string details
+// (e.g. validation error arrays) are re-encoded as JSON rather than dropped:
+// they carry the field-level messages the caller needs.
+func errorDetail(body []byte) string {
```
wojtekwdr: [low] `extractErrorDetail` in `internal/pipeline/pipeline.go:422` already does this same lift, differing only in the fallback (raw body vs empty string). Two copies of the same wire contract means the next change to the server's error envelope has to be found and fixed twice. pipeline already imports drapi if you want to go the other direction.

internal/workload/up/version.go
```diff
@@ -416,7 +416,7 @@ func specMatches(loaded Loaded, doc workload.Document) (bool, error) {
 	// attempt refuses that one too. It is the same question the plan asks, so
 	// it is asked the same way, folded and with only the live side defaulted.
 	if !sameArtifactTypeIn(want, doc) {
-		return false
+		return false, nil
```
wojtekwdr, line 428: [low] Nothing exercises this branch. Coverage reports zero hits for it and there's no test for `specMatches` or `sameArtifactTypeIn` anywhere in the package, so a type mismatch returning `true` would ship green the same way the missing return value did.

wojtekwdr (approved):
LGTM with minor comments

internal/drapi/helpers.go
adamalpi: URL dropped from detail-bearing errors — HTTPError.Error() now appends (url: ...) in both shapes, so a failed POST/PATCH/DELETE names its endpoint again.

internal/drapi/helpers.go
adamalpi: fixed

internal/workload/up/version.go
adamalpi: fixed

internal/drapi/helpers.go
```diff
@@ -43,13 +44,50 @@ func EndpointURL(path string, query url.Values) (string, error) {
 // ErrFromResp wraps a non-2xx *http.Response into a *HTTPError, capturing
 // up to 512 bytes of the response body for context. It always closes
 // resp.Body, so callers must not read it after calling this function.
+//
+// When the body is a FastAPI-style `{"detail": ...}` document, the detail is
+// lifted into HTTPError.Detail so the user reads the server's own sentence
+// instead of a raw JSON dump; anything else keeps the body= form, whose URL
+// is what makes an unstructured failure diagnosable.
 func ErrFromResp(resp *http.Response, requestURL string) error {
```
ajalon1, line 51: ErrFromResp has 6 callers: post.go, patch.go, delete.go, and three in internal/drapi/filesapi -- which fair, you guys already own. :)

The {"detail": ...} envelope works for FastAPI, most of our newer microservices, but not at all for the Public API, which is drflask based. drflask serializes every error as {"message": exc.detail} (the attribute name is a trap). Only one controller in the Public API emits `detail`, while the other ... 468 emit `message`.

If ErrFromResp is in an `internal/workload/apiclient` module? Then yes, perfect. Maybe even for any FastAPI based service.

internal/pipeline/pipeline.go
```diff
@@ -420,25 +420,8 @@ func buildMultipartBody(filePath string, fields map[string]string) (*bytes.Buffe
...
-			return detail
-		default:
-			encoded, encErr := json.Marshal(detail)
-			if encErr == nil {
-				return string(encoded)
-			}
-		}
+	if detail := drapi.ErrorDetail(body); detail != "" {
+		return detail
 	}
```
ajalon1, line 442: Good idea reusing this, just not for all of `drapi` (or even all of pipelines).

Also: it gives HTTPError.Detail a second producer that sets it to the entire unbounded response body (pipeline.go:366 reads with io.ReadAll, :427 falls back to string(body)), so Detail != "" means "the server explained itself" in one layer and "the response was non-empty" in the other.

ajalon1 (changes requested):
Nice @adamalpi -- I've been wanting to tighten up `internal/drapi` for a while. But the error handling is very specific to the Workload API build endpoints and doesn't translate well to the Public API, which `internal/drapi` is obligated to also handle.

`internal/drapi` should be able to handle any DataRobot-related API -- which by my count includes starlette, FlaskAPI, FastAPI and a number of other handrolled situations.

I'd suggest instead a separate `internal/workload/apiclient` module so that we don't touch pipelines code or, really, any of the Public API calls (auth, version, llm_gateway, etc). Then you're free to do whatever you like. Pipelines does this with the `internal/pipeline/` module.

ajalon1:
Separately, I had Claude review the builds route in workload-api, and it seems like the retry-set and the status set there have drifted?

🤖 routes/builds.py raises 404/422/502/503 and never 504, so the 504 branch (and the Detail != "" gate on it) is guarding a status this route can't produce.
🤖 services/artifact.py:470 and :519 raise ArtifactUpdateFailed("Concurrent modification detected; please retry.") → 422 at builds.py:99. That's the one response that explicitly asks for a retry, and it's the one we don't retry. The updated_at pre-check at artifact.py:466 means a 503 retry storm makes it more likely.
🤖 Both 502 paths (:120 storage write, :136 IBS refusal) mean the submission never landed — at least as safe to retry as 503.

Since y'all own both sides: rather than the CLI trying to guess about the response, can we just retry the statuses your route already claims?

🤖 And artifact.py:459 says prior builds are stopped best-effort — worth reflecting that in the timeout message, which currently promises a plain supersede.

adamalpi:
@ajalon1 thanks for the feedback! 

I've tried to reduce the footprint of changes in drapi, please have another look!

adamalpi:
Regarding the 504, I added a PR this morning, I am keeping that in sync!

ajalon1 (approved):
:shipit:

PR #833, opened 2026-08-26
fix(workload): stop an agent redraft forking its own lineage

wojtekwdr:
# RATIONALE

Every redraft of an **agent** artifact silently forked its own lineage.

`specMatches` normalises both artifact blocks in place, because the walk that follows compares by exact equality and cannot answer the type question correctly. The live block it normalises belongs to the caller: `ensureArtifact` hands the same document to `redraftRepository` on the very next line, which reads the type off it. By then the type has been taken away, so an agent read as a service, the comparison against the file's `agent` failed, and the run concluded the file had changed the artifact's kind. That is a real reason to start a fresh repository, so the replacement opened one instead of joining the lineage it is the next version of.

Services were unaffected, since the default the missing type falls back to happens to be the right answer for them. Every fixture in this package is a service, which is why a green suite never saw it.

Found while reviewing the change stacked on top of this one; the fix is independent of it.

## CHANGES

- `internal/workload/up/version.go`: compare against a detached copy. It copies the two maps the normalisation writes to, the top level and the spec, and shares everything deeper, which nothing below reaches.
- A regression test with an agent artifact, which fails without the fix.

### DIFF
--- a/internal/workload/up/run_test.go
+++ b/internal/workload/up/run_test.go
@@ -2123,6 +2123,46 @@ func TestRun_ReplacementForADivergedSpecJoinsTheSameLineage(t *testing.T) {
 	assert.Equal(t, "cat1", *tr.savedCfg.CatalogID)
 }
 
+// The same lineage question asked of an agent, which is what makes it a
+// regression test rather than a duplicate.
+//
+// The comparison that decides whether the linked draft still matches used to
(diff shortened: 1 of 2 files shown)


wojtekwdr:
Superseded: #832 now carries both `version.go` commits off `main`, so this one is empty. GitHub would not let me retarget it, since it is linked into a stack.

PR #834, opened 2026-08-26
[RAPTOR-19686] feat(workload): deploy onto a settling or stopped workload in one run

wojtekwdr:
# RATIONALE

RAPTOR-19686. `dr workload up` refused three live states in a row, and each refusal named the command that led to the next one:

```
errored   -> "check 'dr workload logs' first"          (remedy: stop it)
settling  -> "wait for it to finish, then deploy"      (after the stop)
stopped   -> "start it with 'dr workload start' and deploy again"
```

The errored refusal is the perverse one: an empty plan never reaches it, so it only ever fired when the user already had the fix in hand.

Two of these states are dealt with here. A workload still moving is waited out and re-read, so the plan is built against where it landed rather than against a state that is about to change. A stopped one is started and then reconciled, which is one command whether the file asks for a start alone or for a start and a new version. Errored becomes deployable in the change that follows this one.

Reversing the stopped row is deliberate: it undoes a position amended into RAPTOR-18979 in August, which argued that coming up on the old version and then failing was the worst of both. It is not. The user asked for the workload to be up, so a failed roll that leaves it up on the old version is exactly what a failed roll against a running workload leaves behind, while the refusal left it switched off and asked for two more commands.

## CHANGES

**The two states.** `awaitSteady` waits before the plan is built and then re-reads, so the status and the spec come from one snapshot. A stopped workload is started first, and from that point the run is indistinguishable from one against a running workload.

**Ordering, which the start forces.** A start is a mutation, so everything that can still refuse moves ahead of it: the credential check, the in-flight rollout guard, and the question a locked production version has to answer. A "no" arriving after the workload had been switched on would be a refusal that had already changed something, under a message saying nothing had.

**A new wait primitive.** `WaitForSteadyWorkload` waits for a transition to land rather than for a workload to come up. That is the opposite question `WaitForWorkload` answers, and the reason a `stopping` workload handed to it polls to its timeout. Both now share one loop and layer their own verdict on top. A test pins the two status classifications together across the package boundary.

**The lock moved out of `settle`** into its own step, so a start that is about to be rolled off cannot lock the version it is about to replace. Locking is one way, so that distinction is not a tidiness one.

**Honesty fixes on the paths this opens.** An errored or terminated workload whose file matches no longer reports "Already up to date" and exit 0; a preview of a settling workload no longer claims nothing differs, since what the deploy does depends on where the transition lands; the plan header carries the platform's own status, so a suspended workload is not announced as stopped and then refused for being suspended; and the draft warning is no longer suppressed on a failed run that had itself started the workload and its eight hour clock.

## NOTES

Three API behaviours this design assumes are undocumented and want a staging run before the follow-up ships: whether a replacement, a stop and a settings PATCH are each accepted on an errored workload.

Known and deliberately left alone, rather than folded in here: `--poll-timeout` is a per wait budget across several waits, and this adds one more; the start's boot wait blocks an image build that needs nothing from the workload; and `WaitForWorkload`, `WaitForBuild` and `WaitForReplacement` are three copies of one poll loop.

### DIFF
--- a/cmd/workload/up/cmd.go
+++ b/cmd/workload/up/cmd.go
@@ -142,6 +142,12 @@ allocation, is applied in place instead. Nothing is built and no version is
 made, because what the workload runs has not changed. A deploy that moves both
 sends the sizing with the rollout, so the new version comes up with it.
 
+A workload that is not ready to be deployed onto is dealt with rather than
+refused. One still starting or stopping is waited out and then re-read, so the
+plan is built against where it landed. A stopped one is started and then
+reconciled in the same run, which is one command whether the file asks for a
+start alone or for a start and a new version.
+
 Examples:
   dr workload up
   dr workload up --dry-run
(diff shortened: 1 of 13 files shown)


wojtekwdr:
Recreating against `wojtekw/fix-main-build`, which now carries both `version.go` fixes. The base branch here cannot be retargeted once GitHub has linked a PR into a stack.

PR #837, opened 2026-08-26
[RAPTOR-19602] fix(workload): stop a refused deploy announcing work it will not do

cdevent:
# RATIONALE

[RAPTOR-19602](https://datarobot.atlassian.net/browse/RAPTOR-19602)

> Rebased onto #836. See **Rebase notes** at the bottom for what that changed.

When `up` refuses a live state it cannot deploy onto, it has already printed a full plan
of the work it is about to decline, and then prints three follow-up commands underneath
the error:

```
old-name (68b0term), terminated

  + artifact   new version, 2 spec changes
      containerGroups: [1 item(s)]
      type: service
  ~ runtime    containerGroups: [1 item(s)]

    Workload endpoint: https://…/68b0term0000000000000002/

Next:
  dr workload logs 68b0term…    what the container is saying
  dr workload status 68b0term…  whether it is healthy
  dr workload stop 68b0term…    switch it off, keeping the version
Error: workload old-name is terminated, which cannot be undone…
```

Two separate defects, both pre-existing:

1. `Run` rendered the plan before `apply` ever asked `deployable` whether the state could
   take one, so a refused state announced a rollout it was about to decline. #836 gave
   those states a `!` stateLine naming what is wrong with the workload, which helps, but
   the drift below it still reads as work about to happen and the refusal still arrives
   last.
2. `nextSteps` guarded only on `result.WorkloadID == ""`, so any refusal with an id got
   all three commands. The neighbouring endpoint tick already handles failure — *"a tick
   above an error reads as though both happened"* — and `nextSteps` never got the same
   treatment. `dr workload stop` on a terminated workload is meaningless.

The drift itself is not useless information, which is what made this a design question
rather than a bug to delete: someone who has just been refused is usually about to ask
what the deploy would have been. So the block stays and stops reading as an announcement.

## CHANGES

**The refusal is decided before the plan is printed** — `internal/workload/up/run.go`

`announce` asks `refusal` and hands the answer to `Render`. Only a non-empty plan is
judged; `--lock` on an empty one still asks for itself in `lockOnly`.

Settling is deliberately **not** asked. #836 waits a moving workload out before the plan
is built, so the only run that reaches a plan still settling is a dry run, which does not
wait — refusing it would contradict the note printed two lines above saying a deploy would
plan against where it lands. `deployable` keeps that branch as the backstop it is
documented to be, which is also why `apply` keeps its own call: it is now the only thing
answering for a workload that was steady when it was read and is moving again by the time
it is acted on.

**The plan says it is a description** — `internal/workload/up/render.go`

New `Summary.Refused` puts one line above the block. It says only that the rest will not
be applied: the `!` stateLine directly beneath already names what is wrong with the
workload, and the error below the block already names the remedy, so a note explaining
either would be the third telling on one screen.

```
gpt-oss-20b-vllm (68b0c1d2), terminated

  Nothing below will be applied; it is what differs, not what is about to happen.

  ! workload   terminated, and it still holds its name and artifact
  ~ lock       the running version is locked, so a new one is created and locked to match…
  + artifact   new version, 1 spec change
      containerGroups[default].containers[vllm-server].port: 8000 -> 9000
Error: workload gpt-oss-20b-vllm is terminated, which cannot be undone…
```

**The follow-ups suit the state** — `cmd/workload/up/cmd.go`

`nextSteps` splits into `followUps`:

| run | list |
| --- | --- |
| terminated refusal | nothing — none of the three apply, and the error already names `dr workload delete` |
| errored / timeout / failed roll | `logs`, `status` — right for all three, and the only place the id is attached to the command |
| success | unchanged: `logs`, `status`, and `stop` or `up --lock` |

`stop` is never offered on a failure: it is not a next step for a deploy that did not
land. Neither is `--lock` — since #836 a deploy onto a stopped workload starts it first,
so a run that fails after that really has put a draft on the air (`draftIsServing` says
so and the warning prints), but locking a version this run could not finish is not the
remedy for it, and `draftWarning` names the command inline for anyone who decides
otherwise.

The endpoint on a failed run is untouched, deliberately — losing the id and endpoint is
how a deploy becomes unfindable, and only its success tick is dropped.

## NOTES

One behaviour change beyond the report, intentional:

- **A refused `--dry-run` is now a failure rather than a preview.** The question the flag
  asks is "what would this deploy", and the honest answer is that it would be turned
  away, which a CI job has to be able to see in the exit code. The disclaimed plan is
  still printed, so the answer is not just the refusal. Its "Dry run: nothing was
  changed." footer is dropped on a failed run for the same reason the tick is: sitting
  above an error, it reads as a preview that went to plan.

A dry run onto a **settling** workload is untouched — still a preview, still exit 0, with
or without drift. There is a test pinning that, since it is the one thing the pre-flight
had to be taught not to do.

## TESTING

`task test` and `task lint` (GOOS=linux, darwin, windows) both clean.

New tests, driving terminated and errored separately at both layers since they differ in
which follow-ups make sense:

- `internal/workload/up`: `TestRun_RefusedStateDescribesItsDriftRatherThanAnnouncingIt`
  (both states; asserts the note precedes the change lines), `TestRun_AnAppliedPlanIsStillAnnounced`,
  `TestRun_RefusedDryRunIsAFailureNotAPreview`, and
  `TestRun_DryRunOnASettlingWorkloadWithDriftIsStillAPreview` guarding the settling
  exemption. `TestRender_RefusedPlanIsDescribedRatherThanAnnounced` and
  `TestRender_AppliedPlanCarriesNoRefusedNote` cover the renderer directly.
- `cmd/workload/up`: `TestCmd_TerminatedRefusalOffersNoFollowUps`,
  `TestCmd_ErroredRefusalKeepsLogsAndStatusButNotStop`,
  `TestCmd_AnyFailureKeepsLogsAndStatusButNotStop`,
  `TestCmd_FailureAfterAStartWarnsButOffersNoLock` (the #836 interaction),
  `TestCmd_TerminatedRefusalStillReportsTheEndpoint` (the endpoint must not regress),
  `TestCmd_FailedDryRunDropsTheFooter`.

## RELATED

- RAPTOR-19526 — found during its manual testing; `Run`'s render-then-apply order predates
  it, so this is pre-existing.
- RAPTOR-19686 (#836) — rebased onto it; overlapping surface, complementary fix.
- RAPTOR-19537 — making `up` legible to agents. A refusal that prints a plan plus three
  commands plus an error is exactly the shape that made an agent misread the command.

---

> [!NOTE]
> **Low Risk**
> CLI output and exit-code behavior for refused/failed `workload up` runs only; no auth, data, or deploy API contract changes beyond clearer pre-apply refusal messaging.
> 
> **Overview**
> Fixes misleading `dr workload up` output when a deploy is refused or fails: the plan no longer reads like work that will run, and the **Next:** hints match what actually helps.
> 
> **Plan framing** — Deployability is checked in `announce` before the plan is printed (via `refusal` / `deployable`, with settling still exempt for dry-run previews). When the workload cannot take the deploy, `Summary.Refused` adds a line above the drift: *Nothing below will be applied; it is what differs, not what is about to happen.*
> 
> **Command shell** — `nextSteps` now uses `followUps` with a `failed` flag: **terminated** refusals show no follow-ups; **failed** runs keep `logs` and `status` only (no `stop` or `--lock`); successful paths are unchanged. Failed **dry-run** skips the “Dry run: nothing was changed.” footer so it does not sit above the error. A refused **dry-run** now exits with error (CI can detect “would be turned away”) while still printing the disclaimed plan.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit aab890468f2e128fed072217d5ef4f35bd2ad0f7. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/cmd/workload/up/cmd.go
+++ b/cmd/workload/up/cmd.go
@@ -33,6 +33,7 @@ import (
 	"github.com/datarobot/cli/internal/misc/reader"
 	"github.com/datarobot/cli/internal/outputformat"
 	"github.com/datarobot/cli/internal/telemetry"
+	"github.com/datarobot/cli/internal/workload"
 	"github.com/datarobot/cli/internal/workload/manifest"
 	"github.com/datarobot/cli/internal/workload/up"
 	"github.com/datarobot/cli/tui"
@@ -417,6 +418,16 @@ func render(cmd *cobra.Command, f flags, format outputformat.OutputFormat, resul
 	draft := draftIsServing(f, result, failed)
 
 	if f.dryRun {
+		// A dry run that failed says nothing of its own. The error explains
+		// itself, and the footer's calm "nothing was changed" sitting above it
+		// reads as a preview that went to plan; a refused deploy is exactly the
+		// case, since the plan it just printed is disclaimed rather than
+		// promised. draftIsServing already withholds the warning for the same
+		// reason, so there is nothing left to print here either.
+		if failed {
+			return nil
+		}
+
 		fmt.Fprintln(cmd.ErrOrStderr(), "\nDry run: nothing was changed.")
 		draftWarning(cmd.ErrOrStderr(), draft, true)
 
@@ -492,28 +503,63 @@ func draftWarning(w io.Writer, draft, planned bool) {
 }
 
 // nextSteps lists what to run against the workload this deploy just touched,
-// on stderr so the endpoint on stdout stays pipeable. Nothing is printed for a
-// run that produced no workload: a list of commands that need an id is no help
-// to someone who has not got one.
+// on stderr so the endpoint on stdout stays pipeable. Nothing is printed when
+// there is nothing worth running: see followUps.
+func nextSteps(w io.Writer, result up.Result, dir string, draft, failed bool) {
+	steps := followUps(result, dir, draft, failed)
+	if len(steps) == 0 {
+		return
+	}
+
+	fmt.Fprintf(w, "\n%s\n", tui.HintStyle.Render("Next:"))
+
+	for _, step := range steps {
+		fmt.Fprintf(w, "  %s  %s\n",
+			tui.InfoStyle.Render(step[0]), tui.HintStyle.Render(step[1]))
+	}
+}
+
+// followUps is what is worth running against this workload now, empty when
+// nothing is. A run that produced no workload is the first such case: a list of
+// commands that all name a workload is no help to someone who has not got one.
 //
 // The commands carry no id. A successful deploy leaves the manifest bound, so
 // they resolve from the project directory, and --dir carries a deploy that ran
 // somewhere else.
 //
-// A failed run is the exception, and takes the id instead. One of its shapes
-// is a workload created whose id could not be written back, where the binding
-// the bare commands need is exactly what is missing; printing them bare would
-// hand the reader three commands that cannot run. It is also what keeps
-// reportable's promise that a deploy which failed late is still findable.
+// A failed run is the exception, and takes the id instead. One of its shapes is
+// a workload created whose id could not be written back, where the binding the
+// bare commands need is exactly what is missing; printing them bare would hand
+// the reader commands that cannot run. It is also what keeps reportable's
+// promise that a deploy which failed late is still findable.
+//
+// A failed run also loses two of the lines, for the reason the endpoint's tick
+// is dropped on the same run. 'stop' is not a next step for a deploy that did
+// not land, and neither is --lock: a deploy onto a stopped workload starts it
+// before it rolls, so a run that fails after that really has put a draft on the
+// air, but locking a version this run could not finish is not the remedy for
+// it, and draftWarning names the command inline for anyone who decides
+// otherwise. --lock is in any case the one line that could not be made to name
+// the workload, since it takes no id, and on a failed run the manifest may hold
+// no binding for it to resolve. What survives is logs and status, which are the
+// right pair for an errored workload, a wait that timed out and a rollout that
+// never completed, and which carry the id that the errors naming those same
+// commands do not.
+//
+// A terminated workload gets nothing. None of the three mean anything against
+// something that is not coming back, and the refusal already names
+// 'dr workload delete', which is the only command that helps. Matched on the
+// status rather than through IsWorkloadErrorStatus, which also covers errored:
+// errored is exactly where logs and status earn their place.
 //
 // A draft deploy trades the stop line for --lock, and puts it first so it sits
 // directly under the warning that explains why it is there. Someone who wants
-// to switch a workload off goes looking for the command; someone whose
-// workload switches itself off does not know there is anything to look for,
-// and this list is the one place they will read either way.
-func nextSteps(w io.Writer, result up.Result, dir string, draft, failed bool) {
+// to switch a workload off goes looking for the command; someone whose workload
+// switches itself off does not know there is anything to look for, and this
+// list is the one place they will read either way.
+func followUps(result up.Result, dir string, draft, failed bool) [][2]string {
 	if result.WorkloadID == "" {
-		return
+		return nil
 	}
 
 	at := manifest.DirFlag(dir)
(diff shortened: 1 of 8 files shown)


internal/workload/up/run.go
```diff
@@ -677,6 +682,61 @@ func credentialRefs(loaded Loaded, plan Plan) []manifest.CredentialRef {
...
+// last line.
+//
+// Settling is deliberately not asked, though deployable refuses it. A moving
+// workload is waited out before the plan is built, so the only run that reaches
+// a plan still settling is a dry run, which does not wait and has nothing to
+// refuse: it previews where the deploy would land rather than acting on where
+// the workload is. deployable keeps that branch as the backstop it is
+// documented to be, at the point of action.
+func refusal(loaded Loaded, live Live, workloadName string) error {
+	if live.State == StateSettling {
```
ajalon1: [P3] The doc comment claims "the only run that reaches a plan still settling is a dry run," but a narrow TOCTOU race in `awaitSteady`'s re-Look can hand a non-dry-run a settling `live`. Not a regression (same behavior as before the PR), just an overstated invariant.

cmd/workload/up/cmd.go
```diff
@@ -446,7 +457,7 @@ func render(cmd *cobra.Command, f flags, format outputformat.OutputFormat, resul
 	}
 
 	draftWarning(cmd.ErrOrStderr(), draft, false)
-	nextSteps(cmd.ErrOrStderr(), result, draft)
+	nextSteps(cmd.ErrOrStderr(), result, draft, failed)
```
ajalon1: [P3, pre-existing] `draftIsServing` checks only `result.Action` in its failed branch, so a workload that started then terminated still gets the "running a draft artifact" warning. This PR's `followUps` terminated check makes the mismatch more visible — next steps are gone but the draft warning remains.

cmd/workload/up/cmd.go
```diff
@@ -497,45 +508,78 @@ func draftWarning(w io.Writer, draft, planned bool) {
...
+	// as a build id, and on a roll the two deliberately belong to different
+	// artifacts: BuildID is the candidate's, ArtifactID stays on the version
+	// still serving until the swap lands. Pairing them would send the reader to
+	// a 404 at exactly the moment they need the logs. The error from a failed
+	// build already names the right pair.
+	logs := [2]string{"dr workload logs " + result.WorkloadID, "View the container logs"}
+	status := [2]string{"dr workload status " + result.WorkloadID, "Check the workload status"}
+
+	if failed {
+		if strings.EqualFold(result.Status, workload.WorkloadStatusTerminated) {
```
ajalon1: [P3] This works because `StateTerminated.String()` returns `workload.WorkloadStatusTerminated` directly (same constant), but the coupling is implicit from this package's perspective. Comparing `result.Plan.State == up.StateTerminated` would be more robust.

internal/workload/up/run_test.go
```diff
@@ -1099,6 +1124,100 @@ func TestRun_TerminatedWorkloadDoesNotAnnounceACreate(t *testing.T) {
...
+// could take one, so a refusal printed a full announcement of a rollout and
+// only disowned it on the last line.
+//
+// The drift itself stays: someone who has just been told the workload cannot
+// take a deploy is usually about to ask what the deploy would have been. What
+// changes is that the block says so before the reader gets to it.
+//
+// Both refusals are driven, because they differ in the remedy they name and in
+// what is worth running afterwards.
+func TestRun_RefusedStateDescribesItsDriftRatherThanAnnouncingIt(t *testing.T) {
```
ajalon1, line 1356: [P3] This test covers terminated and errored but not suspended, which is a distinct refusal path (`deployable` → `startable`). `TestRun_SuspendedIsRefusedRatherThanPolled` only asserts the error message, not the refused-note rendering or note-before-changes ordering.

cmd/workload/up/cmd.go
```diff
@@ -497,45 +508,78 @@ func draftWarning(w io.Writer, draft, planned bool) {
...
 // directly under the warning that explains why it is there. Someone who wants
-// to switch a workload off goes looking for the command; someone whose
-// workload switches itself off does not know there is anything to look for,
-// and this list is the one place they will read either way.
-func nextSteps(w io.Writer, result up.Result, draft bool) {
+// to switch a workload off goes looking for the command; someone whose workload
+// switches itself off does not know there is anything to look for, and this
+// list is the one place they will read either way.
+func followUps(result up.Result, draft, failed bool) [][2]string {
 	if result.WorkloadID == "" {
```
ajalon1, line 584: [P3] The `WorkloadID == ""` guard is correct, but the pre-existing `TestCmd_NoWorkloadSaysNothingAboutDrafts` doesn't assert `Next:` is absent on this failed path. A future change moving this guard below the `failed` branch would go uncaught.

ajalon1 (approved):
Approving — the core logic is correct and well-tested. `announce` asks `refusal` before printing the plan, `followUps` splits follow-up commands by outcome, and the dry-run footer is correctly dropped on failure. The `StateTerminated.String()` → `workload.WorkloadStatusTerminated` comparison in `followUps` works because `state.go` returns the same constant directly.

Five P3 inline comments below — all low priority: two are pre-existing issues made more visible by this PR, two are test coverage gaps, and one is a doc-comment nuance about the `StateSettling` exemption. None are blocking.

I know it's just in draft, and you want to test -- but deferring to @adamalpi or @woj for the final call on whether any of these are worth addressing before merge.

cdevent:
> Approving — the core logic is correct and well-tested. `announce` asks `refusal` before printing the plan, `followUps` splits follow-up commands by outcome, and the dry-run footer is correctly dropped on failure. The `StateTerminated.String()` → `workload.WorkloadStatusTerminated` comparison in `followUps` works because `state.go` returns the same constant directly.
> 
> Five P3 inline comments below — all low priority: two are pre-existing issues made more visible by this PR, two are test coverage gaps, and one is a doc-comment nuance about the `StateSettling` exemption. None are blocking.
> 
> I know it's just in draft, and you want to test -- but deferring to @adamalpi or @woj for the final call on whether any of these are worth addressing before merge.

thanks for the good review... let me then fix that in this PR

cdevent:
@adamalpi or @woj any objections ?

PR #838, opened 2026-08-26
[RAPTOR-18978] feat(workload): stream build logs during the deploy

adamalpi:
# RATIONALE

The build is the longest phase of `dr workload up` — routinely minutes — and it rendered as a single spinner line: no sign of whether a builder had even picked the job up, and a failure told you nothing until you went hunting with `dr artifact build logs`. This PR streams the build's own log lines beneath the phase header while it runs, the way docker buildx does, so a deploy's slowest phase is also its most legible one.

Ticket: [RAPTOR-18978](https://datarobot.atlassian.net/browse/RAPTOR-18978)

## CHANGES

- The build phase renders into a fixed-height live window (header + up to 20 rows) that never scrolls: the newest lines slide through, success collapses the window to the usual check-marked line so the final transcript stays clean, and failure keeps the tail as the freshest context for the error.
- Build log lines come from the artifact's OTEL logs route filtered by `external_build_id`. A poll-driven `BuildLogTail` rides `WaitForBuild`'s own cadence (no clock or goroutine of its own), never fails a build over a log-fetch hiccup, and after sustained fetch failures says so once and goes quiet.
- Status transitions are narrated in words (accepted → builder running → pushing), log levels are colored so an error cannot scroll past dimmed, and a heartbeat line covers the gap between trigger and first ingested log line.
- When the tree is unchanged and a build of that code is already running (a previous `up` died, or another terminal is mid-deploy), `up` attaches to it instead of triggering a duplicate.
- The running spinner is aligned with the rest of the phase list: it now sits in the same column as the `⠿`/`✓` glyphs, and the accidental double space after the glyph is gone for every spinner caller.
- Off the terminal (CI, `--output-format json`) every line prints permanently on stderr, so logs stay complete and stdout stays pure.
- Stream fidelity against real OTEL ingestion (found by dogfooding on staging): an empty first poll no longer pins a zero cursor (which used to gap the build's opening behind a 40-line window), the window-mode fetch grew to the server's page size, the cursor lag widened to 60s because build-log ingestion trails the builder by 20–40s, and a 10s holdback reorder buffer lets late-ingested lines slot in ahead of newer ones already fetched — `Finish()` drains it after the wait so nothing is withheld past the build's end. Residual interleave across `#N` steps is buildkit's genuine parallelism, as `docker buildx --progress=plain` shows it.
- The attach path refuses a non-terminal build older than the poll timeout: that build is wedged (orphaned trigger, lost reconcile), and attaching would wait the whole timeout on something that can never finish; the attach message names `--force-build` as the manual escape.

- Heads-up on a repo-wide cosmetic change: the spinner's glyph→label gap goes from two spaces to one for every `RunWithSpinner` call site (the Dot frames carry their own trailing space), aligning the in-flight glyph with the `✓`/`⠿` column.

## PR Automation

---

> [!NOTE]
> **Medium Risk**
> Changes the core `up` build/wait path (attach-vs-trigger, streamed TTY redraw) and adds OTEL log polling, though log streaming is isolated from build success and is covered by new tests.
> 
> **Overview**
> **`dr workload up`** now shows live build output during the image phase instead of a single spinner line. The build step uses a new **`reporter.stream`** path with a docker buildx-style fixed-height TTY window (animated header, sliding tail, success collapses to the usual ✓ line; failures keep the last lines). Off-TTY/CI, every line is written permanently to stderr.
> 
> Build lines are pulled from the artifact OTEL logs API filtered by **`external_build_id`** via **`BuildLogTail`**, which polls on **`WaitForBuild`**’s cadence, dedupes like workload log follow, and never fails the deploy on log fetch errors (warn once and stop streaming).
> 
> **`maybeBuild`** can **attach to an in-flight build** when the tree is unchanged and no image exists yet, avoiding duplicate triggers; changed code still triggers a fresh build. Status transitions and log levels are narrated/styled; spinners align with other phase glyphs via **`RunWithSpinnerPrefix`**.
> 
> Shared log pagination is factored into **`drainLogPages`**; **`logFollower`** now takes a **`logsFetchFn`** so workload follow and build tail share the same follower.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 874fc89ef777902713a6a27063badba260cce2ae. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/.devcontainer/sync-go.sh
+++ b/.devcontainer/sync-go.sh
@@ -11,7 +11,12 @@ set -euo pipefail
 GO_REQUIRED=$(grep '^go ' go.mod | awk '{print $2}')
 GO_INSTALLED=$(go version | awk '{print $3}' | sed 's/go//')
 
-if [ "$GO_REQUIRED" != "$GO_INSTALLED" ]; then
+# A newer toolchain satisfies an older go.mod directive, and the release
+# feed only lists current versions — a superseded patch release cannot be
+# fetched at all. So only ever upgrade forward.
+if [ "$(printf '%s\n%s\n' "$GO_REQUIRED" "$GO_INSTALLED" | sort -V | head -n1)" = "$GO_REQUIRED" ]; then
+    echo "Go $GO_INSTALLED satisfies go.mod's go $GO_REQUIRED"
+elif [ "$GO_REQUIRED" != "$GO_INSTALLED" ]; then
     echo "Go mismatch: need $GO_REQUIRED, have $GO_INSTALLED. Upgrading..."
     ARCH=$(uname -m | sed 's/x86_64/amd64/;s/aarch64/arm64/')
     TARBALL="go${GO_REQUIRED}.linux-${ARCH}.tar.gz"
--- a/internal/workload/build_logs.go
+++ b/internal/workload/build_logs.go
@@ -1,0 +1,120 @@
+// Copyright 2026 DataRobot, Inc. and its affiliates.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package workload
+
+import (
+	"time"
+
+	"github.com/datarobot/cli/internal/drapi"
+)
+
+// A build's logs live on the artifact-wide OTEL route
+// /api/v2/otel/artifact/{id}/logs/: an artifact accumulates every build it
+// ever ran, and the platform's own build id is internal, so searchKeys/
+// searchValues on external_build_id — the build id this CLI already holds
+// from the trigger response — is the only way to narrow the stream to one
+// build. The query keys are camelCase; their snake_case forms are rejected
+// with a 400.
+
+// buildLogSeedLimit is how many recent lines seed a build tail. A tail on a
+// freshly triggered build starts from an empty stream, so the seed only
+// matters when attaching to a build already in progress, where the recent
+// tail is the context the user needs.
+const buildLogSeedLimit = 200
+
+// fetchArtifactBuildLogs retrieves one build's log lines newest-first across
+// pages, by filtering the artifact's OTEL stream on external_build_id.
+func fetchArtifactBuildLogs(artifactID, buildID string, maxEntries int, level, since, reqInfo string) ([]WorkloadLogEntry, error) {
+	query := logsQueryParams(maxEntries, level, since)
+	query.Set("searchKeys", "external_build_id")
+	query.Set("searchValues", buildID)
+
+	pageURL, err := drapi.EndpointURL("/otel/artifact/"+escapeID(artifactID)+"/logs/", query)
+	if err != nil {
+		return nil, err
+	}
+
+	return drainLogPages(pageURL, maxEntries, reqInfo)
+}
+
+// BuildLogTail streams one build's log lines incrementally, driven by the
+// caller's own cadence (WaitForBuild's per-poll callback) rather than a
+// clock of its own. It is strictly progress feedback: no fetch failure is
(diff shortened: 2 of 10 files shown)


ajalon1:
@adamalpi 👏🏽 a screencap video would help immensely :D

adamalpi:
image

.devcontainer/sync-go.sh
```diff
@@ -11,7 +11,12 @@ set -euo pipefail
 GO_REQUIRED=$(grep '^go ' go.mod | awk '{print $2}')
 GO_INSTALLED=$(go version | awk '{print $3}' | sed 's/go//')
 
-if [ "$GO_REQUIRED" != "$GO_INSTALLED" ]; then
+# A newer toolchain satisfies an older go.mod directive, and the release
+# feed only lists current versions — a superseded patch release cannot be
+# fetched at all. So only ever upgrade forward.
+if [ "$(printf '%s\n%s\n' "$GO_REQUIRED" "$GO_INSTALLED" | sort -V | head -n1)" = "$GO_REQUIRED" ]; then
+    echo "Go $GO_INSTALLED satisfies go.mod's go $GO_REQUIRED"
+elif [ "$GO_REQUIRED" != "$GO_INSTALLED" ]; then
```
ajalon1, line 19: Nice! 👍🏽

cdevent (approved):
LGTM, let's run the smoke tests

internal/workload/build_logs.go
```diff
@@ -0,0 +1,219 @@
...
+		func(e WorkloadLogEntry) error { tail.buffer(e); return nil }, onWarn)
+	if err != nil {
+		// Unreachable with the constants above; a nil follower simply means
+		// the tail never emits, which is the contract's worst case anyway.
+		tail.disabled = true
+
+		return tail
+	}
+
+	follower.lag = buildLogLagAllowance
```
wojtekwdr, line 128: [medium] question: did widening the lag to 60s account for how much the dedup can remember? `logDedup` holds followSeenCap=5000 keys per generation, and a since-based fetch passes maxEntries=0 so it drains every page, which means past roughly 83 lines/sec the 60s overlap outruns the dedup and lines get emitted twice. I measured 189k emissions for 100k real lines at 100 lines/sec, and zero duplicates at the old 10s lag. The sliding window hides it on a terminal, but off-TTY every duplicate prints permanently, which is the transcript this PR is trying to keep complete.

internal/workload/up/phase.go
```diff
@@ -62,14 +69,244 @@ func (r *reporter) run(label string, fn func() error) error {
...
+		window: min(streamWindowRows, max(height-4, 1)),
+	}
+}
+
+// append truncates the line to one row, slides the window, and repaints it.
+func (l *liveRegion) append(line string, style lipgloss.Style) {
+	l.mu.Lock()
+	defer l.mu.Unlock()
+
+	line = ansi.Truncate(line, max(l.width-6, 8), "…")
```
wojtekwdr, line 235: [low] `ansi.Truncate` counts a tab as zero columns while lipgloss renders it as four, so a near-width line carrying one tab comes out two columns over, wraps, and every redraw after it is off by a row. Same shape with a bare CR: it survives both `Truncate` and `Render`, and the `\x1b[K` further down the line then fires after the cursor has already gone back to column 0, blanking the row. Both go away if the line is sanitized before it reaches `Truncate`.

internal/workload/logs.go
```diff
@@ -382,15 +418,43 @@ func (f *logFollower) emit(entries []WorkloadLogEntry, hadSince bool) error {
 		}
 	}
 
+	f.advanceCursor(entries)
+
+	// An empty fetch seeds nothing: marking it seeded would pin a zero
+	// cursor, and the NEXT poll — the first with real lines — would be a
+	// window fetch capped at limit, gapping a producer that burst past it
+	// (a builder's opening seconds do exactly that).
+	if len(entries) > 0 {
```
wojtekwdr, line 427: [low] I don't think this guard changes anything. `fetch()` already requires `!f.cursor.IsZero()` and an empty poll can't advance the cursor, so the next poll is a limit-capped window fetch either way; reverting it to an unconditional `f.seeded = true` leaves the repo green, the test named after it included. Which would make the empty-first-poll fix the description credits here really `buildLogSeedLimit` going up to `maxLogsPageSize`.

wojtekwdr (approved):
LGTM

adamalpi:
Good feedback. I will follow up in the stacked PR: https://github.com/datarobot-oss/cli/pull/839

PR #839, opened 2026-08-26
[RAPTOR-18946] feat(workload): read build logs from OTEL, stream them under --wait

adamalpi:
# RATIONALE

> Rebased onto main after #838 merged — the stack is flattened to two commits: the OTEL migration, plus the fixes for the review feedback #838 collected after its merge (dedup overlap capacity, control-character sanitizing in the live region, the seed guard's honest comment).

The per-build logs proxy (`GET /artifacts/{id}/builds/{id}/logs`) is being deprecated ([RAPTOR-18946](https://datarobot.atlassian.net/browse/RAPTOR-18946)): its records only exist while the build service retains the build, which is why a failed build so often answered "logs are not available" at exactly the moment logs were needed. This PR moves the CLI's last callers off that route onto the artifact's OTEL stream (narrowed per build via `external_build_id` — the same source #838 streams from), and upgrades `dr artifact build create --wait` from a silent poll into a live transcript.

## CHANGES

- `GetArtifactBuildLogs` reads the OTEL stream instead of the deprecated proxy; entries map onto the legacy `BuildLogEntry` shape so `dr artifact build logs`, its level filtering, and the failure tail in build summaries carry over. JSON output now passes through the OTEL record (`{timestamp, level, message}`) — a shape change for anyone parsing that output.
- `dr artifact build create --wait` streams the build's log lines and status transitions to stderr while waiting (same tail and wording as `up`'s stream); stdout still carries only the summary document. The per-poll `INFO Fetching build from …` line is gone — the poll fetch carries no request label.
- Log-line and status wording moved to shared `workload.FormatBuildLogLine` / `workload.BuildStatusLine`, so `up` and `build create --wait` say the same things; `up` keeps the coloring.
- No CLI caller remains on the deprecated route (the backend deprecation marking is the workload-api half of the ticket).
- Review feedback from #838, addressed here since the code now lives on this branch: the log dedup is kept as wide as the fetch overlap (keys are pruned once they leave the fetch floor, and the generation cap — now a memory ceiling only — rises to 50000, covering ~800 lines/s over the 60s lag where 5000 was outrun past ~83 lines/s); streamed lines are sanitized before the live region's width math (tabs render as the spaces they occupy, bare CR and other C0 controls are dropped, ESC survives for colors); and the empty-first-poll seed guard's comment now credits its real effect — keeping the gap warning quiet on a seed batch — rather than the gapping fix that `buildLogSeedLimit` actually made.

## PR Automation

> [!IMPORTANT]
>

## RELATED

- Stacked on: #838 (build-log streaming during `dr wl up`)
- Jira: [RAPTOR-18946](https://datarobot.atlassian.net/browse/RAPTOR-18946)

---

> [!NOTE]
> **Medium Risk**
> Switches build log retrieval to a new API and changes JSON log shape for scripted consumers; streaming and dedup changes affect high-volume build output but are covered by new tests.
> 
> **Overview**
> Moves artifact build logging off the deprecated per-build `/logs` proxy onto the artifact OTEL stream filtered by `external_build_id`, so logs remain available after the build service drops short-lived records.
> 
> **`GetArtifactBuildLogs`** (and `dr artifact build logs`) now read that OTEL route, sort chronologically, and map into the existing `BuildLogEntry` shape for text filtering and failure tails. **JSON output** for those entries now passthroughs the OTEL record (`timestamp`, `level`, `message`) instead of the old JSONL fields—a contract change for anyone parsing `--output-format json`.
> 
> **`dr artifact build create --wait`** streams a live transcript to stderr: human-readable status lines via shared **`workload.BuildStatusLine`**, log lines via **`BuildLogTail`** and **`FormatBuildLogLine`** (same wording as `dr workload up`). Stdout still holds only the summary; build status polls use a silent fetch label so progress lines do not interleave.
> 
> Follow-mode **dedup** prunes keys once they fall outside the fetch overlap (generation cap raised to 50k as a memory ceiling only), and seeding waits for the first non-empty batch so the “possible gap” warning does not fire on a normal seed. The **`up`** live region **sanitizes** tabs and other C0 controls before width math so streamed build output does not corrupt the sliding window.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit d1a38e312ac0059dc63f0e3f810348514635e827. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/cmd/artifact/build/create/cmd.go
+++ b/cmd/artifact/build/create/cmd.go
@@ -17,6 +17,7 @@ package create
 import (
 	"errors"
 	"fmt"
+	"strings"
 
 	"github.com/datarobot/cli/cmd/artifact/build/internal/buildargs"
 	"github.com/datarobot/cli/cmd/internal/pollflags"
@@ -41,10 +42,11 @@ The artifact-id argument is optional when run inside a directory linked
 via 'dr artifact code init': the id is read from .datarobot/workload/config.json.
 
 By default the command prints the new build IDs (one per line) and
-exits. With --wait it polls each build until it reaches a terminal
-status (COMPLETED, FAILED, or CANCELLED) and prints a summary with
-duration and resulting image_uri. On failure the tail of the build
-logs is dumped to stderr.
+exits. With --wait it follows each build to a terminal status
+(COMPLETED, FAILED, or CANCELLED), streaming the build's own log lines
+and status transitions to stderr while it waits, then prints a summary
+with duration and resulting image_uri. On failure the tail of the
+build logs is dumped to stderr.
 
 JSON output emits one document:
   - without --wait: the raw trigger response {"buildIds":[...]}.
@@ -141,7 +143,7 @@ func waitForAllBuilds(
 	for _, buildID := range buildIDs {
 		fmt.Fprintf(cmd.ErrOrStderr(), "Waiting for build %s...\n", buildID)
 
-		build, werr := workload.WaitForBuild(artifactID, buildID, poll.Interval, poll.Timeout, nil)
+		build, werr := waitStreaming(cmd, artifactID, buildID, poll)
 		if werr != nil && firstWaitErr == nil {
 			firstWaitErr = werr
 		}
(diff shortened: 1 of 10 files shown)


internal/workload/build.go
```diff
@@ -318,54 +324,39 @@ func ListArtifactBuilds(artifactID string, limit int) ([]Build, error) {
...
+// GetArtifactBuildLogs returns a build's log lines, oldest first, read from
+// the artifact's OTEL stream filtered by external_build_id — the replacement
+// for the deprecated GET /artifacts/{id}/builds/{id}/logs proxy, whose
+// records only existed for as long as IBS retained the build. The OTEL
+// records are mapped onto the legacy BuildLogEntry shape so level filtering
+// and both renderers carry over; Raw holds the OTEL record itself, which is
+// what JSON output now passes through.
 func GetArtifactBuildLogs(artifactID, buildID string) ([]BuildLogEntry, error) {
-	url, err := config.GetEndpointURL("/api/v2/artifacts/" + escapeID(artifactID) + "/builds/" + escapeID(buildID) + "/logs")
+	otel, err := fetchArtifactBuildLogs(artifactID, buildID, 0, "", "", "build logs")
```
wojtekwdr, line 335: [low] non-blocking: with `maxEntries` 0 this drains every page of the build's stream. That's right for `dr artifact build logs`, but `BuildSummaryFor` walks all of them just to take `lastN(logs, 50)` off a failed build. The route answers newest-first, so a capped fetch gets the same 50 lines in one request.

cmd/artifact/build/create/cmd.go
```diff
@@ -141,7 +143,7 @@ func waitForAllBuilds(
 	for _, buildID := range buildIDs {
 		fmt.Fprintf(cmd.ErrOrStderr(), "Waiting for build %s...\n", buildID)
 
-		build, werr := workload.WaitForBuild(artifactID, buildID, poll.Interval, poll.Timeout, nil)
+		build, werr := waitStreaming(cmd, artifactID, buildID, poll)
```
wojtekwdr, line 146: [low] non-blocking: on a failed build the same lines now reach stderr twice, once streamed from here and once as `--- last 50 log lines ---` from the summary. The tail earned its place when the wait was silent. Worth deciding whether it shrinks or drops when a transcript already ran.

internal/workload/logs.go
```diff
@@ -525,13 +549,36 @@ func (d *logDedup) filterUnseen(entries []WorkloadLogEntry) []WorkloadLogEntry {
...
 			d.prev = d.cur
-			d.cur = make(map[string]struct{}, d.genCap)
+			d.cur = make(map[string]time.Time, d.genCap)
+		}
+
+		at, ok := parseLogTimestamp(e.Timestamp)
+		if ok && at.After(d.watermark) {
+			d.watermark = at
+		} else if !ok {
+			at = d.watermark
```
wojtekwdr, line 559: [low] nit: a line whose timestamp doesn't parse adopts `watermark`, which is still the zero time until the first parseable line lands, so `prune` drops that key immediately instead of letting it age out with its neighbours the way the comment above promises.

wojtekwdr (approved):
LGTM

PR #840, opened 2026-08-26
[RAPTOR-19725] feat(workload): sort the base-image picker by language, then name

adamalpi:
# RATIONALE

The wizard's base-image picker showed execution environments in whatever order the server returned them — and trimmed to the 200-item pick limit *while paging*, so what got dropped was whichever environments the server listed last: an arbitrary subset masquerading as "the list". The endpoint offers no server-side ordering to lean on (checked the monolith validators: `searchFor`/`useCases`/`isPublic` only), so the CLI has to make the list legible itself. Ticket: [RAPTOR-19725](https://datarobot.atlassian.net/browse/RAPTOR-19725).

## CHANGES

- `ListExecutionEnvironments` drains the whole listing (still bounded by the existing 100-page cap) before sorting and trimming: languages cluster alphabetically with the catch-all `other` (and anything unlabeled) last, names sort case-insensitively within a language, and the pick-limit trim now cuts a deterministic, explainable prefix instead of a random page boundary.
- The picker gains a LANGUAGE column so the grouping is visible; `other`/empty renders as a dash — a word that means "unlabeled" reads better as absence than as a category.
- `ExecutionEnvironment` carries `programmingLanguage` from the server document.
- Bug fix that surfaced while testing this: the picker's filter could not take a space. Space arrives as its own bubbletea key type (never inside `KeyRunes`), so it fell through to bubbles/table, which binds it to page-down — a name like "Py 3.12" could not be typed. In a filterable table every printable character is filter input (`f` and `b` already are), so space is too; paging keeps pgup/pgdown.

## PR Automation

> [!IMPORTANT]
>

## RELATED

- Jira: [RAPTOR-19725](https://datarobot.atlassian.net/browse/RAPTOR-19725)

---

> [!NOTE]
> **Low Risk**
> Wizard listing and display only; no auth, deploy, or build-resolution logic changes beyond ordering and which rows appear when over the limit.
> 
> **Overview**
> The setup wizard’s **base-image picker** is easier to scan: execution environments are ordered by **programming language** (case-insensitive, with empty/`other` last), then **name**, instead of whatever order the API returned.
> 
> **`ListExecutionEnvironments`** now **pages through the full list** (still capped at 100 pages), **sorts**, then applies the pick limit. Trimming during paging is removed so the limit keeps a **deterministic prefix** of the sorted set rather than an arbitrary early page.
> 
> The picker adds a **LANGUAGE** column from `programmingLanguage` on the API document; unlabeled/`other` values show as **—**. Tests cover drain-then-trim behavior, sort order, and the new column.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit e0622e08afe73811b2d867e01bd3a6f924b7ec72. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/internal/workload/execenv_test.go
+++ b/internal/workload/execenv_test.go
@@ -331,7 +331,7 @@ func TestListExecutionEnvironments_ClustersLanguagesWithOtherLast(t *testing.T) 
 		fmt.Fprintf(w, `{"data": [%s, %s, %s, %s, %s], "next": ""}`,
 			eeDocLang("ee-other", "AAA first by name alone", "other", "v1"),
 			eeDocLang("ee-none", "BBB unlabeled", "", "v2"),
-			eeDocLang("ee-r", "R Lab", "r", "v3"),
+			eeDocLang("ee-r", "R Lab", "R", "v3"),
 			eeDocLang("ee-java", "JVM", "java", "v4"),
 			eeDocLang("ee-py", "Py", "python", "v5"),
 		)
@@ -352,7 +352,8 @@ func TestListExecutionEnvironments_ClustersLanguagesWithOtherLast(t *testing.T) 
 
 	// java < python < r, then the catch-all and the unlabeled at the end by
 	// name — "other" is where the platform files what it cannot name, so it
-	// belongs after the languages that say something.
+	// belongs after the languages that say something. The capital "R" is the
+	// pin on the case folding: byte-sorted it would land ahead of java.
 	assert.Equal(t, []string{"ee-java", "ee-py", "ee-r", "ee-other", "ee-none"}, got)
 }
 
--- a/internal/workload/execenv_test.go
+++ b/internal/workload/execenv_test.go
@@ -25,6 +25,14 @@ import (
 )
 
 // eeDoc is one execution environment as the server serializes it.
+// eeDocLang is eeDoc with a programmingLanguage, for the sorting tests.
+func eeDocLang(id, name, language, versionID string) string {
+	return fmt.Sprintf(
+		`{"id": %q, "name": %q, "programmingLanguage": %q, "latestSuccessfulVersion": {"id": %q}}`,
+		id, name, language, versionID,
+	)
+}
+
 func eeDoc(id, name, versionID string) string {
 	if versionID == "" {
 		return fmt.Sprintf(`{"id": %q, "name": %q, "latestSuccessfulVersion": null}`, id, name)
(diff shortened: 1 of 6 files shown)


internal/workload/wizard/table.go
```diff
@@ -233,6 +233,18 @@ func (t *rowTable) update(msg tea.KeyMsg) bool {
...
 
+		return true
+
+	case msg.Type == tea.KeySpace:
+		// Space arrives as its own key type, never inside KeyRunes, and the
+		// table underneath binds it to page-down — so without this case a
+		// name like "Py 3.12" could not be typed. In a filterable table every
+		// printable character is filter input ("f" and "b" already are), and
+		// paging keeps pgup/pgdown.
+		t.filter += " "
```
wojtekwdr: [low] question: should this carry the `t.filter == ""` guard the `/` case above has? A space as the first keystroke sets the filter to a single space, and then the next Esc gets eaten by `clearFilter` instead of going back a screen.

internal/workload/wizard/screens.go
```diff
@@ -1143,12 +1143,24 @@ func newExecEnvPicker(environments []workload.ExecutionEnvironment, width, heigh
...
 
-	return newRowTable([]string{"BASE IMAGE", "ID"}, rows, true, width, height)
+	return newRowTable([]string{"BASE IMAGE", "LANGUAGE", "ID"}, rows, true, width, height)
+}
+
+// languageLabel renders the language column: the platform's value as-is, and
+// a dash where it says "other" or nothing — a word that means "unlabeled"
+// reads better as absence than as a category.
+func languageLabel(language string) string {
+	lower := strings.ToLower(strings.TrimSpace(language))
```
wojtekwdr: [low] question: the comment above says the platform's value goes in as-is, but this lowercases it. Deliberate? A server value of `R` would render as `r` in the column, which reads oddly next to `python`.

internal/workload/execenv_test.go
```diff
@@ -296,8 +312,48 @@ func TestListExecutionEnvironments_StopsAtTheLimit(t *testing.T) {
 	environments, err := ListExecutionEnvironments(3)
 	require.NoError(t, err)
 
-	assert.Len(t, environments, 3)
-	assert.Equal(t, 2, pages, "paging stops as soon as the limit is reached")
+	// Every page is read BEFORE trimming: an unsorted early stop would keep
+	// whichever environments the server listed first, an arbitrary subset.
+	assert.Equal(t, 2, pages, "the whole listing is drained before sorting and trimming")
+
+	// Languages cluster (case-insensitively), names sort within a language,
```
wojtekwdr, line 319: [low] The fixtures don't actually pin the language folding: drop `strings.ToLower` from `languageSortKey` and the package still passes, since `Python` and `python` bracket nothing else here. A capitalized `R` would catch it, since `R` byte-sorts ahead of `python`.

wojtekwdr (approved):
LGTM

PR #842, opened 2026-08-26
[RAPTOR-19728] fix(workload): no default readiness probe; verify the endpoint after deploy

adamalpi:
# RATIONALE

Since RAPTOR-19535 the wizard defaulted the readiness probe to `/` — and moved the field behind advanced settings, making the guess invisible. API-only frameworks (FastAPI, Express, …) answer 404 at `/` unless a route is defined, so a **working container was turned ERRORED** roughly four minutes after the green tick, with nothing telling the user why. Hit independently twice, on the CLI's own FastAPI example and on a Node app. A wrong probe kills a healthy deploy; that is worse than no probe.

Ticket: [RAPTOR-19728](https://datarobot.atlassian.net/browse/RAPTOR-19728)

## CHANGES

- **No probe by default.** Nothing writes a readiness probe unless `--health <path>` or the settings screen says one. The wizard's health field starts empty (its placeholder already read "none: no readiness probe"); `--no-readiness-probe` keeps its distinct job of actively dropping a deployed probe when binding to a live workload.
- **Dropping the default alone would just move the surprise later** — with no probe, `running` only means the container started. So a deploy now ends with **one anonymous GET against the workload endpoint**, reported and never fatal: "Endpoint check: an anonymous GET answered HTTP 404 Not Found." when something answers, and a plain warning naming `dr workload logs <id>` when nothing does. The status is stated, not judged — a 404 at `/` is how a healthy API-only framework answers, and only the app's author knows whether that is fine.
- The GET is deliberately anonymous (the user's API token has no business in a container's access log; a 401 proves serving as well as a 200 does), runs on every path that ends settled (create, roll, start, retune), and is skipped on `--detach` and `--dry-run`, which never settle.

Docs note: the `config`/`up` doc sections live in #780, which is still open — the probe-default wording there should be reconciled by whichever merges second.

## PR Automation

---

> [!NOTE]
> **Medium Risk**
> Changes default deploy readiness behavior for new manifests and adds network I/O on every settling `up`, which affects first-deploy UX and rollout gating but avoids the prior silent probe failures.
> 
> **Overview**
> Fixes a regression where the wizard silently defaulted readiness probes to **`/`**, which breaks many API frameworks that 404 on the root and leaves deploys **ERRORED** long after a green tick.
> 
> **No probe unless you ask.** `DefaultHealthPath` is removed; manifests and wizard answers omit `readinessProbe` unless `--health <path>` or the settings screen supplies one. `--no-readiness-probe` still explicitly drops a probe when binding to a live workload. Flag help and confirm UI now say **“no probe”** when the health field is empty.
> 
> **Post-deploy sanity check.** After `dr workload up` waits for **running**, `settle` runs one **anonymous GET** on the workload URL (10s timeout), prints the HTTP status or a warning with `dr workload logs <id>`, and **never fails** the command—so a 404 at `/` is reported, not treated as a bad deploy.
> 
> Tests cover rendering without a probe, wizard defaults, endpoint messaging, and that a dead endpoint still exits 0.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 4e446b588ffedb9fa49c59dedc0731da8191b823. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/cmd/workload/config/cmd.go
+++ b/cmd/workload/config/cmd.go
@@ -165,8 +165,9 @@ func addFlags(cmd *cobra.Command, f *flags) {
 
 	cmd.Flags().IntVar(&f.answers.Port, "port", 0, "Container port (default: the Dockerfile's EXPOSE, else 8080).")
 	cmd.Flags().StringVar(&f.answers.HealthPath, "health", "",
-		"Readiness path, e.g. /health. Without it no probe is written: a probe has to be written for the app "+
-			"it probes, and a wrong one stops a healthy deploy from ever going ready.")
+		"Readiness path, e.g. /health. A new manifest gets no probe without it: a probe has to be written "+
+			"for the app it probes, and a wrong one stops a healthy deploy from ever going ready. "+
+			"With --workload-id the bound workload's own probe is kept.")
 	// A flag of its own rather than an empty --health. Declining a probe is a
 	// different act from not mentioning one, and a flag that says so is both
 	// greppable in a CI script and impossible to type by accident.
--- a/cmd/workload/config/cmd.go
+++ b/cmd/workload/config/cmd.go
@@ -164,7 +164,9 @@ func addFlags(cmd *cobra.Command, f *flags) {
 	cmd.Flags().StringVar(&f.answers.Image, "image", "", "Published image URI. Required with --build-mode image.")
 
 	cmd.Flags().IntVar(&f.answers.Port, "port", 0, "Container port (default: the Dockerfile's EXPOSE, else 8080).")
-	cmd.Flags().StringVar(&f.answers.HealthPath, "health", "", "Readiness path (default /). Use --no-readiness-probe to write none.")
+	cmd.Flags().StringVar(&f.answers.HealthPath, "health", "",
+		"Readiness path, e.g. /health. Without it no probe is written: a probe has to be written for the app "+
+			"it probes, and a wrong one stops a healthy deploy from ever going ready.")
 	// A flag of its own rather than an empty --health. Declining a probe is a
 	// different act from not mentioning one, and a flag that says so is both
 	// greppable in a CI script and impossible to type by accident.
--- a/internal/workload/manifest/write.go
+++ b/internal/workload/manifest/write.go
@@ -90,6 +90,7 @@ const (
 	// deploy whose framework answers 404 there, minutes after the green tick
 	// and with nothing saying why. No probe by default, and the deploy's own
 	// endpoint check catches a container that answers nothing.
+
 	DefaultReplicas = 1
 	DefaultCPU      = 0.5
 	DefaultMemory   = "512MB"
--- a/internal/workload/manifest/write.go
+++ b/internal/workload/manifest/write.go
@@ -85,18 +85,14 @@ func ValidImportance(value string) bool {
 const (
 	DefaultImportance = "low"
 	DefaultPort       = 8080
-	// DefaultHealthPath is the root rather than /health because the probe has
-	// to pass on an application nobody wrote it for. Almost every HTTP service
-	// answers something at /, where /health is an endpoint the app has to have
-	// implemented; pointing a probe at one that does not exist is the most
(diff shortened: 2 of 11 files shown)


internal/workload/up/endpoint_check.go
```diff
@@ -0,0 +1,91 @@
...
+// just reported running, so an app that is up answers quickly; the allowance
+// is for one that is still booting, which the report then says.
+const endpointCheckTimeout = 10 * time.Second
+
+// checkEndpointFn performs the GET, swapped by tests. The request is
+// anonymous on purpose: the user's API token has no business in a container's
+// access log, and a 401 or 403 proves something is serving as well as a 200
+// does.
+var checkEndpointFn = func(url string) (int, error) {
+	resp, err := drapi.NewHTTPClient(endpointCheckTimeout).Get(url)
```
wojtekwdr: [medium] question: `NewHTTPClient` leaves `CheckRedirect` at the default, so this reports the status of the last hop after up to ten redirects. If the tenant gateway 302s an anonymous request to a login page, the line says "answered HTTP 200 OK" for a container that serves nothing, which is the case the check exists to catch. Do you know what the gateway does with an anonymous GET? `http.ErrUseLastResponse` would pin the report to the endpoint's own answer.

cmd/workload/config/cmd.go
```diff
@@ -164,7 +164,9 @@ func addFlags(cmd *cobra.Command, f *flags) {
 	cmd.Flags().StringVar(&f.answers.Image, "image", "", "Published image URI. Required with --build-mode image.")
 
 	cmd.Flags().IntVar(&f.answers.Port, "port", 0, "Container port (default: the Dockerfile's EXPOSE, else 8080).")
-	cmd.Flags().StringVar(&f.answers.HealthPath, "health", "", "Readiness path (default /). Use --no-readiness-probe to write none.")
+	cmd.Flags().StringVar(&f.answers.HealthPath, "health", "",
+		"Readiness path, e.g. /health. Without it no probe is written: a probe has to be written for the app "+
```
wojtekwdr: [low] nit: "Without it no probe is written" isn't true with `--workload-id`. `Live.Defaults` seeds the path from the running container and `applyReadiness` leaves it alone, so the live probe survives and gets written. The `--no-readiness-probe` help just below scopes itself to the bound case, this one could too.

internal/workload/up/run.go
```diff
@@ -947,10 +947,19 @@ func isConflict(err error) bool {
...
 	result, err := awaitRunning(workloadID, result, opts, report)
-	if err != nil || !opts.Lock {
+	if err != nil {
 		return result, err
 	}
 
+	// One GET against the endpoint, reported and never fatal. Running means
+	// the container started; with no probe written by default, whether
+	// anything answers is a question nobody has asked yet.
+	verifyEndpoint(result, report)
```
wojtekwdr, line 957: [low] question: any reason this isn't wrapped in `report.run`? It can block for ten seconds printing nothing, where every other wait in the package gets a label and a spinner.

internal/workload/manifest/write.go
```diff
@@ -85,18 +85,14 @@ func ValidImportance(value string) bool {
...
-	// implemented; pointing a probe at one that does not exist is the most
-	// common first-deploy failure, and it presents as a deploy that never
-	// becomes ready rather than as a misconfigured probe. An app with a real
-	// health endpoint says so on the settings screen, and one that wants no
-	// probe at all clears the field.
-	DefaultHealthPath = "/"
-	DefaultReplicas   = 1
-	DefaultCPU        = 0.5
-	DefaultMemory     = "512MB"
+	// There is no default health path. A probe has to be written for the app
```
wojtekwdr, line 88: [low] nit: the comment sits directly above `DefaultReplicas` inside the block, so godoc hangs "There is no default health path" off that constant. A blank line after it, or moving it above the const, keeps it where it belongs.

wojtekwdr (approved):
LGTM

PR #843, opened 2026-08-26
[RAPTOR-19529] feat(workload): make the workload id optional with .datarobot.yaml

wojtekwdr:
# RATIONALE

`dr workload up` ends by printing what to run next, with the workload id pasted into every line. Standing in the project directory that just deployed, you had to type that id back. The binding is already on disk, and `up` itself refuses `--workload-id` because the manifest is where that lives, but every other subcommand demanded it.

Closes RAPTOR-19529.

## CHANGES

The id is now optional on all seven subcommands that take one: `logs`, `status`, `stop`, `get`, `endpoint`, `start`, `delete`. Omitted, it is read from the `workloadId` in the nearest `.datarobot.yaml`, searched upward from a new `--dir` flag.

- A typed id still wins, and is used **without reading any manifest**, so a broken or foreign one nearby cannot fail a command that was told which workload it means.
- The workload picked for you is named on stderr, except under `--output-format json`, where anything on stderr would break `2>&1 | jq .`.
- `stop`, `start` and `delete` confirm a workload named by the manifest rather than by you; `--yes` skips it. A typed id is never questioned, because the id is the consent. `delete` keeps asking either way, as it already did.
- `up` no longer pastes the id into its Next block. The lines print as they can be typed, carrying `--dir` when the deploy did. A run that failed after creating the workload names the id above them, since that is the one case where the manifest holds no binding to resolve.

Resolution parses the manifest but never validates it: the ledger has no rule about `workloadId`, and `logs` is what you reach for when a project is broken.

Comments are deliberately terser than this package's usual style.

## TESTING

`task lint` clean on linux, darwin and windows; `go test -race -shuffle=on ./...` green.

Verified against staging with real deployments: full `up` → `stop` → `start` → `lock` → `delete` cycle; resolution from the project root, a subdirectory and the parent with `--dir`; JSON purity; a deploy onto a stopped workload to check against RAPTOR-19686; and the confirmation prompt driven through a real pty, where declining leaves the workload running and a typed id never prompts.

Staging caught two bugs the unit tests missed, both now fixed with regression tests: `dr workload status "$UNSET_VAR"` resolved from the manifest instead of failing loudly, and `up --lock` was the one Next-block line that would not run as printed under `--dir`.

## NOTES

`--yes` on `stop` and `start` is new surface. It only gates an interactive prompt that never fires in automation, so it should not affect the IaC equivalence PBMP-7947 asks for, but flagging it.

---

> [!NOTE]
> **Medium Risk**
> Broad CLI behavior change across read and mutating workload commands; manifest-based targeting plus new stop/start confirmation paths could affect scripts and automation if `--dir` or cwd resolution is wrong, though explicit ids and `--yes` mitigate destructive mistakes.
> 
> **Overview**
> Workload subcommands can omit the positional **workload id** and resolve it from the nearest **`.datarobot.yaml`** (`workloadId`), walking upward from **`--dir`** (default: cwd). A new **`cmd/workload/internal/idargs`** package centralizes **`--dir`**, shared help text, **`Resolve`**, stderr “using workload …” (suppressed for **`--output-format json`**), and **`Ref.Wrap`** for clearer 404/403 when the id came from a manifest.
> 
> **`status`**, **`get`**, **`endpoint`**, and **`logs`** use this resolution; **`delete`**, **`stop`**, and **`start`** do too. **`stop`**/**`start`** add **`--yes`** and prompt only when the target was manifest-chosen (typed ids skip the extra prompt; **`delete`** still always confirms). Telemetry records **`workload_id_source`**.
> 
> **`dr workload up`** “Next” hints no longer embed the id—they print runnable commands (with **`--dir`** when deploy used it); failed runs that created a workload without writing the manifest still print the id explicitly.
> 
> Docs and tests cover optional arity, end-to-end manifest resolution, JSON stderr silence, and stop confirmation behavior.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit c52fe5bf86474c991606cbb44d4e275765a007b7. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/cmd/workload/del/cmd.go
+++ b/cmd/workload/del/cmd.go
@@ -26,7 +26,7 @@ import (
 
 	"github.com/datarobot/cli/cmd/workload/internal/idargs"
 	"github.com/datarobot/cli/internal/auth"
-	"github.com/datarobot/cli/internal/config/viperx"
+	"github.com/datarobot/cli/internal/cli"
 	"github.com/datarobot/cli/internal/drapi"
 	"github.com/datarobot/cli/internal/telemetry"
 	"github.com/datarobot/cli/internal/workload"
@@ -100,7 +100,7 @@ Example:
 	idargs.AddYesFlag(cmd, "Skip the confirmation prompt.")
 
 	telemetry.TrackWith(cmd, func(cmd *cobra.Command, args []string) map[string]any {
-		yesFlag, _ := cmd.Flags().GetBool("yes")
+		yesFlag, _ := cmd.Flags().GetBool(cli.YesFlagName)
 
 		return map[string]any{
 			"workload_id":        idargs.TelemetryID(ref, args),
@@ -108,7 +108,7 @@ Example:
 			// The environment variable is only consent for a workload the user
 			// named, so reporting it unconditionally would say yes about the
 			// runs this command refuses for want of it.
-			"yes": yesFlag || (!ref.FromManifest() && viperx.GetBool("yes")),
+			"yes": yesFlag || (!ref.FromManifest() && cli.IsNonInteractive(cmd)),
 		}
 	})
 
--- a/cmd/workload/del/cmd.go
+++ b/cmd/workload/del/cmd.go
@@ -68,9 +68,6 @@ Example:
 		PreRunE:      auth.EnsureAuthenticatedE,
 		SilenceUsage: true,
 		RunE: func(cmd *cobra.Command, args []string) error {
-			// Resolve first: it reads --dir and refuses a path that is not a
-			// directory, and a typo is worth catching while it still costs
-			// nothing, on this side of an irreversible call.
 			var err error
 
 			ref, err = idargs.Resolve(cmd, args)
@@ -78,13 +75,13 @@ Example:
 				return err
 			}
 
-			confirmed, err := confirmDelete(cmd, ref.ID)
+			confirmed, err := confirmDelete(cmd, ref)
 			if err != nil || !confirmed {
 				return err
 			}
 
 			if err := workload.DeleteWorkload(ref.ID); err != nil {
-				return handleDeleteError(err, ref.ID)
+				return handleDeleteError(err, ref)
 			}
 
 			fmt.Println(tui.BaseTextStyle.Render("Deleted workload: " + ref.ID))
@@ -96,37 +93,51 @@ Example:
 	}
 
 	idargs.AddDirFlag(cmd)
+	// Only here does --dir also decide which manifest gets its binding cleared,
+	// which is the whole reason a delete by typed id still takes the flag.
+	cmd.Flags().Lookup("dir").Usage += " Its binding to the deleted workload is cleared too."
+
 	idargs.AddYesFlag(cmd, "Skip the confirmation prompt.")
 
-	telemetry.TrackWith(cmd, func(cmd *cobra.Command, _ []string) map[string]any {
+	telemetry.TrackWith(cmd, func(cmd *cobra.Command, args []string) map[string]any {
 		yesFlag, _ := cmd.Flags().GetBool("yes")
 
 		return map[string]any{
-			"workload_id":        ref.ID,
+			"workload_id":        idargs.TelemetryID(ref, args),
 			"workload_id_source": ref.Source,
-			"yes":                yesFlag || viperx.GetBool("yes"),
+			// The environment variable is only consent for a workload the user
+			// named, so reporting it unconditionally would say yes about the
+			// runs this command refuses for want of it.
+			"yes": yesFlag || (!ref.FromManifest() && viperx.GetBool("yes")),
 		}
 	})
 
 	return cmd
 }
 
-// confirmDelete returns (true, nil) when the deletion may proceed: either
-// --yes / DATAROBOT_CLI_NON_INTERACTIVE was given, or the user confirmed
-// interactively. A declined prompt is (false, nil) so the command exits 0
-// as a no-op.
-func confirmDelete(cmd *cobra.Command, workloadID string) (bool, error) {
-	confirmed, err := idargs.Confirm(cmd,
-		"Delete workload "+workloadID+"? This stops and removes a running workload. [y/N] ")
-	if err != nil {
-		return false, err
+// confirmDelete returns (true, nil) when the deletion may proceed: either it
+// was already answered, or the user confirmed interactively. A declined prompt
+// is (false, nil) so the command exits 0 as a no-op.
+//
+// A workload the manifest named rather than the user takes the explicit --yes
+// only. Delete is the one irreversible verb here, and the environment variable
+// that suppresses wizards in CI should not also stand as consent to remove
+// something nobody typed.
+func confirmDelete(cmd *cobra.Command, ref idargs.Ref) (bool, error) {
+	env := idargs.EnvMayConsent
+	if ref.FromManifest() {
+		env = idargs.EnvMayNotConsent
 	}
 
-	if !confirmed {
-		fmt.Println(tui.DimStyle.Render("Aborted."))
+	named := "workload " + ref.ID
+	if ref.FromManifest() {
+		// An id the reader never typed is nothing to compare against; the file
+		// that chose it is what makes the answer meaningful.
+		named += ", named by " + ref.DisplayPath()
 	}
 
-	return confirmed, nil
+	return idargs.Confirm(cmd,
+		"Delete "+named+"? This stops and removes a running workload. [y/N] ", env)
 }
 
 // handleDeleteError converts a 404 into a friendly informational message
@@ -141,16 +152,26 @@ func confirmDelete(cmd *cobra.Command, workloadID string) (bool, error) {
 // branch where nothing was actually deleted, would be the stronger claim made
 // on the weaker evidence. A binding that really is dead is cleared by the
 // deploy that recreates it.
-func handleDeleteError(err error, workloadID string) error {
+func handleDeleteError(err error, ref idargs.Ref) error {
 	var httpErr *drapi.HTTPError
 
 	if errors.As(err, &httpErr) && httpErr.StatusCode == http.StatusNotFound {
-		fmt.Println(tui.DimStyle.Render("No workload found with id: " + workloadID))
+		// The manifest is named when it, rather than the user, chose the id:
+		// otherwise this reports an id the reader has never seen and gives
+		// them nowhere to look.
+		said := "No workload found with id: " + ref.ID
+		if ref.FromManifest() {
+			said += ", named by " + ref.DisplayPath()
+		}
+
+		fmt.Println(tui.DimStyle.Render(said))
 
 		return nil
 	}
 
-	return err
+	// Anything else keeps the provenance too, which is what turns a bare 403
+	// against an ambient id into something actionable.
+	return ref.Wrap(err)
 }
 
 // clearStaleBinding takes back the workloadId the CLI wrote into a manifest,
@@ -206,7 +227,8 @@ func clearStaleBinding(w io.Writer, dir, workloadID string) {
 	// Appending a single generic "remove that line" walked the user into the
 	// empty manifest the only-key refusal exists to prevent.
 	if err != nil {
-		fmt.Fprintln(w, tui.DimStyle.Render("Could not remove workloadId from "+path+": "+err.Error()))
+		fmt.Fprintln(w, tui.DimStyle.Render(
+			"Could not remove workloadId from "+idargs.DisplayPath(path)+": "+err.Error()))
 
 		return
 	}
@@ -216,7 +238,8 @@ func clearStaleBinding(w io.Writer, dir, workloadID string) {
 	}
 
 	fmt.Fprintln(w, tui.DimStyle.Render(
-		"Removed workloadId from "+path+"; the next 'dr workload up"+manifest.DirFlag(dir)+"' creates a new workload."))
+		"Removed workloadId from "+idargs.DisplayPath(path)+
+			"; the next 'dr workload up"+manifest.DirFlag(dir)+"' creates a new workload."))
 
 	noteLinkedArtifact(w, filepath.Dir(path))
 }
--- a/cmd/workload/del/cmd.go
+++ b/cmd/workload/del/cmd.go
@@ -24,13 +24,10 @@ import (
 	"net/http"
 	"path/filepath"
 
-	"github.com/datarobot/cli/cmd/helpers"
+	"github.com/datarobot/cli/cmd/workload/internal/idargs"
 	"github.com/datarobot/cli/internal/auth"
-	"github.com/datarobot/cli/internal/cli"
 	"github.com/datarobot/cli/internal/config/viperx"
 	"github.com/datarobot/cli/internal/drapi"
-	"github.com/datarobot/cli/internal/fsutil"
-	"github.com/datarobot/cli/internal/misc/reader"
 	"github.com/datarobot/cli/internal/telemetry"
 	"github.com/datarobot/cli/internal/workload"
 	"github.com/datarobot/cli/internal/workload/manifest"
@@ -40,8 +37,10 @@ import (
 )
 
 func Cmd() *cobra.Command {
+	var ref idargs.Ref
+
 	cmd := &cobra.Command{
-		Use:   "delete <workload-id>",
+		Use:   "delete [<workload-id>]",
 		Short: "Delete a workload.",
 		Long: `Delete a workload by id.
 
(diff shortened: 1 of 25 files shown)


cmd/workload/internal/idargs/confirm.go
```diff
@@ -0,0 +1,65 @@
...
+		return true, nil
+	}
+
+	if !reader.IsStdinTerminal() {
+		return false, errors.New(
+			"confirmation required: pass --yes (or set " + reader.NonInteractiveEnv +
+				"=1) to run without a prompt")
+	}
+
+	return helpers.Confirm(cmd.OutOrStdout(), cmd.InOrStdin(), question)
```
adamalpi: **[Medium] The confirmation prompt writes to stdout**

Under `--output-format json` with an interactive terminal and an ambient id, this prompt prints ahead of the JSON document — breaking exactly the `| jq .` contract `announce` so carefully protects (it skips JSON; `Confirm` doesn't know about formats at all). In text mode it also lands in `stop`/`start`'s documented "acknowledgement on stdout" when captured.

In-repo precedent already settles the stream question: `up.typedConfirm` sends its question to **stderr** for precisely this reason. Suggest: prompt to `cmd.ErrOrStderr()`, and decide the JSON case explicitly — either json ⇒ non-interactive like `up` (error with the `--yes` remedy) or prompt on stderr. There's currently no test covering the JSON+prompt path.

cmd/workload/del/cmd.go
```diff
@@ -58,59 +58,53 @@ in a subdirectory is not visible from its parent.
...
 
-	cmd.Flags().BoolP("yes", "y", false, "Skip the confirmation prompt.")
-	cmd.Flags().String("dir", "", "Project directory whose manifest holds the binding, searched upward from here.")
-
-	// Bind only the env var (DATAROBOT_CLI_NON_INTERACTIVE) to viper. The --yes
-	// flag itself is read directly from cmd.Flags() so an explicit --yes does
-	// not leak into viper.AllSettings() and persist to drconfig.yaml.
-	_ = viperx.BindEnv("yes", "DATAROBOT_CLI_NON_INTERACTIVE")
+	idargs.AddDirFlag(cmd)
+	idargs.AddYesFlag(cmd, "Skip the confirmation prompt.")
```
adamalpi, line 100: **[Medium] Ambient delete + `DATAROBOT_CLI_NON_INTERACTIVE` auto-consent**

The env⇒yes binding is pre-existing, but delete previously required a typed id — the victim was at least named. Now `DATAROBOT_CLI_NON_INTERACTIVE=1 dr workload delete` with **no id and no `--yes`** resolves whatever the nearest manifest names and deletes it, prompt-free. An env var set globally in CI to suppress wizards becomes standing consent to an irreversible delete of an unnamed target.

Suggest: when the target is ambient, require the explicit `--yes` flag (env not sufficient) — the same "the typed id is the consent" reasoning this PR applies elsewhere, extended to its conclusion. (The NOTES flag `--yes` on stop/start re PBMP-7947, but this combination is the sharper one.)

cmd/workload/internal/idargs/idargs.go
```diff
@@ -0,0 +1,241 @@
...
+
+	if httpErr.StatusCode != http.StatusNotFound && httpErr.StatusCode != http.StatusForbidden {
+		return err
+	}
+
+	// A 404 means "not on this instance", which is not the same as gone: the
+	// same manifest read against the wrong endpoint, organisation or token
+	// answers exactly this way, so the remedy names both.
+	if httpErr.StatusCode == http.StatusForbidden {
+		return fmt.Errorf("no access to workload %s, named by %s", r.ID, shortPath(r.Path))
```
adamalpi: **[Low] `Wrap` discards the underlying error**

The 404/403 rewrites drop the original entirely (no `%w`), so the `*drapi.HTTPError` disappears from the chain — status-code-aware callers (`errors.As`) and the URL/status detail for debugging are both lost. Wrapping with `: %w` keeps the friendly sentence and the chain.

cmd/workload/up/cmd.go
```diff
@@ -501,28 +501,42 @@ func draftWarning(w io.Writer, draft, planned bool) {
...
 
 	if draft {
 		steps = append([][2]string{
-			{"dr workload up --lock", "Lock the artifact to make it permanent"},
+			{"dr workload up --lock" + at, "Lock the artifact to make it permanent"},
 		}, steps...)
 	} else {
 		steps = append(steps,
-			[2]string{"dr workload stop " + result.WorkloadID, "Stop the workload, retaining its version"})
+			[2]string{"dr workload stop" + at, "Stop the workload, retaining its version"})
```
adamalpi, line 533: **[Nit] The deploy now suggests a command that will interrogate you**

Bare `dr workload stop` here is ambient, so running it interactively prompts about the workload `up` itself just deployed. Probably the right safety trade — just flagging it as a conscious choice rather than an accident.

adamalpi (approved):
Careful work — the empty-positional guard, the explicit-id-never-reads-the-filesystem invariant, and the failed-deploy exception in the Next block are all exactly right. Two consent/stream edges worth tightening, neither blocking.

cmd/workload/internal/idargs/confirm.go
ajalon1: `Confirm` writes the question to `cmd.OutOrStdout()`. That was fine when only `delete` used it (no JSON mode there), but `stop`/`start` do render JSON now, and `reader.IsStdinTerminal()` only checks stdin, not stdout. So `dr workload stop --output-format json | jq .` from a terminal still fires the prompt, and it goes down the pipe: `jq` chokes, and the user just sees a hang with no visible question.

`up` already solved this exact problem two files over (`cmd/workload/up/cmd.go:333`) and left a comment on why: *"stdout is the endpoint, or one JSON document, and a question printed into it would break whatever is parsing it."* Worth reusing that reasoning here, swap to `cmd.ErrOrStderr()`.

Requesting changes for this one.

cmd/workload/internal/idargs/idargs.go
```diff
@@ -0,0 +1,241 @@
...
+	announce(cmd, ref)
+
+	return ref, nil
+}
+
+// announce names a workload the user did not, so an ambient target is never
+// silent. JSON runs stay quiet: stdout purity is only half the contract, and
+// `--output-format json 2>&1 | jq .` has to parse too.
+func announce(cmd *cobra.Command, ref Ref) {
+	if !ref.FromManifest() || outputformat.GetFormat(cmd) == outputformat.OutputFormatJSON {
```
ajalon1: `announce()` suppresses the "Using workload X from Y" notice whenever `outputformat.GetFormat(cmd)` is JSON, but `--output-format` is a root persistent flag (`cmd/root.go:266`), so `delete` and `endpoint` inherit it even though neither ever renders JSON. Set `DATAROBOT_CLI_OUTPUT_FORMAT=json` and `delete` silently resolves from a manifest with zero attribution, for a suppression that buys nothing on that leaf.

Maybe key this on whether the command actually has an `--output-format` flag registered, rather than the inherited value?

Requesting changes for this one.

cmd/workload/internal/idargs/idargs.go
```diff
@@ -0,0 +1,241 @@
...
+		return
+	}
+
+	fmt.Fprintln(cmd.ErrOrStderr(),
+		tui.DimStyle.Render("Using workload "+ref.ID+" from "+ref.Path))
```
ajalon1: `announce` and `AmbientPrompt` (`confirm.go:64`) both print `ref.Path` raw, but `Ref.Wrap` runs the same value through `shortPath` before using it. So a manifest a few directories deep prints its full absolute path on every single `status`/`stop` call, but gets nicely shortened only on a 404. `shortPath`'s own doc comment argues against this: *"Naming /Users/me/.../.datarobot.yaml at someone standing in it spends three lines saying 'here'."* Seems worth running `announce`/`AmbientPrompt` through the same helper.

cmd/workload/internal/idargs/idargs.go
```diff
@@ -0,0 +1,241 @@
...
+		"Project directory whose .datarobot.yaml names the workload, searched upward from here.")
+}
+
+// Resolve returns the workload the command should act on, announcing one read
+// from a manifest on stderr.
+//
+// The notice is printed here rather than at each call site so seven commands
+// cannot each decide differently whether to say.
+func Resolve(cmd *cobra.Command, args []string) (Ref, error) {
+	dir, err := dirFlag(cmd)
```
ajalon1, line 68: `Resolve` calls `dirFlag` (which stats `--dir` and errors on a non-directory) before `resolve()` gets a chance to short-circuit on a typed id. So `dr workload logs 68b0... --dir /gone` fails on `--dir`, even though the PR's own design intent is that an explicit id "is used without reading any manifest" and shouldn't care what `--dir` points at.

Test coverage matches the bug, not the intent: `idargs_test.go`'s "explicit id needs no manifest at all" case calls the unexported `resolve()` directly, which never goes through `dirFlag`, so this path isn't actually exercised. `delete` genuinely needs the early check (`clearStaleBinding` uses `ref.Dir`), but the other six leaves probably shouldn't fail on a bad `--dir` when an id was typed.

cmd/workload/logs/cmd.go
```diff
@@ -80,28 +86,36 @@ Example:
...
+			}
+
 			if follow {
 				// Warnings go to stderr so stdout stays log lines only.
 				onWarn := func(msg string) {
 					fmt.Fprintln(cmd.ErrOrStderr(), "warning: "+msg)
 				}
 
-				return workload.FollowWorkloadLogs(cmd.Context(), args[0], limit, parsedLevel, interval,
+				return workload.FollowWorkloadLogs(cmd.Context(), ref.ID, limit, parsedLevel, interval,
```
ajalon1: The `--follow` branch returns the raw error from `FollowWorkloadLogs` without going through `ref.Wrap`. Two lines down, the non-follow branch does call `ref.Wrap(err)`. So a manifest-resolved id that 404s (stale binding, wrong endpoint/org) gets the nice "workload X not found, Y names it, redeploy with..." message on the plain path, but a raw `drapi.HTTPError` on `--follow`, which happens to be the exact example in the new docs section (`dr workload logs --follow` after `cd my-app`).

cmd/workload/del/cmd.go
```diff
@@ -58,59 +58,53 @@ in a subdirectory is not visible from its parent.
...
-				return fmt.Errorf("--dir %s: %w", dir, manifest.ErrNotADirectory)
+			confirmed, err := confirmDelete(cmd, ref.ID)
+			if err != nil || !confirmed {
+				return err
 			}
 
-			if err := workload.DeleteWorkload(args[0]); err != nil {
-				return handleDeleteError(err, args[0])
+			if err := workload.DeleteWorkload(ref.ID); err != nil {
+				return handleDeleteError(err, ref.ID)
```
ajalon1: `delete` is the only leaf that resolves from a manifest but never calls `ref.Wrap` (see `handleDeleteError` below). A stale binding prints `"No workload found with id: 68b0..."`, an id the user never typed, no mention of `.datarobot.yaml`. A 403 is worse, the raw HTTP error passes through untouched. `Wrap`'s own doc comment describes this exact gap as the reason it exists. `handleDeleteError` returns `nil` on 404 so `Wrap` can't drop in verbatim, but the informational line could still name `ref.Path` when `ref.FromManifest()`.

cmd/workload/del/cmd.go
```diff
@@ -58,59 +58,53 @@ in a subdirectory is not visible from its parent.
...
+	idargs.AddYesFlag(cmd, "Skip the confirmation prompt.")
 
-	telemetry.TrackWith(cmd, func(cmd *cobra.Command, args []string) map[string]any {
+	telemetry.TrackWith(cmd, func(cmd *cobra.Command, _ []string) map[string]any {
 		yesFlag, _ := cmd.Flags().GetBool("yes")
 
 		return map[string]any{
-			"workload_id": telemetry.FirstArg(args),
-			"yes":         yesFlag || viperx.GetBool("yes"),
+			"workload_id":        ref.ID,
```
ajalon1: `ref` is only assigned inside `RunE`, but `PreRunE: auth.EnsureAuthenticatedE` runs first on every leaf (same pattern across all seven commands, not just this one). Any failure before `Resolve` runs (auth, bad `--limit`, etc.) now reports `workload_id: ""` in telemetry where the old `telemetry.FirstArg(args)` would've captured the typed id. Might be worth falling back to `telemetry.FirstArg(args)` when `ref.ID` is empty.

cmd/workload/up/cmd.go
```diff
@@ -501,28 +501,42 @@ func draftWarning(w io.Writer, draft, planned bool) {
...
 // and this list is the one place they will read either way.
-func nextSteps(w io.Writer, result up.Result, draft bool) {
+func nextSteps(w io.Writer, result up.Result, dir string, draft, failed bool) {
 	if result.WorkloadID == "" {
 		return
 	}
 
+	at := manifest.DirFlag(dir)
+
+	if failed {
```
ajalon1, line 514: On a failed run, this prints `"Workload: <id>"` and then id-less commands below it, but the one failure shape this line exists for (workload created, id couldn't be written back to the manifest) is exactly the case where the manifest holds no binding. So none of the printed commands can actually resolve; the reader has to manually splice the id from the line above into each one. The comment right above (line 503) acknowledges this but the code still drops the id. Pasting the id into the commands themselves (like the pre-PR behavior) seems safer for this branch specifically.

docs/commands/workload.md
```diff
@@ -202,6 +206,24 @@ dr workload logs <workload-id> [--limit N] [--level <level>] [--follow] [--outpu
 - `--follow`, `-f`: stream new lines as they arrive.
 - `--output-format <text|json>`: output format. Defaults to `text`. With `--follow`, JSON is emitted as one object per line (JSON Lines).
 
+## Working in a project directory
+
+Every command that takes a `<workload-id>` can leave it out inside a project that `dr workload up` has deployed. The id is read from the `workloadId` in the nearest `.datarobot.yaml`, searched upward from `--dir` (the current directory by default), and the command says on stderr which workload it picked:
```
ajalon1, line 213: This claims "every command that takes a `<workload-id>` can leave it out", but the `get` synopsis and the `start`/`stop` synopsis further up in this file still show a required id with no `--dir`/`--yes`. Only `status`, `endpoint`, and `logs` got their synopsis updated. Also the `delete` `--dir` description still says "whose manifest holds the binding" while `idargs.AddDirFlag`'s actual help text now says "whose `.datarobot.yaml` names the workload", worth reconciling.

cmd/workload/stop/cmd.go
```diff
@@ -44,40 +47,64 @@ The acknowledgement message is printed on stdout. Use
...
+			var err error
+
+			ref, err = idargs.Resolve(cmd, args)
 			if err != nil {
 				return err
 			}
 
+			// Only an ambient target is confirmed: a typed id is the consent.
+			if ref.FromManifest() {
+				confirmed, err := idargs.Confirm(cmd, idargs.AmbientPrompt("Stop", ref))
```
ajalon1: When the user answers `n` here, `stop`/`start` exit 0 with nothing printed after the prompt line, no confirmation the decline registered. `delete` prints `"Aborted."` for the identical case (`cmd/workload/del/cmd.go:126`). Since all three now share `idargs.Confirm`, might make sense to put that acknowledgment there once instead of duplicating it.

cmd/workload/resolve_test.go
```diff
@@ -0,0 +1,139 @@
...
+	"github.com/datarobot/cli/internal/config/viperx"
+	"github.com/datarobot/cli/internal/testutil"
+	"github.com/datarobot/cli/internal/workload/manifest"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+const wireID = "68b0c1d2e3f4a5b6c7d8e9f0"
+
+// TestUnboundID covers the whole feature end to end for every verb that takes
```
ajalon1: The comment says this test "covers the whole feature end to end for every verb that takes an id" and is "what catches a leaf that forgot to resolve", but it only runs `status`, `get`, `endpoint`, `logs`. `delete` and `start` have no end-to-end manifest-resolution test; `del/cmd_test.go` only checks arity/confirmation with an explicit id, and `stop/confirm_test.go` covers `stop` while just asserting in a comment that "`start` shares this code path." The destructive leaf is the one with no coverage of the new resolution behavior.

ajalon1 (commented):
Nice extraction — one place for arity, `--dir`, help text, and the explicit-id-wins invariant instead of seven commands drifting apart. Found a few gaps at the seams where a leaf opted out of the shared path, one of them a real correctness bug (confirmation prompt on stdout breaks JSON piping). Details inline.

cmd/workload/del/cmd.go
```diff
@@ -58,59 +58,53 @@ in a subdirectory is not visible from its parent.
...
-				return handleDeleteError(err, args[0])
+			if err := workload.DeleteWorkload(ref.ID); err != nil {
+				return handleDeleteError(err, ref.ID)
 			}
 
-			fmt.Println(tui.BaseTextStyle.Render("Deleted workload: " + args[0]))
+			fmt.Println(tui.BaseTextStyle.Render("Deleted workload: " + ref.ID))
 
-			clearStaleBinding(cmd.ErrOrStderr(), dir, args[0])
+			clearStaleBinding(cmd.ErrOrStderr(), ref.Dir, ref.ID)
```
ajalon1, line 89: Minor: `clearStaleBinding` re-walks the directory tree via `manifest.Locate` even though `ref.Path` already has the located manifest when `ref.FromManifest()`. Not a bug, just a redundant search plus stat calls, and passing `ref.Path` through would also guarantee the file cleared is the same file the id came from, rather than assuming two separate walks agree.

ajalon1 (changes requested):
Requesting changes for the output-format json break and for the stdout / stderr bit, otherwise LGTM and you can re-request or dismiss my review after fixing.

cmd/workload/internal/idargs/confirm.go
wojtekwdr: Fixed, it goes to stderr now.

Your point about `IsStdinTerminal` only watching stdin was the one that mattered: a JSON run now refuses outright instead of asking, since the question would land in the document either way. While testing that I found the mirror image, a redirected stderr swallowing the prompt and leaving it hanging, so it checks both ends before asking.

cmd/workload/internal/idargs/idargs.go
wojtekwdr: Fixed. Keyed on `LocalFlags()` now, so `delete` and `endpoint` keep their attribution when the root value happens to be json.

cmd/workload/internal/idargs/idargs.go
wojtekwdr: I did this, then walked it back, so flagging it rather than letting you find it.

A later pass argued that a silently discarded `--dir` is how a typo becomes "why did nothing change", and both `up` and `delete` reject the same mistake. So `--dir` is checked whenever it is typed, but a typed id still never reads a manifest, which I think was the invariant you were actually protecting. The test now covers that directly instead of going through the unexported half. Say if you would rather have it the other way.

cmd/workload/up/cmd.go
wojtekwdr: Fixed, the failed branch puts the id back in the commands.

Chasing it turned up something worse one line over: `draftIsServing` can be true on a failed run, so the block could print a bare `dr workload up --lock` next to id-carrying siblings. With no binding written that creates a second workload instead of locking the first. That line is dropped on a failed run now.

cmd/workload/del/cmd.go
wojtekwdr: Agreed, and this was the sharper one of the two. A manifest-named delete takes the explicit `--yes` only now. The env var still answers when you typed the id, so nothing scripted changes.

cmd/workload/del/cmd.go
wojtekwdr: Leaving this one for now. `ClearWorkloadID` compares the id inside the edit, so a second walk landing on a different file just would not match rather than clear the wrong binding. Happy to thread `ref.Path` through if you want the stat calls gone.

ajalon1: nah, let's drop it

cmd/workload/internal/idargs/idargs.go
ajalon1: fine by me as is

ajalon1 (approved):
send it

PR #844, opened 2026-08-26
[RAPTOR-19527] fix(workload): wait for the new version to serve

wojtekwdr:
# RATIONALE

`dr workload up` printed a tick, exited 0, and the endpoint kept serving the previous build.

The wait asked only whether the workload was running, which is true from beginning to end of a rolling swap, so it returned on the first poll having proved nothing. Confirming the artifact is not enough either: the workload names the new one about a minute before the endpoint stops answering with the old build, and four minutes in one run. The proton list is the only place the platform says the outgoing generation has stopped taking requests.

## CHANGES

The wait takes a `Serving` struct saying what it must see, and settles only once the workload reports the artifact and the generation it replaced has stopped draining. A create and a start ask for neither and skip the extra request; a resize asks for the drain with no artifact to name.

Nothing about the proton route can fail a deploy: a refusal degrades at once, anything else is retried and then degrades, and the transcript says the handover went unconfirmed and why.

This also fixes `up --lock` making the wrong artifact permanent. On a draft workload the lock is applied after the wait, to whatever it last saw, so an early return locked the version being rolled off.

## NOTES

Rolls now take around eight minutes rather than thirty seconds. The endpoint switched at t+171s in one run and t+284s in another, and nothing in the API marks that moment, so the wait sits through the drain window. `--detach` still returns at once. RAPTOR-19347 and RAPTOR-19734 are the platform-side asks that would let this be quick as well as correct.

## TESTING

Against staging with real deployments: create, roll, resize, stop and start. The roll previously exited after 35s with the old build still serving, and now exits after 8m49s with the new build live. At t+43s of a roll both generations are running and the outgoing one still holds the active role, which is the window the predecessor check exists for.

---

> [!NOTE]
> **Medium Risk**
> Changes core deploy wait semantics and duration (rolls block through drain); incorrect proton/artifact logic could hang or mis-report success, but behavior is heavily tested and missing protons degrades safely.
> 
> **Overview**
> Fixes **`dr workload up`** reporting success while the endpoint still serves the previous build during rolling swaps.
> 
> **`WaitForWorkload`** now accepts an optional **`wantArtifactID`**. Rolls pass the candidate artifact and only finish when the workload reports that artifact *and* **`GET /workloads/{id}/protons/`** shows no generation still **draining** (with artifact-only fallback if protons 404). Create, start, and resize pass an empty artifact and keep prior behavior. Timeouts distinguish “never promoted” vs “promoted but still draining”; **`pollWorkload`** retries transient poll errors like log follow.
> 
> **`workload up`** threads the candidate through **`settle`** / **`awaitRunning`** so **`--lock`** runs after the wait and locks the version actually serving, not the one being rolled off. Roll flows show **“Waiting for the new version to serve”** when appropriate.
> 
> **`WaitForReplacement`** requires **three** consecutive 404s before treating a seeded-but-never-seen rollout as settled, avoiding an early exit when the replacement record has not caught up right after POST.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 34e8bd7173ec3a8bf6a7dcd59efc723e91bcdf39. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/internal/workload/proton.go
+++ b/internal/workload/proton.go
@@ -1,0 +1,268 @@
+// Copyright 2026 DataRobot, Inc. and its affiliates.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package workload
+
+import (
+	"fmt"
+	"strings"
+	"time"
+
+	"github.com/datarobot/cli/internal/config"
+	"github.com/datarobot/cli/internal/drapi"
+)
+
+// Proton statuses observed live during a rolling replacement. An unrecognised
+// one is not assumed to mean anything.
+const (
+	ProtonStatusRunning  = "running"
+	ProtonStatusDraining = "draining"
+)
+
+// ProtonRoleActive marks the generation the platform considers to be serving;
+// the others carry no role. Sharper than "which one is running", since a
+// candidate runs for a while before it is promoted.
+const ProtonRoleActive = "active"
+
+// maxProtonPages bounds the next-link walk. A workload has one generation most
+// of the time and two during a replacement, so anything approaching this is a
+// route that is not ending rather than a workload that is unusually busy.
+const maxProtonPages = 20
+
+// Proton is one generation of a workload's running containers: the thing that
+// answers the endpoint. A workload has one most of the time and two during a
+// rolling replacement, the outgoing one in "draining". It is the only place the
+// platform says whether the previous version has stopped serving.
+type Proton struct {
+	ID         string    `json:"id"`
+	Name       string    `json:"name"`
+	WorkloadID string    `json:"workloadId"`
+	ArtifactID string    `json:"artifactId"`
+	Status     string    `json:"status"`
+	Role       string    `json:"role"`
+	CreatedAt  time.Time `json:"createdAt"`
+	UpdatedAt  time.Time `json:"updatedAt"`
+}
+
+type protonList struct {
+	Data       []Proton `json:"data"`
+	Count      int      `json:"count"`
+	TotalCount int      `json:"totalCount"`
+	Next       string   `json:"next"`
+}
+
+// IsDrainingProtonStatus reports whether s is a generation on its way out that
+// is still answering requests.
+func IsDrainingProtonStatus(s string) bool {
+	return strings.EqualFold(s, ProtonStatusDraining)
+}
+
+// IsRunningProtonStatus reports whether s is serving and not on its way out.
+func IsRunningProtonStatus(s string) bool {
+	return strings.EqualFold(s, ProtonStatusRunning)
+}
+
+// IsActiveProtonRole reports whether r marks the serving generation.
+func IsActiveProtonRole(r string) bool {
+	return strings.EqualFold(r, ProtonRoleActive)
+}
+
+// ListProtons fetches the generations currently attached to a workload.
+//
+// It follows next-links like every other list in this package. A page-one-only
+// read would be worse here than elsewhere: the generation being replaced is the
+// oldest, so a truncated page can omit exactly the draining proton the wait
+// exists to see, and the wait would settle early with no error to show for it.
+//
+// The walk is bounded. It runs inside a poll predicate, which sits downstream
+// of the deadline check, so a cursor that points back at itself would spin
+// past --poll-timeout with no way out. Every other list here is bounded by a
+// caller-supplied limit; this one has no limit to pass, so it counts pages.
+//
+// A body carrying no recognisable list is an error rather than an empty one.
+// Decoding into protonList succeeds for any JSON object, so a route that
+// answers a different shape would otherwise read as "no generations" and settle
+// the wait silently, which is the one failure mode that looks exactly like
+// success.
+func ListProtons(workloadID string) ([]Proton, error) {
+	pageURL, err := config.GetEndpointURL("/api/v2/workloads/" + escapeID(workloadID) + "/protons/")
+	if err != nil {
+		return nil, err
+	}
+
+	var all []Proton
+
+	for page := 0; pageURL != ""; page++ {
+		if page >= maxProtonPages {
+			return nil, fmt.Errorf(
+				"proton list for workload %s did not end after %d pages; refusing to follow more",
+				workloadID, maxProtonPages)
+		}
+
+		var list protonList
+
+		if err := drapi.GetJSON(pageURL, "protons", &list); err != nil {
+			return nil, err
+		}
+
+		if list.Data == nil && list.TotalCount == 0 && list.Count == 0 {
+			return nil, fmt.Errorf("proton list for workload %s carried no data field", workloadID)
+		}
+
+		all = append(all, list.Data...)
+
+		if list.Next == "" {
+			break
+		}
+
+		if err := drapi.AssertNextOnSameHost(list.Next); err != nil {
+			return nil, err
+		}
+
+		pageURL = list.Next
+	}
+
+	return all, nil
+}
+
+// protonsServing reports whether the generation the caller waits for is the one
+// answering, and the one it replaced has stopped.
+//
+// This is what the workload document cannot answer. Measured live: at t+110s it
+// names the NEW artifact, the old generation starts draining and the new one is
+// marked active, yet the endpoint answers with the old build until t+171s; the
+// old generation only leaves draining at t+530s.
+//
+// A predecessor blocks the wait while it is running or draining, which are the
+// two states in which it was observed still answering. Both matter: at t+27s
+// both generations were running and the endpoint was still on the old build.
+// Anything past those is treated as gone, because an outgoing generation sits
+// in stopping long after it stopped mattering and waiting for the list to empty
+// would hold a finished deploy open for minutes.
+//
+// Known limitation: a predecessor status this code has not seen is read as
+// gone. The alternative, blocking on anything unrecognised, trades a rare early
+// exit for a deploy that hangs to its timeout on a status the platform added.
+//
+// Without an artifact named, which is a resize, both generations share one and
+// draining is the only thing that tells them apart. An empty list settles that
+// case rather than blocking, so a cluster whose route answers differently does
+// not hang; with an artifact named it does not, since an empty list is also
+// what the gap between one generation being collected and the next appearing
+// looks like.
+func protonsServing(protons []Proton, want Serving) protonVerdict {
+	if len(protons) == 0 {
+		// An empty list is two things at once: a route with nothing to say,
+		// and the blink between one generation being collected and the next
+		// being listed. With an artifact named the second reading matters, so
+		// it is reported as unresolved rather than settled, and the caller
+		// gives it a bounded number of polls to resolve before accepting it.
+		return protonVerdict{Serving: want.ArtifactID == "", Empty: true}
+	}
+
+	if want.AwaitDrain && anyServingPredecessor(protons, want.ArtifactID) {
+		return protonVerdict{}
+	}
+
+	if want.ArtifactID == "" {
+		return protonVerdict{Serving: true}
+	}
+
+	// The platform's own answer, when it gives one.
+	if at := activeProtonIndex(protons); at >= 0 {
+		active := protons[at]
+
+		return protonVerdict{
+			Serving: active.ArtifactID == want.ArtifactID && IsRunningProtonStatus(active.Status),
+		}
+	}
+
+	// No role reported, so fall back to whether anything runs the version.
+	return protonVerdict{Serving: anyRunning(protons, want.ArtifactID)}
+}
+
+// protonVerdict is what the proton list said. Empty separates "the list is
+// there and the handover is not done" from "the list said nothing at all",
+// which are the same answer to the wait and different answers to how long it
+// is worth waiting.
+type protonVerdict struct {
+	Serving bool
+	Empty   bool
+}
+
+// anyServingPredecessor reports whether a generation other than the one being
+// waited for is still taking requests.
+//
+// Draining and running both count. Measured at t+27s of a rollout, both
+// generations were running and the endpoint was still on the old build, so
+// treating running as settled would return before the handover.
+//
+// The successor is identified by artifact when one is named, and otherwise by
+// the active role, which is what makes this work for a resize: a resize
+// replaces a workload with itself, so both generations carry the same artifact
+// and only the role separates them. On an install that reports no role, a
+// resize falls back to draining alone, since there is then nothing to tell a
+// predecessor from its replacement.
+func anyServingPredecessor(protons []Proton, wantArtifactID string) bool {
+	// By index, not by id. Identifying the successor by Proton.ID would treat
+	// every generation as the successor on any response that omits the field.
+	activeAt := activeProtonIndex(protons)
+
+	for i := range protons {
+		p := &protons[i]
+
+		if wantArtifactID != "" && p.ArtifactID == wantArtifactID {
+			continue
+		}
+
+		if i == activeAt {
+			continue
+		}
+
+		if IsDrainingProtonStatus(p.Status) {
+			return true
+		}
+
+		// Running only counts as a predecessor once something identifies the
+		// successor, or every generation would look like one.
+		if IsRunningProtonStatus(p.Status) && (wantArtifactID != "" || activeAt >= 0) {
+			return true
+		}
+	}
+
+	return false
+}
+
+// activeProtonIndex is where the generation marked as serving sits, or -1 when
+// none is marked.
+func activeProtonIndex(protons []Proton) int {
+	for i := range protons {
+		if IsActiveProtonRole(protons[i].Role) {
+			return i
+		}
+	}
+
+	return -1
+}
+
+// anyRunning reports whether some generation runs wantArtifactID.
+func anyRunning(protons []Proton, wantArtifactID string) bool {
+	for i := range protons {
+		if protons[i].ArtifactID == wantArtifactID && IsRunningProtonStatus(protons[i].Status) {
+			return true
+		}
+	}
+
+	return false
+}
--- a/internal/workload/replacement.go
+++ b/internal/workload/replacement.go
@@ -42,18 +42,10 @@ import (
 // WaitForReplacement reports the rollout as a success. That "errored" had to be
 // added at all is proof the vocabulary was incomplete once already.
 //
-// A roll no longer rests on this. The deploy confirms after the fact which
-// artifact the workload ended up running, so a rollout this file mistook for a
-// success is caught by the wait that follows it rather than reported as one.
-// See WaitForWorkload's wantArtifactID.
-//
-// A resize still does rest on it, and that is the limitation that remains.
-// Applying new sizing replaces the workload with itself on the artifact it is
-// already running, so there is no artifact change for anything downstream to
-// confirm, and the record's own status is the only account of how it went. The
-// only other candidate signal is the runtime block converging on what was sent,
-// which cannot be trusted without knowing how the platform normalises it:
-// guessing there would trade a rare false success for a routine false failure.
+// A roll no longer rests on this: the deploy confirms afterwards which artifact
+// the workload ended up running (see WaitForWorkload). A resize has no artifact
+// change to confirm, and is covered instead by that same wait refusing to
+// settle while a generation is still draining.
 const (
 	ReplacementStatusCompleted = "completed"
 	ReplacementStatusFailed    = "failed"
@@ -333,7 +325,10 @@ func confirmWorkloadExists(workloadID string) error {
 //   - the settled replacement and an error, when it failed, so the caller can
 //     still show what happened
 //   - the last-seen replacement and an error, on timeout, for the same reason
-//   - nil and an error, when a poll failed or nothing was ever active
+//   - the last-seen replacement and an error, when polling failed; that is the
+//     seed when nothing was ever observed, so its status is what StartReplacement
+//     answered rather than anything this wait saw
+//   - nil and an error, when nothing was ever active
 //
 // The two 404 cases mean opposite things, which is why this is not a plain
 // poll loop. Seen after a status, a 404 means the platform settled the
@@ -342,14 +337,10 @@ func confirmWorkloadExists(workloadID string) error {
 // fast rollout before the first poll landed; passing started is what lets the
 // wait tell those from each other.
 //
-// A 404 that arrives before any record has been watched has to earn its
-// meaning, because the first poll fires immediately after the POST that created
-// the record and a route that has not caught up yet answers exactly the same
-// way as one whose rollout is over. Believing it at once turned a swap that had
-// barely started into a wait that returned in no time at all, reporting a
-// rollout as finished while the endpoint went on serving the previous version
-// for as long as the swap really took. So an unwatched absence is counted
-// rather than acted on, and only a run of them settles the wait.
+// A 404 before any record has been watched has to earn its meaning: the first
+// poll fires immediately after the POST, and a route that has not caught up
+// answers exactly like one whose rollout is over, so it is counted rather than
+// acted on.
 //
 // A settled rollout does not guarantee a terminal Status on the way out. When
 // the record is collected before the first poll lands, the seed is the only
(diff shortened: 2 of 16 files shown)


internal/workload/workload.go
```diff
@@ -406,6 +428,56 @@ func WaitForWorkload(
...
+}
+
+// servingWanted asks the proton list whether the previous generation has
+// stopped answering. A 404 leaves the wait on the artifact check alone, since
+// this route is not on every cluster and a missing one must not fail every
+// rollout there. Other failures are real poll errors.
+func servingWanted(workloadID, wantArtifactID string) (bool, error) {
+	protons, err := ListProtons(workloadID)
+	if err != nil {
+		if isNotFound(err) {
```
adamalpi: **[Low] A 403 on the protons route fails the deploy the route was only corroborating**

Only a 404 falls back to the artifact-only check; a 403 (a role that can deploy but not list protons, or an RBAC-restricted install) is a non-transient poll error, so after the budget the deploy *reports failure while the rollout proceeds server-side*. If a protons-forbidden-but-deploy-capable role is real anywhere, 403 deserves the same fallback as 404 — this route is corroboration, not the rollout itself.

internal/workload/proton.go
```diff
@@ -0,0 +1,158 @@
...
+// generation still taking requests. Stopping does not block: an outgoing
+// generation sits there long after it stopped mattering.
+//
+// With an artifact named something must also be serving it, so a rollout that
+// drained the old generation without bringing the new one up is not mistaken
+// for finished. Without a name, which is a resize, both generations share an
+// artifact and draining is the only handle. An empty list settles rather than
+// blocks, so a cluster whose route answers differently does not hang.
+func protonsServing(protons []Proton, wantArtifactID string) bool {
+	if len(protons) == 0 {
```
adamalpi, line 165: **[Low] The empty-list fallback also covers the eventual-consistency blink**

Settling on an empty list is deliberate and documented (odd clusters must not hang) — but it also covers the window where the drained generation is already collected and the new one not yet listed: a premature success in exactly the gap this PR exists to close. Cheap hardening: treat an empty list like the 404 fallback only after it repeats across a couple of polls, mirroring the `uncorroboratedAbsences` pattern this same PR introduces for the replacement route.

internal/workload/up/run.go
```diff
@@ -983,6 +997,18 @@ func awaitRunning(workloadID string, result Result, opts Options, report *report
 	return result, nil
 }
 
+// waitLabel names what is actually being waited for. On a roll the workload
+// never stopped running, so the old label described a wait that was over before
+// it began. result.ArtifactID is still the serving version here, so a want that
+// differs from it is precisely the roll case.
+func waitLabel(wantArtifactID string, result Result) string {
```
adamalpi: **[Nit] The label leans on `result.ArtifactID` still naming the outgoing version here**

True today per `settle`'s ordering, but it's an implicit coupling: a future reorder that moves the envelope update earlier silently flips this label back to the lie it replaces. Worth a test pinning "Waiting for the new version to serve" on the roll path so the coupling breaks loudly.

adamalpi (approved):
Verified the proton contract against the workload-api source: DRAINING, role active|candidate, and the route all exist exactly as consumed, and the WARMING candidate state is handled correctly by the IsRunningProtonStatus gate. The lock-after-wait fix alone justifies the PR. Two Lows and a Nit, none blocking.

internal/workload/up/run.go
```diff
@@ -945,8 +945,13 @@ func isConflict(err error) bool {
...
-func settle(workloadID string, result Result, opts Options, report *reporter) (Result, error) {
-	result, err := awaitRunning(workloadID, result, opts, report)
+//
+// wantArtifactID is what makes the lock safe as well as the report honest. lock
+// makes result.ArtifactID permanent, and that is whatever the wait last saw: a
+// wait that returned early would lock the artifact being rolled off, leaving it
+// undeletable while the version actually serving stays a draft.
+func settle(workloadID, wantArtifactID string, result Result, opts Options, report *reporter) (Result, error) {
+	result, err := awaitRunning(workloadID, wantArtifactID, result, opts, report)
 	if err != nil || !opts.Lock {
```
ajalon1: Blocking, and it's a merge hazard rather than a bug in what you wrote.

Main moved under this branch (#842 landed `verifyEndpoint`), and the merge conflicts on exactly this line. #842 split it into `if err != nil {`, then `verifyEndpoint(result, report)`, then `if !opts.Lock { return result, nil }`. Keep your side verbatim and the `|| !opts.Lock` early return jumps past `verifyEndpoint`, so every deploy without `--lock` silently loses the endpoint check. I tried that resolution locally: it builds clean. #842's two tests do catch it, but only after you widen the two `wait` fakes in `endpoint_check_test.go` (a file that did not exist when you wrote this), so the compiler points at a test file instead of at `settle`. Resolution that works: your signature and doc, main's body. That builds, passes `./internal/workload/...`, and lints clean on all three GOOS.

Separately, and this one survives a correct resolution: `verifyEndpoint` sits after this early return, so it never runs when the wait errors. Your new timeout message says "a previous version is still answering its endpoint", which is a claim about the endpoint inferred entirely from the proton list, while the one GET that could substantiate it is seven lines below and unreachable on that path. A 30 minute roll timeout is exactly when someone wants the endpoint's own answer.

internal/workload/workload.go
```diff
@@ -406,6 +428,56 @@ func WaitForWorkload(
...
+// empty want is a caller that changed none.
+func onWantedArtifact(wl *Workload, wantArtifactID string) bool {
+	return wantArtifactID == "" || wl.ArtifactID == wantArtifactID
+}
+
+// servingWanted asks the proton list whether the previous generation has
+// stopped answering. A 404 leaves the wait on the artifact check alone, since
+// this route is not on every cluster and a missing one must not fail every
+// rollout there. Other failures are real poll errors.
+func servingWanted(workloadID, wantArtifactID string) (bool, error) {
```
ajalon1: Blocking. This fallback now leaks into #842's output.

On a cluster without `/protons/` this returns true, so `stop` settles the moment the workload names the new artifact. Your own measurement puts that about a minute before the endpoint stops answering the old build, and four minutes in one run. `settle` then calls `verifyEndpoint`, which GETs inside that gap and prints `an anonymous GET answered HTTP 200 OK`, describing the version being rolled off, as confirmation of the deploy.

With the route present, the drain wait puts the GET after cutover and the line is correct. So the same transcript means two different things depending on the cluster, with nothing distinguishing them. The fallback probably needs to suppress or caveat that line, not only skip the drain wait.

Worth saying the merge also makes #842 better on the normal path: before this PR, `verifyEndpoint` on a roll GETted during the swap and reported the old build every time. Your drain wait fixes that. It is only the fallback that leaks.

internal/workload/workload.go
ajalon1: Only a 404 is forgiven, and the comment above says this route "is not on every cluster". The other ways a cluster says no all fall through: 403 (route exists, token lacks permission), 405 from a gateway that rejects unknown paths, 422 on the trailing slash. None are transient, so `pollWorkload` returns `poll workload <id>: 403 Forbidden` and fails a deploy where the build succeeded, the replacement completed, and the new version is serving. That deploy exited 0 before this PR.

Broadening the guard to any non-transient 4xx keeps the intent and loses the regression.

internal/workload/proton.go
```diff
@@ -0,0 +1,158 @@
...
+		return true
+	}
+
+	if anyDraining(protons) {
```
ajalon1: This runs before the `wantArtifactID == ""` return below it, so the drain check applies to every caller. `create` (`run.go:800`), `startFirst` (`run.go:843` and `846`) and `retune` (`retune.go:102`) all pass `""`, so all three now block until nothing is draining, up to the 30 minute `--poll-timeout`.

Your comment justifies this for resize only, and I read that as the intent. But as written, a `dr workload up` that starts a stopped workload whose predecessor has not finished draining waits where it used to return in seconds. The PR body's "a create, a start and a resize name no artifact and keep the behaviour they had" is not true for any of the three, so either the ordering or the claim needs to move.

internal/workload/proton.go
```diff
@@ -0,0 +1,158 @@
...
+		return false
+	}
+
+	if wantArtifactID == "" {
+		return true
+	}
+
+	// The platform's own answer, when it gives one.
+	if active := activeProton(protons); active != nil {
```
ajalon1: This settles while the predecessor is still running. Given `[{art-2, running, role=active}, {art-1, running, role=""}]` and `want=art-2`: `anyDraining` is false, `activeProton` finds art-2, it matches and it runs, so this returns true with art-1 still up and still answering. The `anyRunning` fallback on line 124 has the same shape, since it only asks whether something runs the wanted artifact and never what the other generations are doing.

The header says only two statuses were observed live and an unrecognised one is not assumed to mean anything, but here any predecessor status other than `draining` is treated as stopped. `TestProtonsServing` has no two-running case, so neither branch is covered. Whether this is reachable depends on question 1 above.

internal/workload/workload.go
```diff
@@ -427,41 +499,62 @@ func WaitForSteadyWorkload(
...
+
+				transient = 0
+			}
 		}
 
-		if stop(wl.Status) {
-			return wl, nil
+		if err != nil {
+			if !isTransientPollError(err) || transient >= maxTransientPollErrors {
+				return nil, fmt.Errorf("poll workload %s: %w", workloadID, err)
```
ajalon1: This discards the workload it just read. `stop` is now a second network call, so a non-transient failure there (or six transient ones) returns nil even though `GetWorkload` succeeded and `last` is sitting right there.

Downstream: `awaitRunning` leaves `Status`, `Endpoint` and `ArtifactID` unset so the envelope names nothing and `verifyEndpoint` has no endpoint to check, `unpromoted` short-circuits on `wl == nil` and can say nothing about how the rollout stalled, and the message blames the read that worked. Returning `last` and naming the protons route costs nothing.

internal/workload/replacement.go
```diff
@@ -353,43 +356,73 @@ func WaitForReplacement(
...
+	var (
+		lastSeen = started
+		watched  bool
+		absences int
+	)
 
 	for {
 		replacement, err := GetActiveReplacement(workloadID)
 		if err != nil {
 			return nil, fmt.Errorf("poll replacement for workload %s: %w", workloadID, err)
```
ajalon1: The transient tolerance went into `pollWorkload` but not here, and this is the wait that runs first in a roll. One 502, 504 or network blip still kills the whole deploy and leaves the swap in flight.

Your rationale for adding it reads identically for this loop: a roll now spends minutes on this route too, so the blip window you widened for covers it.

internal/workload/proton.go
```diff
@@ -0,0 +1,158 @@
...
+		return nil, err
+	}
+
+	var list protonList
+
+	if err := drapi.GetJSON(url, "protons", &list); err != nil {
+		return nil, err
+	}
+
+	return list.Data, nil
```
ajalon1: This cannot tell "no generations" from "a body we don't parse". `json.Decode` into `protonList` succeeds with `Data == nil` for any object without a `data` key, so `{"protons":[...]}`, `{"items":[...]}` and `{}` all return `(nil, nil)`, `protonsServing` takes the `len(protons) == 0` branch, and the wait settles on the first poll. Silently, unlike the 404 path.

`workload_test.go:987` demonstrates it live (see that comment). Cross-checking `totalCount`, or requiring the wanted artifact to be present before accepting an empty list, would fail loudly instead.

ajalon1 (commented):
Nice. Asking "is it running" during a rolling swap proves nothing — you're gonna need a bigger drain check. The old version doesn't stop serving just because you stopped asking about it; it's still out there, in the water, taking bites out of the endpoint minutes after everyone's declared it safe to go back in. The proton list being the only place the platform admits that matches what I'd expect, and the `--lock` fix that falls out of it is the part I'd have missed.

Comments are inline. Two I'd call blocking: the conflict with #842 in `settle`, and what the protons 404 fallback now does to #842's endpoint line. The rest are ordinary review.

Two questions I can't answer from the code:

1. Does the platform ever report a predecessor as `active` or `running` while the successor is up, or is `draining` set at promotion? If it's always set at promotion, the `protonsServing` comment below is theoretical.
2. Does `/workloads/{id}/protons/` paginate?

On the eight minutes: taking the honest wait over the early exit is the right call and the writeup makes the case. What I'd want alongside it is a heartbeat, because eight silent minutes and a hang look identical from outside, especially in CI.

internal/workload/proton.go
```diff
@@ -0,0 +1,158 @@
...
+	Name       string    `json:"name"`
+	WorkloadID string    `json:"workloadId"`
+	ArtifactID string    `json:"artifactId"`
+	Status     string    `json:"status"`
+	Role       string    `json:"role"`
+	CreatedAt  time.Time `json:"createdAt"`
+	UpdatedAt  time.Time `json:"updatedAt"`
+}
+
+type protonList struct {
```
ajalon1, line 58: This is the only list call in the package that ignores pagination. `execenv.go:190`, `build.go:307`, `credential.go:80`, `artifact.go:558` and `logs.go:203` all loop on `list.Next` through `drapi.AssertNextOnSameHost`, and this struct parses `count` and `totalCount` without using either, and doesn't declare `next` at all.

If the route paginates, page one can omit the draining predecessor, `protonsServing` settles early, and we're back to the bug this PR exists to fix with no error to show for it. `.cursor/bugbot-testing.md` also asks for single page, multiple pages, edge cases and host boundary coverage; `TestListProtons_ParsesTheList` is single page only. See question 2.

internal/workload/workload_test.go
```diff
@@ -822,7 +870,140 @@ func TestWaitForWorkload_Timeout(t *testing.T) {
...
+	wl, err := WaitForWorkload("wl-1", "art-2", time.Millisecond, time.Minute, nil)
+	require.Error(t, err)
+	require.NotNil(t, wl)
+	assert.Contains(t, err.Error(), WorkloadStatusErrored)
+	assert.NotContains(t, err.Error(), "timeout", "the failure is the answer, not something to wait out")
+}
+
+// These waits run for as long as a rollout takes, long enough for one bad
+// response to land between two healthy polls.
+func TestWaitForWorkload_RidesOutATransientPollError(t *testing.T) {
```
ajalon1: This passes for the wrong reason. The handler ignores `r.URL.Path`, so `GET /workloads/wl-1/protons/` returns the workload document, `ListProtons` decodes it into `protonList` with no error and an empty `Data`, and `protonsServing` settles on the empty-list branch.

So the test would still pass with `servingWanted` deleted entirely, and it asserts nothing about the drain check surviving a blip mid-wait, which is the interaction this PR introduces. `serveWorkloadAndProtons` already routes on the `/protons/` suffix; using it here fixes both.

internal/workload/up/run.go
```diff
@@ -955,11 +960,20 @@ func settle(workloadID string, result Result, opts Options, report *reporter) (R
...
+	workloadID, wantArtifactID string,
+	result Result,
+	opts Options,
+	report *reporter,
+) (Result, error) {
 	var final *workload.Workload
 
-	err := report.run("Waiting for the workload to run", func() error {
-		wl, waitErr := waitWorkloadFn(workloadID, opts.PollInterval, opts.PollTimeout, nil)
+	err := report.run(waitLabel(wantArtifactID, result), func() error {
```
ajalon1: `pollWorkload`'s transient-retry logic reuses `maxTransientPollErrors` from `logs.go` (same threshold, same reset-on-success) but drops the half of that pattern that matters: `FollowWorkloadLogs` calls `onWarn` on every retry so the user sees it happening; this swallows it silently. With `r.spinner` false the deploy emits zero bytes through every retry in the whole drain window.

`WaitForWorkload` already takes an `onTick` that could carry that signal (and could report the draining generation each poll, since this is now the longest phase in the deploy at 8m37s), but nothing here passes one.

internal/workload/workload.go
```diff
@@ -378,24 +378,46 @@ func GetWorkload(workloadID string) (*Workload, error) {
...
-// mirroring WaitForBuild.
+// wantArtifactID is empty when the caller changed no artifact: a create, a
+// start or a resize. A roll names its candidate, and needs to, because the
+// platform rolls by replacing the workload with itself and so reports "running"
+// from before the swap until after it, leaving a status-only wait to return on
+// the first poll having proved nothing. The artifact is still not sufficient on
+// its own, which is what protonsServing covers.
+//
+// onTick may be nil and is invoked after each poll, mirroring WaitForBuild.
 func WaitForWorkload(
```
ajalon1, line 425: Two same-typed id parameters side by side, so `WaitForWorkload(artifactID, workloadID, ...)` compiles and polls the wrong resource with nothing in the signature to catch it.

More to the point, `""` now carries three meanings: create (nothing to confirm), start (nothing to confirm), and resize (confirm nothing but wait out the drain). That overload is what makes the `proton.go:110` change easy to miss. A named mode, or a `WaitForServingArtifact` sibling, would make the resize case explicit, and would have kept this from breaking two fakes in a test file the PR never saw.

internal/workload/proton.go
```diff
@@ -0,0 +1,158 @@
...
+//
+// So draining is what this waits out, being the platform's word for a
+// generation still taking requests. Stopping does not block: an outgoing
+// generation sits there long after it stopped mattering.
+//
+// With an artifact named something must also be serving it, so a rollout that
+// drained the old generation without bringing the new one up is not mistaken
+// for finished. Without a name, which is a resize, both generations share an
+// artifact and draining is the only handle. An empty list settles rather than
+// blocks, so a cluster whose route answers differently does not hang.
```
ajalon1: Both new limitations are documented without a ticket: this fallback, and the unknown-replacement-failure-status hole you rewrote at `replacement.go:40`.

RAPTOR-19347 and RAPTOR-19734 are in the PR description, but `.cursor/bugbot-package-design.md` asks for them in the code, and the repo does it that way (`sync/diskspace_windows.go:23` and `sync/synclock_windows.go:22` both cite RAPTOR-16928). The next reader of this file won't have the PR body open.

internal/workload/workload.go
```diff
@@ -378,24 +378,46 @@ func GetWorkload(workloadID string) (*Workload, error) {
...
+
+		return servingWanted(workloadID, wantArtifactID)
+	}
+
+	wl, err := pollWorkload(workloadID, interval, timeout, stop, onTick)
+	if err != nil {
+		return wl, unpromoted(workloadID, wl, wantArtifactID, err)
+	}
+
+	if wl == nil {
```
ajalon1: Unreachable. `pollWorkload` returns a nil error only from `return wl, nil` inside the loop, where `wl` is always non-nil.

While you're here: the `IsTerminalWorkloadStatus` in `stop` above can only mean "running", because `IsWorkloadErrorStatus` already returned on the error statuses. Saying `running` outright would save the next reader deriving that a terminated workload never reaches `servingWanted`.

internal/workload/proton.go
ajalon1: Confirmed harder while checking whether this PR is consistent with the rest of the codebase: every list type in the package declares `Next` and loops on it — `BuildList`, `ArtifactList`, `CredentialList`, `WorkloadList`, `executionEnvironmentList`, no exceptions. `protonList` is the only one that doesn't. This isn't a style nit, it's 5-for-5 against 1.

internal/workload/up/run.go
ajalon1: Also from that consistency check: this is the first silent long-poll in the package to add a retry-count-based tolerance at all, and it borrowed the threshold from `logs.go` (`maxTransientPollErrors`) without the behavior that constant is attached to there (`onWarn` on every retry). `WaitForBuild` and `WaitForReplacement` still have zero transient tolerance and fail hard on the first error, so this one is also inconsistent with two of its three siblings in the other direction — worth deciding once, since both look like they'd want the same fix for the same reason.

internal/workload/workload.go
```diff
@@ -427,41 +499,62 @@ func WaitForSteadyWorkload(
...
+				transient = 0
+			}
 		}
 
-		if stop(wl.Status) {
-			return wl, nil
+		if err != nil {
+			if !isTransientPollError(err) || transient >= maxTransientPollErrors {
+				return nil, fmt.Errorf("poll workload %s: %w", workloadID, err)
+			}
```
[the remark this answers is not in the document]
wojtekwdr: Fixed. `pollWorkload` returns `last` on a failed poll now, so a read that worked is not thrown away by a later one that did not.

The second half of this went away rather than being fixed: `stop` no longer returns an error at all. A failure of the protons request is the corroboration falling over, not the deploy, so it is absorbed where it happens and the loop only ever reports on the workload read it names. That also removes the case where an error about `/protons/` came back labelled `poll workload`.

internal/workload/workload.go
wojtekwdr: Agreed, and broadened past 403. Any non-transient 4xx now degrades the same way 404 did: 403 for a token that may deploy but not list protons, 405 from a gateway that does not know the path, 422 on the trailing slash.

Went further after thinking about your framing of it as corroboration rather than the rollout. Nothing about this route can fail a deploy now. A 5xx or a network blip is retried and then degrades too, and so does a body we cannot parse, because a deploy that has built, promoted and started serving should not be reported as failed on account of the second request that was only checking.

internal/workload/proton.go
wojtekwdr: Good catch, and it was worse than a blink: with an artifact named the empty list never settled at all, so any install whose route answers empty turned a finished rollout into a full `--poll-timeout` hang.

Done the way you suggested. `protonsServing` returns a verdict with `Empty` separate from `Serving`, and the caller gives an absence `uncorroboratedAbsences` polls to resolve before accepting it, which is the same bound the replacement 404 already had.

internal/workload/up/run.go
wojtekwdr: Fair, and the coupling was looser than it looked: I had deleted that test in an earlier round of trimming. It is back, asserting `Waiting for the new version to serve` on the roll path.

The label also keys on `AwaitDrain` rather than the artifact now, which fixes a second case in the same place: a resize waits out a drain with no artifact to name, so it kept the wording the roll had just stopped using. It says `Waiting for the new settings to serve`, confirmed live through a 7m55s resize.

internal/workload/up/run.go
wojtekwdr: Resolved exactly as you describe: my signature and doc, main's body. `verifyEndpoint` runs before the `!opts.Lock` return, so a deploy without `--lock` keeps the check.

You were right about where the compiler points, too. `endpoint_check_test.go`'s two `wait` fakes needed widening and nothing else complained until they were.

On `verifyEndpoint` sitting after the early return on the error path: left as is, deliberately. The timeout message no longer claims anything about the endpoint it has not established, because it now derives from what the proton list actually shows: a predecessor still answering, nothing listed at all, or nothing running the wanted version yet. It only says that on a timeout; a wait whose polls failed carries its own failure and gets no diagnosis invented for it.

internal/workload/workload.go
wojtekwdr: Fixed, and thank you for chasing it into the other PR's output. That interaction would not have been visible from either side alone.

`OnUnconfirmed` now carries a reason string and `verifyEndpoint` prints it, so the transcript says the handover went unconfirmed and why. The reason matters as much as the fact: an install without the route and a token without permission are different problems for whoever reads it, and the old wording asserted the first for both.

Your other observation holds on the normal path: the drain wait puts that GET after cutover, so the endpoint line describes the new version now instead of the one being replaced.

internal/workload/workload.go
wojtekwdr: Fixed, and widened further than the 4xx you listed. Every non-transient refusal degrades, and the failures that are not refusals no longer fail the deploy either: 5xx, network errors and unparseable bodies are retried and then degrade.

The rule is now that nothing about this route can fail a deploy, which is what the doc claimed before the code matched it.

internal/workload/proton.go
wojtekwdr: You read the intent correctly and the code did not implement it. A drain wait went in front of every create and start.

The empty-string convention was the cause, so it is gone. There is a `Serving` struct now with `ArtifactID` and `AwaitDrain` as separate fields, so a create and a start ask for neither, a resize asks for the drain with no artifact, and a roll asks for both. `servingWanted` returns before spending a request when neither is set, which a staging create confirmed: no drain wait, no proton call, 6.6s.

The PR body's claim was wrong as written and is corrected.

internal/workload/proton.go
wojtekwdr: Reachable, and it happens on every roll. Probing a live rollout every 3s:

```
 43s  6198e2/running,  6b49cd/running*   endpoint: BUILD-1
 50s  6198e2/running*, 6b49cd/draining   endpoint: BUILD-1
579s  6198e2/running*, 6b49cd/stopping   endpoint: BUILD-2
```

`*` is `role: active`. At t+43s both are running and the old one is still active, so `anyRunning` was satisfied about five minutes before the endpoint moved. So this was not theoretical and the two-running case now has a test.

`anyServingPredecessor` treats draining and running alike, identifying the successor by artifact when one is named and by the active role otherwise. The role fallback is what makes a resize work at all, since both generations carry the same artifact there.

Fixing it turned up one of mine: the successor check compared `Proton.ID`, which some responses omit, so the guard was inert whenever the field was missing. It compares by index now.

internal/workload/workload.go
wojtekwdr: Fixed. `last` comes back with every error that has one, so the envelope keeps its status, endpoint and artifact, and `unpromoted` has something to reason about.

The rest of it dissolved instead: `stop` no longer returns an error, because a protons failure is absorbed where it happens rather than propagated. So the loop only reports on the workload read it names, which also answers the mislabelling half of this.

internal/workload/replacement.go
wojtekwdr: Right, and the rationale does read identically. Added.

Both sites share a `transientBudget` now rather than each re-deriving the counter, which also settles the `>= max` against `> max` drift you spotted between them.

internal/workload/proton.go
wojtekwdr: Fixed. A response carrying no recognisable list is an error rather than an empty one: `data` absent with `count` and `totalCount` both zero is refused by name instead of decoding to "no generations" and settling the wait silently.

You were right that this was the worst of the failure modes here, since it looks exactly like success.

internal/workload/proton.go
wojtekwdr: Fixed. It follows `next` through `AssertNextOnSameHost` like the other five.

To your question: staging does not paginate this route today. No `next`, and `limit` is ignored. I added the loop anyway because your argument holds regardless of today's behaviour, and a route that starts paginating later would put the bug back with nothing to show for it.

It is also capped at a page count, which the other five do not need. This loop runs inside the poll predicate, which sits downstream of the deadline check, so a cursor pointing at itself would spin past `--poll-timeout` with no way out but Ctrl-C. The others are bounded by a caller-supplied limit; there is no limit to pass here.

internal/workload/workload_test.go
wojtekwdr: You are right that it passed for the wrong reason, and the diagnosis was exact: the handler ignored the path, so `ListProtons` decoded a workload document into an empty list and the wait settled on the empty branch.

That test is gone and the ones around it route through `serveWorkloadAndProtons`. The blip-mid-wait interaction you point at has its own test now, and it asserts the deploy survives rather than that it fails, which is the behaviour that changed since.

internal/workload/up/run.go
wojtekwdr: Added. Every 30s while the wait runs, off the `onTick` that was already there and unused.

Two departures from `logs.go`. It reports elapsed rather than the artifact, because naming the version being waited for read as though it had arrived, which invites exactly the wrong question. And it carries the phase name, because `report.run` prints its label only on completion, so an unlabelled line sits under the previous phase's checkmark and reads as belonging to that one.

Live, on a resize:

```
waiting for the new settings to serve, 3m2s so far; the workload is running
```

internal/workload/workload.go
wojtekwdr: Both halves fixed by the same change. `WaitForWorkload` takes a `Serving` struct, so there are no longer two same-typed ids side by side, and the three meanings of `""` are three explicit fields.

Your last clause turned out to be the load-bearing one: the overload is exactly what made the ordering bug at `proton.go:110` easy to miss, and separating the fields is what surfaced it.

internal/workload/proton.go
wojtekwdr: This one I want to push back on rather than quietly skip. House style I work to is that ticket ids belong in git blame, not in comments: an id tells a future reader nothing the comment should already say, and the comment should stand once the ticket is closed and the id is dead. So both limitations are documented, unnumbered, and the PR body carries RAPTOR-19347 and RAPTOR-19734 for anyone who wants the history.

I take the point that the next reader of the file will not have the PR body open, and the precedent you cite is real. Happy to add them if you would rather follow `bugbot-package-design.md` here; say the word and they go in.

internal/workload/workload.go
wojtekwdr: Removed, and the derivation you point at is spelled out where the check happens: only `running` reaches that line, the error statuses having returned above it.

The `wl == nil` guard survives in `unpromoted`, where it is reachable, since a failed poll genuinely returns no workload.

internal/workload/proton.go
wojtekwdr: Taken, and it is 6-for-6 now. `protonList` declares `Next` and loops on it through `AssertNextOnSameHost` like the rest.

Worth recording that staging does not paginate this route today, so this is consistency and future-proofing rather than a live bug. The page cap on top is specific to this one: it runs inside a poll predicate, downstream of the deadline, so it needed its own bound.

internal/workload/up/run.go
wojtekwdr: Both halves of this landed, and I have half-settled the wider point rather than settled it.

The `onWarn` behaviour is back: the wait reports every 30s while it runs. The policy is a shared `transientBudget` now, so `pollWorkload` and `WaitForReplacement` no longer re-derive it, which took the three copies to two.

`WaitForBuild` still has no tolerance and `logs.go` still hand-rolls its own, so I have moved the inconsistency rather than removed it. Both sit outside what this PR touches, which is why I stopped there rather than widening the diff. You are right that it wants deciding once; happy to take both here if you would rather not leave them.

ajalon1: If you haven't already, feel free to tackle WaitForBuild and logs.go in a separate issue and PR.

internal/workload/proton.go
ajalon1: LOL. 🤖 is wrong. Need to update bugbot rules here, we want work tracked, but we don't need to have them in the docstrings. ETA: https://github.com/datarobot-oss/cli/pull/855

internal/workload/workload.go
```diff
@@ -406,6 +490,110 @@ func WaitForWorkload(
...
+
+	return protonsServing(protons, want), nil
+}
+
+// refusedRoute reports whether err is the route saying no in a way that will
+// not change on the next poll.
+func refusedRoute(err error) bool {
+	var httpErr *drapi.HTTPError
+
+	return errors.As(err, &httpErr) && httpErr.StatusCode >= 400 && httpErr.StatusCode < 500
```
ajalon1, line 542: [Low] 429 gets classified as a permanent refusal here, so a rate-limited protons route degrades on the first hit instead of taking the retry-then-degrade path the doc above promises — `isTransientPollError` in logs.go explicitly retries 429. Consider excluding it: `... < 500 && httpErr.StatusCode != http.StatusTooManyRequests`.

ajalon1 (approved):
LGTM — the drain wait and lock-after-wait fix are solid, and all prior review comments are resolved in the force-push. One inconsistency flagged inline, non-blocking.

PR #845, opened 2026-08-26
test(plugin): deflake the subprocess tests under load

adamalpi:
# RATIONALE

Three `internal/plugin` tests race their own subprocesses and lose under load — they blocked three unrelated CI/local runs today alone (`TestGetManifestValid` and `TestGetManifestWithConfiguredTimeout` on windows-latest and in local full-suite runs, `TestExecutePluginContextCancellation` locally). All three pass in isolation every time; only a loaded machine breaks them, which is what makes them flakes rather than failures.

## CHANGES

- The `ManifestTestSuite` now raises `plugin.manifest_timeout_ms` in `SetupTest`, the way every other suite in the file already does. Its tests are about parsing and error paths, not the deadline — but they ran mock plugins against the 500ms production default, and a loaded full-suite `-race` run (or Windows CI's slow process creation) can spend longer than that just spawning the shell.
- `TestGetManifestWithConfiguredTimeout` keeps proving the configured branch is honored, with a generous value instead of a 1000ms one that races the spawn.
- The two context-cancellation tests replace their blind `time.Sleep(500ms)` with a readiness handshake: the script touches a file once its SIGTERM trap is installed, and the test waits for the file before cancelling. Cancelled early, the default handler kills the shell and the assertion reads the race instead of the graceful-shutdown path it exists to test.

Verified: `go test ./internal/plugin/ -race -count=2` green, and a full `task test` sweep — the exact load condition that flaked all day — passes clean.

No production code changes; test-only.

## PR Automation

---

> [!NOTE]
> **Low Risk**
> No production code changes; only test timing and synchronization for plugin subprocess tests.
> 
> **Overview**
> **Test-only** changes to stop flaky `internal/plugin` failures when CI or full `-race` runs are slow to spawn mock shell scripts.
> 
> `ManifestTestSuite` now sets `plugin.manifest_timeout_ms` to **5000** in `SetupTest`, matching other discovery suites, so manifest parsing tests are not tripped by the **500ms** production default on Windows or under load. `TestGetManifestWithConfiguredTimeout` still exercises the configured-timeout path but uses a **15s** value so spawn latency does not race the assertion.
> 
> The two Unix context-cancellation tests drop fixed **500ms** sleeps in favor of a **readiness file** touched after the SIGTERM trap is installed, plus a shared **`waitForFile`** helper that polls up to 10s. Cancellation runs only after the subprocess is ready, so assertions target graceful SIGTERM handling instead of startup races.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit d172bfde9ec7ea71074e5070121a0b0665e9e8c6. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/internal/plugin/exec_test.go
+++ b/internal/plugin/exec_test.go
@@ -198,12 +198,17 @@ func TestExecutePluginContextCancellation(t *testing.T) {
 		t.Skip("SIGTERM handling is Unix-specific")
 	}
 
-	// Create a script that traps SIGTERM, writes a marker, and exits cleanly
+	// A script that traps SIGTERM, writes a marker, and exits cleanly. It
+	// also announces readiness: the trap has to be installed before the
+	// cancel fires, or the default handler kills the shell and the exit code
+	// asserts on the race rather than on the graceful path.
 	markerFile := filepath.Join(t.TempDir(), "sigterm-received")
+	readyFile := filepath.Join(t.TempDir(), "ready")
 	script := fmt.Sprintf(`#!/bin/sh
 trap 'touch %s; exit 55' TERM
+touch %s
 while true; do sleep 0.1; done
-`, markerFile)
+`, markerFile, readyFile)
 
 	scriptPath := filepath.Join(t.TempDir(), "trap-term.sh")
 	createScript(t, scriptPath, script)
@@ -216,8 +221,7 @@ while true; do sleep 0.1; done
 		done <- ExecutePlugin(ctx, PluginManifest{}, scriptPath, []string{}, nil)
 	}()
 
-	// Give the subprocess time to start and install its signal handler
-	time.Sleep(500 * time.Millisecond)
+	waitForFile(t, readyFile)
 
 	cancel()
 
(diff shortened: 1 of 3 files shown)


internal/plugin/exec_test.go
```diff
@@ -216,8 +221,7 @@ while true; do sleep 0.1; done
 		done <- ExecutePlugin(ctx, PluginManifest{}, scriptPath, []string{}, nil)
 	}()
 
-	// Give the subprocess time to start and install its signal handler
-	time.Sleep(500 * time.Millisecond)
+	waitForFile(t, readyFile)
```
ajalon1, line 224: Minor (not blocking): if `waitForFile` hits its 10s deadline and calls `t.Fatalf`, `cancel()` below never runs and the mock script's `while true` loop is orphaned (previously the unconditional `cancel()` always fired, even on the failure path). Registering the cancel func as a cleanup closes it — `t.Cleanup` runs even when `Fatalf` triggers `Goexit`, and cancelling drives `ExecutePlugin`'s normal shutdown (SIGTERM, then SIGKILL via `WaitDelay` for the trap-ignoring script), reaping the subprocess:

```go
ctx, cancel := context.WithCancel(context.Background())
t.Cleanup(cancel)
```

The explicit `cancel()` can stay (double-cancel is a no-op) — it documents when cancellation is the point of the test. Same applies to `TestExecutePluginContextCancellationKillsUnresponsive` at line 267.

ajalon1 (approved):
Thanks, we've been cleaning up the unit test suite the last few weeks. Much appreciated. Consider the one shell process cleanup from the 🤖 review

ajalon1:
Going to merge this in if it passes smoke tests @adamalpi

PR #846, opened 2026-08-26
[RAPTOR-19729] fix(workload): recovery never ends at the file

adamalpi:
# RATIONALE

Recovering from a stuck workload required insider knowledge, and the CLI's own messages formed a loop. The errored refusal ended at "check the logs" with no way forward; the folk remedy — hand-deleting the `workloadId` line — sent the next `up` into a create, the create into a 409, and the 409's advice was to set the exact line just deleted. The other remedy in circulation, deleting the `.datarobot` state directory, only "works" because the wizard then lets a different name be picked — and it throws the code catalog and sync state away with it. The supported exit, `dr workload delete <id> --dir .` (which clears the binding too), was undiscoverable.

Ticket: [RAPTOR-19729](https://datarobot.atlassian.net/browse/RAPTOR-19729)

## CHANGES

- The **errored refusal carries its own exit**: check `dr workload logs <id>` first (a slow start can recover), and if it stays errored, `dr workload delete <id> --dir <path>` — which also clears the binding — then deploy again to recreate the name. The `--dir` suffix is computed the same way the terminated refusal already computes it, so the command is correct from any working directory.
- The **name-conflict (409) refusal reads the holder's status before advising**. Binding advice ("set `workloadId: <id>`") is only given for a workload a deploy can act on. An errored or terminated holder gets the delete advice instead — recommending the state being refused was what closed the loop.

Narrower than the ticket text because main has moved since it was filed: the terminated refusal already names the delete-with-binding-cleanup exit, and a missing workload now falls through to a clean re-create. This PR closes the two remaining holes — the errored dead end and the status-blind 409 — after which no `up` recovery path ends at "edit or delete the file".

Wording note: the advice prints `dr workload delete <id> --dir <path>`, which is correct today and stays correct after #843 (where the id becomes optional); once #843 lands, a follow-up could shorten the printed command.

## PR Automation

---

> [!NOTE]
> **Low Risk**
> Changes are limited to CLI refusal messages and conflict-handling branches in `up`; no auth, API contracts, or deploy mutations beyond which error text is shown.
> 
> **Overview**
> **`dr workload up` error text now walks users through recovery** instead of stopping at “check the logs” or suggesting manifest edits that loop back into the same failure.
> 
> When deploy refuses an **errored** bound workload, the message now includes `dr workload logs <id>`, then—if it stays broken—`dr workload delete <id> --dir <path>` (same `--dir` suffix as the terminated case) so delete clears the manifest binding and redeploy can recreate the name.
> 
> On **create name conflict (409)**, `nameTaken` looks up the existing holder’s status. **Errored or terminated** holders get delete-and-redeploy guidance, not `workloadId` binding advice (which would only hit the same refusal again). Healthy holders still get the bind-to-existing-id path. Tests lock in both behaviors.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit e40bb89129b0e4a8f793465fc9956f85b4f031e7. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/internal/workload/up/run.go
+++ b/internal/workload/up/run.go
@@ -700,9 +700,9 @@ func deployable(live Live, workloadName, dirFlag string) error {
 	switch live.State {
 	case StateTerminated:
 		return fmt.Errorf(
-			"workload %s is terminated, which cannot be undone, and it still holds its name and artifact. "+
-				"Remove it with 'dr workload delete %s%s', which also clears the binding in %s, then deploy again",
-			workloadName, live.WorkloadID, dirFlag, manifest.FileName)
+			"workload %s is terminated, which cannot be undone, and it still holds its name and artifact — "+
+				"%s, then deploy again",
+			workloadName, deleteRemedy(live.WorkloadID, dirFlag, true))
 
 	case StateErrored:
 		// The refusal carries its own exit. Ending at "check the logs" left
@@ -711,10 +711,8 @@ func deployable(live Live, workloadName, dirFlag string) error {
 		// name conflict below or throws away the code catalog with it.
 		return fmt.Errorf(
 			"workload %s is errored, so there is nothing safe to deploy onto. "+
-				"Check 'dr workload logs %s' first; a slow start can report errored and then recover. "+
-				"If it stays errored, remove it with 'dr workload delete %s%s', which also clears the "+
-				"binding in %s, and deploy again to recreate it under the same name",
-			workloadName, live.WorkloadID, live.WorkloadID, dirFlag, manifest.FileName)
+				"%sIf it stays errored, %s, and deploy again to recreate it under the same name",
+			workloadName, erroredCaveat(live.WorkloadID), deleteRemedy(live.WorkloadID, dirFlag, true))
 
 	case StateSettling:
 		// The backstop, not the answer. Every run waits a moving workload out
@@ -929,13 +927,20 @@ func nameTaken(createErr error, workloadName, path, dirFlag string) error {
 		// they just deleted. The advice is the one those refusals give:
 		// delete what holds the name.
 		if workload.IsWorkloadErrorStatus(existing[i].Status) {
+			// Errored gets the same hedge deployable's own refusal carries;
+			// terminated really is final, so it keeps the certainty.
+			caveat := ""
+			if !strings.EqualFold(existing[i].Status, workload.WorkloadStatusTerminated) {
+				caveat = erroredCaveat(existing[i].ID)
+			}
+
 			return fmt.Errorf(
 				"a workload named %s already exists (%s), is %s, and nothing was deployed onto it. "+
 					"A deploy cannot act on it, so binding to it would only move this refusal. "+
-					"If it is this project's dead workload, remove it with 'dr workload delete %s%s' "+
+					"%sIf it is this project's dead workload, %s "+
 					"and deploy again to recreate the name. If it is not, rename this one in %s: %w",
 				workloadName, existing[i].ID, strings.ToLower(existing[i].Status),
-				existing[i].ID, dirFlag, path, createErr)
+				caveat, deleteRemedy(existing[i].ID, dirFlag, false), path, createErr)
 		}
 
 		return fmt.Errorf(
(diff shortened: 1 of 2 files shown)


internal/workload/up/run.go
```diff
@@ -914,6 +921,23 @@ func nameTaken(createErr error, workloadName, path string) error {
...
+		// holder sends the next run straight into that state's refusal — and
+		// for the user who cleared the binding by hand because their workload
+		// was stuck, it closes a loop: the 409 recommending the exact line
+		// they just deleted. The advice is the one those refusals give:
+		// delete what holds the name.
+		if workload.IsWorkloadErrorStatus(existing[i].Status) {
+			return fmt.Errorf(
+				"a workload named %s already exists (%s), is %s, and nothing was deployed onto it. "+
+					"A deploy cannot act on it, so binding to it would only move this refusal. "+
+					"If it is this project's dead workload, remove it with 'dr workload delete %s%s' "+
```
ajalon1: This match comes from `listWorkloadsFn(conflictSearchLimit, nil, "")`, which 🤖  tells me reads org-wide, not scoped to this project.

Is there a real chance two projects share a name and this can point at someone else's dead workload? If between-org collision aren't possible, then no worries, just curious.

internal/workload/up/run.go
```diff
@@ -914,6 +921,23 @@ func nameTaken(createErr error, workloadName, path string) error {
...
+		// binding. Advising 'set workloadId' against an errored or terminated
+		// holder sends the next run straight into that state's refusal — and
+		// for the user who cleared the binding by hand because their workload
+		// was stuck, it closes a loop: the 409 recommending the exact line
+		// they just deleted. The advice is the one those refusals give:
+		// delete what holds the name.
+		if workload.IsWorkloadErrorStatus(existing[i].Status) {
+			return fmt.Errorf(
+				"a workload named %s already exists (%s), is %s, and nothing was deployed onto it. "+
+					"A deploy cannot act on it, so binding to it would only move this refusal. "+
```
ajalon1, line 939: This treats errored the same as terminated (both pass `IsWorkloadErrorStatus`), but `deployable()`'s own `StateErrored` case a few lines up hedges that "a slow start can report errored and then recover." Should this branch carry the same caveat for errored (vs. terminated, which really is final), or is a conflicting workload in a different enough situation that the certainty here is fine?

internal/workload/up/run.go
```diff
@@ -704,10 +705,16 @@ func deployable(live Live, workloadName, dirFlag string) error {
...
+		// The refusal carries its own exit. Ending at "check the logs" left
+		// recovery to folklore, and the folk remedy — hand-deleting the
+		// workloadId, or the whole state directory — either loops through the
+		// name conflict below or throws away the code catalog with it.
 		return fmt.Errorf(
 			"workload %s is errored, so there is nothing safe to deploy onto. "+
-				"Check 'dr workload logs' first; a slow start can report errored and then recover",
-			workloadName)
+				"Check 'dr workload logs %s' first; a slow start can report errored and then recover. "+
+				"If it stays errored, remove it with 'dr workload delete %s%s', which also clears the "+
```
ajalon1: Minor: "remove it with `dr workload delete <id><dirFlag>`, clears the binding" is now hand-written three times (`StateTerminated` above, here, and the new branch around line 935). Not blocking, just flagging in case a small `deleteRemedy(id, dirFlag, path)` helper is worth it so a future wording tweak doesn't need three synced edits.

ajalon1 (commented):
Good stuff, I enjoy closing the UX problems. Two notes, just me checking a few things.

internal/workload/up/run.go
ajalon1: Also, no offset on ListWorkloads()? How do you get an arbitrary page, then?

ajalon1 (commented):
LGTM either way, defer to @wojtekwdr

internal/workload/up/run_test.go
```diff
@@ -630,6 +630,58 @@ func TestRun_ConflictNamesTheWorkloadThatOwnsTheName(t *testing.T) {
...
+					dead.Status = status
+
+					return []workload.Workload{dead}, nil
+				},
+			})
+
+			_, _, err := runIn(t, unboundImageManifest, Options{NonInteractive: true})
+			require.Error(t, err)
+
+			assert.Contains(t, err.Error(), "dr workload delete wl-dead", "the exit is the delete")
```
wojtekwdr: [medium] question: is the `--dir` suffix asserted anywhere? I stubbed `dirFlagFor` to return "" and, separately, swapped `dirFlag` and `path` in the new format args, and `go test ./internal/workload/up/` stayed green both times. Asserting on `dr workload delete wl-dead --dir ` here would pin the new parameter down.

wojtekwdr (approved):
LGTM

internal/workload/up/run.go
adamalpi: Cross-org, no: the API scopes every workload list to the caller's organization, so the search can never surface another org's workload. Within one org, yes — two projects can share a name, which is why the message never claims ownership: it offers both branches ("if it is this project's dead workload… if it is not, rename") and leaves the decision with the user rather than binding or deleting anything itself.

internal/workload/up/run.go
adamalpi: Good catch — it should carry the same hedge. Fixed in 087f5a7: an errored holder now gets "Check 'dr workload logs <id>' first; a slow start can report errored and then recover." before the delete advice (shared `erroredCaveat` helper, so the two messages can't drift), while terminated keeps the certainty since it really is final.

internal/workload/up/run.go
adamalpi: Extracted in 087f5a7 as `deleteRemedy(id, dirFlag, clearsBinding)` — the flag is why the wording differed at the third site: a create conflict has no binding to clear, so that clause only appears where one exists.

internal/workload/up/run_test.go
adamalpi: Pinned in 087f5a7: the conflict test now asserts `dr workload delete wl-dead --dir `, which catches both a stubbed-out `dirFlagFor` and swapped format args.

PR #848, opened 2026-08-26
[RAPTOR-19726] feat(workload): validate the project directory before the wizard asks

adamalpi:
# RATIONALE

Setup run from the wrong directory — the app's parent, most often — used to fail only at code-sync time, after six questions were answered and the artifact already existed. Nothing checked whether the directory looked like a project root; the only related check (the `pyproject.toml`/`uv.lock` one) lives in the sync engine, warn-only, and runs last. The check now happens before anything is asked, and — per the ticket's agreed direction — it **warns and offers, never fails**: a valid project cannot be reliably recognized, so a wrong suspicion has to cost one keystroke.

Ticket: [RAPTOR-19726](https://datarobot.atlassian.net/browse/RAPTOR-19726)

## CHANGES

- `Detect` now records which of the usual project files (Dockerfile, pyproject.toml, uv.lock, requirements.txt, package.json, go.mod, setup.py) the directory carries. When it carries none, it looks one level down for subdirectories that do — skipping hidden ones and ones already holding a manifest (those are entered with `--dir`, not set up afresh).
- **Interactive**: a suspect directory with candidates becomes the wizard's first question, ahead even of the binding — every suggestion the other screens show was read from the directory. The offer opens on the likeliest candidate ("./my-app — has Dockerfile, pyproject.toml"), "use this directory anyway" is one keystroke below, and choosing a candidate re-reads name, port, Dockerfile and .env from it. The manifest is then written where the flow ended up, and `up`'s wizard path follows the file it wrote instead of re-searching upward from where the command ran (which cannot see one level down).
- **Headless** (`--yes`, JSON, no terminal): a stderr warning naming the candidate directories and the `--dir` flag that takes them. Never a refusal, and silent for any directory that carries a project file.
- `wizard.SetInteractiveFlowForTest` gains a string return (the directory the flow settled on) — test-seam signature change, no production callers outside this repo.

## PR Automation

---

> [!NOTE]
> **Medium Risk**
> Changes which directory is detected, where the manifest is written, and how `up` loads after the wizard—directly affecting what code gets uploaded on deploy—but behavior is warn/offer only with broad test coverage.
> 
> **Overview**
> Setup and deploy used to assume the current directory was the project root, so running from a parent folder could waste the whole wizard and still sync the wrong tree. **This PR adds an early “wrong directory?” check that warns and offers a fix but never blocks.**
> 
> **Detection** now records usual root markers (Dockerfile, `pyproject.toml`, etc.). If the cwd has none, it scans immediate subdirectories (skipping hidden dirs and dirs that already have a manifest) and surfaces up to six candidates with the markers they carry.
> 
> **Interactive** runs show a new **directory screen first** when there are candidates—before binding or name—defaulting to the likeliest subfolder. Picking a subdirectory re-runs detection (name, port, Dockerfile, `.env`); staying put is one extra choice. The manifest is written under the directory the flow settled on, and **`dr workload up`** reloads from that manifest path after the inline wizard so deploy does not miss a manifest one level down.
> 
> **Headless** (`--yes`, JSON, no TTY) prints a stderr warning (and candidate names plus `--dir`) instead of prompting; runs still succeed. Test hooks like `SetInteractiveFlowForTest` now return the settled project directory as an extra string.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit f4e202cf376db986488b79a0e333fb3961430689. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/internal/workload/up/run.go
+++ b/internal/workload/up/run.go
@@ -489,8 +489,16 @@ func load(dir string, opts Options) (Loaded, error) {
 			"%w. Run 'dr workload config' to create one, then deploy", err)
 	}
 
-	if _, err := runWizardFn(wizard.Options{Dir: dir, Stderr: opts.Stderr}); err != nil {
+	setup, err := runWizardFn(wizard.Options{Dir: dir, Stderr: opts.Stderr})
+	if err != nil {
 		return Loaded{}, err
+	}
+
+	// The wizard's directory question may have moved the project, and the
+	// deploy has to follow the file it wrote: re-searching from where the
+	// command ran walks upward and cannot see a manifest one level down.
+	if setup.Path != "" {
+		return Load(filepath.Dir(setup.Path))
 	}
 
 	return Load(dir)
--- a/internal/workload/wizard/detect.go
+++ b/internal/workload/wizard/detect.go
@@ -82,11 +82,96 @@ type Detected struct {
 	// but it must be reported: silence here reads as "there was nothing to
 	// import" and the missing variables surface as a runtime failure instead.
 	EnvErr error
+	// RootMarkers is which of the usual project files the directory itself
+	// carries. See SuspectDir for what an empty list means.
+	RootMarkers []string
+	// Candidates are immediate subdirectories that do carry project files,
+	// looked for only when the directory itself carries none: the offer the
+	// wrong-directory question makes instead of assuming ".".
+	Candidates []DirCandidate
+}
+
+// DirCandidate is a subdirectory that looks like a project root when the
+// directory setup ran from does not.
+type DirCandidate struct {
+	// Rel is the candidate relative to the checked directory.
+	Rel string
+	// Markers is which project files it carries, so the offer says why it is
+	// being offered rather than presenting a bare path.
+	Markers []string
+}
+
+// rootMarkers are the files a project root usually carries. The list is for
+// suspicion, not proof, in both directions: a directory with none can still
+// be a deployable project, which is why nothing here refuses.
+var rootMarkers = []string{
+	DockerfileName, "pyproject.toml", "uv.lock", "requirements.txt", "package.json", "go.mod", "setup.py",
+}
+
+// maxDirCandidates caps the offer. Past a handful the list stops being an
+// answer and --dir is the tool.
+const maxDirCandidates = 6
+
+// markersIn is which of the usual project files dir carries.
+func markersIn(dir string) []string {
+	var found []string
+
+	for _, marker := range rootMarkers {
+		if fsutil.FileExists(filepath.Join(dir, marker)) {
+			found = append(found, marker)
+		}
+	}
+
+	return found
+}
+
+// dirCandidates lists the immediate subdirectories carrying project files.
+// Hidden directories are never the project, and one already holding a
+// manifest is skipped too: it is entered with --dir, not set up afresh.
+func dirCandidates(dir string) []DirCandidate {
+	entries, err := os.ReadDir(dir)
+	if err != nil {
+		return nil
+	}
+
+	var candidates []DirCandidate
+
+	for _, entry := range entries {
+		if !entry.IsDir() || strings.HasPrefix(entry.Name(), ".") {
+			continue
+		}
+
+		sub := filepath.Join(dir, entry.Name())
+		if fsutil.FileExists(manifest.Path(sub)) {
+			continue
+		}
+
+		markers := markersIn(sub)
+		if len(markers) == 0 {
+			continue
+		}
+
+		candidates = append(candidates, DirCandidate{Rel: entry.Name(), Markers: markers})
+		if len(candidates) == maxDirCandidates {
+			break
+		}
+	}
+
+	return candidates
 }
 
 // HasEnvFile reports whether the project has a .env worth asking about.
 func (d Detected) HasEnvFile() bool {
 	return len(d.EnvVars) > 0
+}
+
+// SuspectDir reports that the directory carries none of the usual project
+// files, which is what running setup from the wrong place — the app's
+// parent, most often — looks like. It is a suspicion, never a verdict: a
+// valid project cannot be reliably recognized, so the absence only ever
+// produces a warning and an offer, not a refusal.
+func (d Detected) SuspectDir() bool {
+	return len(d.RootMarkers) == 0
 }
 
 // Plural picks the singular or plural word for count. Exported so the
@@ -112,6 +197,13 @@ func Detect(dir string) Detected {
 	}
 
 	detected.EnvVars, detected.EnvErr = envVars(filepath.Join(dir, EnvFileName))
+
+	detected.RootMarkers = markersIn(dir)
+	if detected.SuspectDir() {
+		// The subdirectory scan runs only when there is something to suspect:
+		// a directory that looks like a project needs no offer.
+		detected.Candidates = dirCandidates(dir)
+	}
 
 	if !detected.HasDockerfile {
 		return detected
--- a/internal/workload/wizard/detect.go
+++ b/internal/workload/wizard/detect.go
@@ -23,6 +23,7 @@ import (
 	"strings"
 
 	"github.com/datarobot/cli/internal/fsutil"
+	"github.com/datarobot/cli/internal/log"
 	"github.com/datarobot/cli/internal/workload/manifest"
 	"github.com/joho/godotenv"
 )
@@ -89,6 +90,11 @@ type Detected struct {
 	// looked for only when the directory itself carries none: the offer the
 	// wrong-directory question makes instead of assuming ".".
 	Candidates []DirCandidate
+	// MoreCandidates reports that the candidate scan stopped early — at the
+	// offer cap or the entry budget — so the offer must say the list is not
+	// the whole answer: a project whose name sorts late would otherwise
+	// vanish without a trace, --dir unmentioned.
+	MoreCandidates bool
 }
 
 // DirCandidate is a subdirectory that looks like a project root when the
@@ -112,6 +118,13 @@ var rootMarkers = []string{
 // answer and --dir is the tool.
 const maxDirCandidates = 6
 
+// maxDirEntriesScanned caps the work, not just the offer: every entry
+// examined costs a manifest stat plus one per root marker, synchronously
+// before the wizard's first screen, and a directory with thousands of
+// entries (running setup from $HOME by mistake being the classic) must not
+// hang the wizard on stats.
+const maxDirEntriesScanned = 256
+
 // markersIn is which of the usual project files dir carries.
 func markersIn(dir string) []string {
 	var found []string
@@ -128,36 +141,81 @@ func markersIn(dir string) []string {
 // dirCandidates lists the immediate subdirectories carrying project files.
 // Hidden directories are never the project, and one already holding a
 // manifest is skipped too: it is entered with --dir, not set up afresh.
-func dirCandidates(dir string) []DirCandidate {
+// truncated reports that the scan stopped before the listing did, at either
+// cap, so a caller can say the offer is incomplete.
+//
+// The home directory is never scanned, same as repo.FindRepoRoot and
+// manifest.Locate stop there: everything under $HOME "carrying project
+// files" is not an offer, it is an inventory.
+func dirCandidates(dir string) (candidates []DirCandidate, truncated bool) {
+	if home, err := os.UserHomeDir(); err == nil && dir == home {
+		log.Debug("wizard: not scanning the home directory for project candidates")
+
+		return nil, false
+	}
+
 	entries, err := os.ReadDir(dir)
 	if err != nil {
-		return nil
+		log.Debug("wizard: cannot list directory for project candidates", "dir", dir, "err", err)
+
+		return nil, false
 	}
 
-	var candidates []DirCandidate
+	for i, entry := range entries {
+		if i == maxDirEntriesScanned {
+			return candidates, true
+		}
 
-	for _, entry := range entries {
-		if !entry.IsDir() || strings.HasPrefix(entry.Name(), ".") {
+		candidate, ok := candidateFrom(dir, entry)
+		if !ok {
 			continue
 		}
 
-		sub := filepath.Join(dir, entry.Name())
-		if fsutil.FileExists(manifest.Path(sub)) {
-			continue
-		}
-
-		markers := markersIn(sub)
-		if len(markers) == 0 {
-			continue
-		}
-
-		candidates = append(candidates, DirCandidate{Rel: entry.Name(), Markers: markers})
 		if len(candidates) == maxDirCandidates {
-			break
+			return candidates, true
 		}
+
+		candidates = append(candidates, candidate)
 	}
 
-	return candidates
+	return candidates, false
+}
+
+// candidateFrom examines one directory entry and reports the candidate it
+// yields, if any.
+func candidateFrom(dir string, entry os.DirEntry) (DirCandidate, bool) {
+	if !isSubdirectory(dir, entry) || strings.HasPrefix(entry.Name(), ".") {
+		return DirCandidate{}, false
+	}
+
+	sub := filepath.Join(dir, entry.Name())
+	if fsutil.FileExists(manifest.Path(sub)) {
+		return DirCandidate{}, false
+	}
+
+	markers := markersIn(sub)
+	if len(markers) == 0 {
+		return DirCandidate{}, false
+	}
+
+	return DirCandidate{Rel: entry.Name(), Markers: markers}, true
+}
+
+// isSubdirectory reports whether entry is a directory, following one level of
+// symlink: DirEntry.IsDir reflects the entry's own on-disk type, so a project
+// reached through a symlink would otherwise be invisible to the offer.
+func isSubdirectory(dir string, entry os.DirEntry) bool {
+	if entry.IsDir() {
+		return true
+	}
+
+	if entry.Type()&os.ModeSymlink == 0 {
+		return false
+	}
+
+	info, err := os.Stat(filepath.Join(dir, entry.Name()))
+
+	return err == nil && info.IsDir()
 }
 
 // HasEnvFile reports whether the project has a .env worth asking about.
@@ -202,7 +260,10 @@ func Detect(dir string) Detected {
 	if detected.SuspectDir() {
 		// The subdirectory scan runs only when there is something to suspect:
 		// a directory that looks like a project needs no offer.
-		detected.Candidates = dirCandidates(dir)
+		detected.Candidates, detected.MoreCandidates = dirCandidates(dir)
+
+		log.Debug("wizard: directory carries no project markers",
+			"dir", dir, "candidates", len(detected.Candidates), "more", detected.MoreCandidates)
 	}
 
 	if !detected.HasDockerfile {
--- a/internal/workload/wizard/detect_test.go
+++ b/internal/workload/wizard/detect_test.go
@@ -222,3 +222,44 @@ func TestParseProblem_UnrecognizablePayloadYieldsNoLine(t *testing.T) {
 
 	assert.Equal(t, `unexpected character "-" in variable name`, parseProblem(err, "A=1\n"))
 }
+
+// A directory with none of the usual project files is suspect, and the offer
+// is its marker-bearing subdirectories: hidden ones are never the project,
+// empty ones say nothing, and one already holding a manifest is entered with
+// --dir rather than set up afresh.
+func TestDetect_SuspectDirOffersItsProjectLookingSubdirectories(t *testing.T) {
+	dir := t.TempDir()
+
+	app := filepath.Join(dir, "my-app")
+	require.NoError(t, os.MkdirAll(app, 0o755))
+	require.NoError(t, os.WriteFile(filepath.Join(app, "Dockerfile"), []byte("FROM scratch\n"), 0o644))
+	require.NoError(t, os.WriteFile(filepath.Join(app, "pyproject.toml"), []byte("[project]\n"), 0o644))
+
+	hidden := filepath.Join(dir, ".cache")
+	require.NoError(t, os.MkdirAll(hidden, 0o755))
+	require.NoError(t, os.WriteFile(filepath.Join(hidden, "Dockerfile"), []byte("FROM scratch\n"), 0o644))
+
+	require.NoError(t, os.MkdirAll(filepath.Join(dir, "docs"), 0o755))
+
+	configured := filepath.Join(dir, "configured")
+	require.NoError(t, os.MkdirAll(configured, 0o755))
+	require.NoError(t, os.WriteFile(filepath.Join(configured, "go.mod"), []byte("module x\n"), 0o644))
+	require.NoError(t, os.WriteFile(manifest.Path(configured), []byte("name: x\n"), 0o644))
+
+	detected := Detect(dir)
+
+	assert.True(t, detected.SuspectDir())
+	require.Len(t, detected.Candidates, 1)
+	assert.Equal(t, "my-app", detected.Candidates[0].Rel)
+	assert.Equal(t, []string{"Dockerfile", "pyproject.toml"}, detected.Candidates[0].Markers)
+}
+
+// A directory carrying any project file is not suspect, and the subdirectory
+// scan never runs: a directory that looks like a project needs no offer.
+func TestDetect_MarkerBearingDirIsNotSuspect(t *testing.T) {
+	detected := Detect(writeDockerfile(t, t.TempDir(), "FROM scratch\n"))
+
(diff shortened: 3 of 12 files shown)


ajalon1:
Gonna look at this one more closely since it mirrors some of the detect headaches we had with dotenv and and component and template.

internal/workload/wizard/detect.go
ajalon1, line 1: I would highly recommend adding logging (both warn and debug levels) to this module. You will get plenty of errors and reports about this not working, and plenty more users will complain in silence. At least showing logs in stderr and in .dr-tui-debug.log will give them and us more confidenc.e

ajalon1: Also, I don't remember off hand if you have telemetry at the `cmd` level, but if you do, don't emit the user's working directory in telemetry. We're not currently allowed to do that.

The 🤖 tells me that `dir` is not tracked, which is good. Leave it that way!

internal/workload/wizard/detect.go
```diff
@@ -82,13 +82,98 @@ type Detected struct {
...
+	// being offered rather than presenting a bare path.
+	Markers []string
+}
+
+// rootMarkers are the files a project root usually carries. The list is for
+// suspicion, not proof, in both directions: a directory with none can still
+// be a deployable project, which is why nothing here refuses.
+var rootMarkers = []string{
+	DockerfileName, "pyproject.toml", "uv.lock", "requirements.txt", "package.json", "go.mod", "setup.py",
+}
```
ajalon1, line 115: This reminds me of how starship.rs modules check whether to render in a shell prompt.

We have a LOT of code in the `dependencies` modules that could be worth reusing here, and would make this a lot more resilient, extensible beyond what you've already configured, and helpful. @taras-pokornyy wrote most of that. ;)

ajalon1 (commented):
You will want to test that all of this code works in DR Codespaces, as well.

internal/workload/wizard/model.go
```diff
@@ -204,9 +209,21 @@ func defaultDraft(detected Detected) manifest.Draft {
 	return Answers{}.partialDraft(detected)
 }
 
-// first picks the opening screen. Binding is skipped when there is nothing to
-// bind to, or when a flag already said which workload this is.
+// first picks the opening screen. The directory question, when it exists,
+// comes before everything: a bound workload still syncs its code from here,
+// and every suggestion the other screens show was read from this directory.
 func (f flow) first(answers Answers) screen {
+	if f.detected.SuspectDir() && len(f.detected.Candidates) > 0 {
```
ajalon1, line 223: **Blocking.** This is where `screenDirectory` becomes reachable as the opening screen — and two other unconditional spots in this file don't know that happened, so the screen this line picks can end up never shown.

`newFlow` (line 199) sets `f.loading` whenever `f.pendingBind != ""` (i.e. `--workload-id`), with no check of what `f.at` actually is. `View()` renders only the loading spinner while `f.loading` is set, and `key()` swallows all input, so a `screenDirectory` picked here is built but never actually shown.

Then `liveLoaded()` (line 1126) does `f.at = screenKind` unconditionally once the live fetch resolves — no `f.at == screenDirectory` check, no call to `acceptDirectory()`. `f.detected` stays pointed at the original (possibly wrong) directory for the rest of the flow.

I reproduced this: build a flow with `Answers{WorkloadID: "..."}` from a suspect directory with candidates — `model.at == screenDirectory` while `loading` is set, a `down` keypress is a no-op, and after `liveLoaded()` fires, `f.at` is `screenKind` with `f.detected.Dir` unchanged. This is exactly the scenario the doc comment above (`first()`) calls out: *"a bound workload still syncs its code from here."* Right now it doesn't.

Suggest gating the unconditional pieces in `newFlow`/`liveLoaded` on `f.at`, or not starting `fetchLiveCmd` until `screenDirectory` has actually been resolved.

internal/workload/wizard/model.go
```diff
@@ -671,6 +693,28 @@ func (f *flow) acceptKind() (tea.Cmd, error) {
...
+// always the first screen, so no other answer exists yet and the draft is
+// rebuilt exactly the way newFlow built it — everything Detect suggests
+// (name, port, Dockerfile, .env) belonged to the old directory.
+func (f *flow) acceptDirectory() (tea.Cmd, error) {
+	chosen := f.choice.value()
+	if chosen == f.detected.Dir {
+		return nil, nil
+	}
+
+	f.detected = Detect(chosen)
```
ajalon1: **Blocking.** `acceptDirectory` mutates `f.detected` in place with nothing preserving the original. `directoryOptions()` (screens.go) rebuilds the choice list from `f.detected.Candidates`/`.Dir` every time the screen is entered — including via `back()`.

So: pick a candidate (say `api/`), advance to `screenName`, then press Escape to reconsider — `back()` re-enters `screenDirectory` through the same path, but `f.detected` now describes `api/` itself (not suspect anymore, so `Candidates` is empty). The user sees only "Use this directory anyway" pointing at `api/` — the original offer (`web/`, the suspect parent) is gone, with no way back short of cancelling the wizard entirely.

Worth snapshotting the pre-accept `Detected` somewhere `back()` can restore from, the way other screens preserve their option sets across back-navigation.

internal/workload/wizard/model.go
```diff
@@ -671,6 +693,28 @@ func (f *flow) acceptKind() (tea.Cmd, error) {
...
+
+	draft, err := f.answers.draft(f.detected)
```
ajalon1: **Blocking (simplification).** This is the same `draft`-falling-back-to-`partialDraft` branch `newFlow` already implements a few lines up (~line 177), just with the receiver swapped. Your own comment on `acceptDirectory` says as much ("the draft is rebuilt exactly the way newFlow built it") — which reads as a sign it should be a named, shared step rather than restated code:

```go
func (a Answers) draftOrPartial(detected Detected) manifest.Draft {
    if d, err := a.draft(detected); err == nil {
        return d
    }
    return a.partialDraft(detected)
}
```

called from both `newFlow` and `acceptDirectory`. Flagging as blocking mainly because it's cheap to fix alongside the `acceptDirectory` change above, and keeps the fallback rule from drifting out of sync between the two call sites.

internal/workload/wizard/detect.go
```diff
@@ -82,13 +82,98 @@ type Detected struct {
...
+		}
+	}
+
+	return found
+}
+
+// dirCandidates lists the immediate subdirectories carrying project files.
+// Hidden directories are never the project, and one already holding a
+// manifest is skipped too: it is entered with --dir, not set up afresh.
+func dirCandidates(dir string) []DirCandidate {
```
ajalon1: **Blocking.** Two related things on `dirCandidates`:

1. `maxDirCandidates` bounds candidates *found*, not entries *scanned* — every non-matching subdirectory still costs a manifest-shadow stat plus `markersIn`'s 7 sequential stats, unbounded, synchronous in `Detect()` before the wizard's first screen.
2. No boundary check for `$HOME` — which stands out because both pre-existing "does this look like a project" walks in this codebase *do* stop there: `repo.FindRepoRoot` ("reached home directory") and `manifest.Locate` ("if dir == home ... return ErrNotFound"). If setup runs from `$HOME` by mistake, this doesn't just cost time — it can offer an unrelated, stale project (some abandoned clone in `~/old-projects/` with a leftover Dockerfile) as a plausible-looking candidate, actively steering the user toward the wrong directory rather than just being slow about it.

Suggest a `dir == home` short-circuit (skip the scan, fall through to a plain warning) consistent with the rest of the codebase's treatment of `$HOME`.

internal/workload/wizard/detect.go
```diff
@@ -82,13 +82,98 @@ type Detected struct {
...
+	entries, err := os.ReadDir(dir)
+	if err != nil {
+		return nil
+	}
+
+	var candidates []DirCandidate
+
+	for _, entry := range entries {
+		if !entry.IsDir() || strings.HasPrefix(entry.Name(), ".") {
```
ajalon1: **Blocking.**

```go
if !entry.IsDir() || strings.HasPrefix(entry.Name(), ".") {
```

`os.DirEntry.IsDir()` reflects the entry's own on-disk type from the directory read — it does **not** resolve symlinks. A subdirectory reached through a symlink reports `IsDir() == false` (its type is `ModeSymlink`), so it's skipped here even if it carries a Dockerfile.

Confirmed locally:
```
real-project: IsDir()=true  Type()=d---------
linked-project (-> real-project): IsDir()=false Type()=L---------
```
See https://pkg.go.dev/io/fs#DirEntry — `Info()`'s doc note ("If the entry denotes a symbolic link, Info reports the information about the link itself, not the link's target") describes the same non-following behavior that applies to `Type()`/`IsDir()`.

This is a real pattern in the wild — a `current -> release-2024-01` release pointer, or a monorepo tool that symlinks a package in. The root directory itself is fine (`markersIn` uses `fsutil.FileExists` → `os.Stat`, which *does* follow symlinks) — it's specifically this one-level-down scan that's blind to it. Might be worth `entry.IsDir() || (entry.Type()&fs.ModeSymlink != 0 && fsutil.DirExists(sub))` or similar, if symlinked project dirs are meant to be supported.

internal/workload/wizard/detect.go
```diff
@@ -82,13 +82,98 @@ type Detected struct {
...
+			continue
+		}
+
+		markers := markersIn(sub)
+		if len(markers) == 0 {
+			continue
+		}
+
+		candidates = append(candidates, DirCandidate{Rel: entry.Name(), Markers: markers})
+		if len(candidates) == maxDirCandidates {
```
ajalon1, line 174: `os.ReadDir` returns entries in lexical order, so past 6 marker-bearing subdirectories, this always drops whichever sorts last — silently, in both `directoryOptions()` (interactive) and `warnSuspectDir()` (headless). Neither place indicates more candidates existed. A directory whose real project name happens to sort 7th+ never gets offered, with no hint to reach for `--dir`. Worth at least an "and N more — pass --dir" note when the cap is hit.

internal/workload/wizard/run.go
```diff
@@ -130,19 +132,25 @@ func Run(opts Options) (Result, error) {
 		warnUnreadEnvFile(opts.Stderr, detected)
```
ajalon1: This runs once, against the *original* directory's `Detected`, before the wizard/`resolve()` runs. If the directory screen later moves to a different directory, `acceptDirectory` calls `Detect(chosen)` and computes a fresh `EnvErr` for that directory's own `.env` — but nothing calls `warnUnreadEnvFile` again. A parse failure in the directory the user actually picks is silently swallowed; `EnvVars` ends up empty with no warning, which is exactly what this function's own doc comment says it exists to prevent ("it cannot pass in silence").

internal/workload/wizard/run.go
```diff
@@ -199,33 +207,67 @@ func existing(path string) (Result, error) {
...
+func (o Options) resolve(detected Detected) ([]byte, manifest.Draft, string, error) {
 	if o.NonInteractive || !isStdinTerminalFn() {
 		return o.resolveHeadless(detected)
 	}
 
 	return runInteractiveFlowFn(o, detected)
 }
 
-func (o Options) resolveHeadless(detected Detected) ([]byte, manifest.Draft, error) {
+func (o Options) resolveHeadless(detected Detected) ([]byte, manifest.Draft, string, error) {
```
ajalon1: (Simplification, non-blocking.) `resolveHeadless`/`resolveHeadlessBound` both picked up the new 4th `string` return this PR threads through five signatures, but both always return `detected.Dir` unchanged — only the interactive path (`runInteractiveFlow`) can actually produce a different directory. Three existing tests had to add a throwaway `""` to mock signatures purely to satisfy a value they don't use. `Run()` could default `projectDir := detected.Dir` up front and let only the interactive branch override it, leaving these two headless functions' signatures untouched.

internal/workload/wizard/run.go
```diff
@@ -199,33 +207,67 @@ func existing(path string) (Result, error) {
...
+	}
+
+	return content, draft, detected.Dir, nil
+}
+
+// warnSuspectDir says when the directory looks like the wrong place to run
+// setup from — none of the usual project files — and names the subdirectories
+// that look right, because the fix is one --dir away and this is the last
+// moment before six questions and an artifact are spent on the wrong tree.
+func warnSuspectDir(w io.Writer, detected Detected) {
```
ajalon1: Its two siblings in this same file both guard against a nil `Stderr` — `warnShadowedManifest` (`if stderr == nil { return }`) and `warnUnreadEnvFile` (`if stderr == nil || ...`). This one doesn't, so it panics on a nil-interface call if ever invoked with one. Not reachable through any current production caller (both `cmd/workload/config/cmd.go` and `up/run.go` always set `Stderr` via `cmd.ErrOrStderr()`), but it's an easy inconsistency to carry forward given the established pattern right next to it.

internal/workload/wizard/run.go
```diff
@@ -199,33 +207,67 @@ func existing(path string) (Result, error) {
...
+	if !detected.SuspectDir() {
+		return
+	}
+
+	line := fmt.Sprintf("Warning: %s has none of the usual project files (Dockerfile, pyproject.toml, "+
```
ajalon1, line 270: `SuspectDir()`/`warnSuspectDir()` fire purely off filesystem markers, with no awareness of `Answers.BuildMode`. For `--build-mode image`, `up/run.go`'s `defaultCodeChange` only syncs for `BuildModeDockerfile`/`BuildModeGenerated` — image-mode deploys never touch local directory contents. So for an image build from a correct-but-marker-less directory, this prints a warning describing a risk that literally cannot happen for that build mode, and interactively the new directory-confirmation screen gets inserted ahead of every other question for a check that doesn't apply. Might be worth gating the whole check on whether the requested (or eventual) build mode actually syncs local code.

ajalon1 (changes requested):
Good stuff -- the "warn and offer, never fail" framing is the right call. Found a few issues worth fixing before merge (marked **Blocking** below), plus some smaller ones and two simplification suggestions.

The big one: the safety check silently doesn't fire for exactly the case the PR description calls out as motivating it — `--workload-id` binding — because the live-workload fetch and the new screen fight over `f.at`/`f.loading` without either one knowing about the other.

One non-blocking nudge, no action needed in this PR: `internal/dependencies/registry/registry_managers.go` already has the declarative "table of {Name, predicate}" pattern this file's `rootMarkers`/`markersIn` are reaching for by hand. You might consider that for a followup as that is a rabbit hole in itself, but maybe when you have to start expanding this detection logic...

internal/workload/wizard/run.go
```diff
@@ -130,19 +132,25 @@ func Run(opts Options) (Result, error) {
...
+	content, draft, projectDir, err := opts.resolve(detected)
 	if err != nil {
 		return Result{}, err
 	}
 
+	// The interactive flow may have moved the project: everything from here
+	// on — the write, the validation's Dockerfile check, the reported path —
+	// belongs to the directory resolve settled on, which headless runs return
+	// unchanged.
+	path = manifest.Path(projectDir)
```
wojtekwdr, line 142: [high] The write follows `projectDir`, but the user's shell doesn't. `config` still prints a bare `Next: dr workload up`, and `manifest.Locate` only walks upward, so that command from the parent hits ErrNoManifest and re-runs the wizard, which no longer offers `./my-app` because `dirCandidates` skips subdirectories that already hold a manifest. The second run configures the parent and deploys the parent tree, which is the failure this PR exists to prevent, one command later.

internal/workload/wizard/model.go
wojtekwdr: [medium] question: does a malformed `.env` in the chosen directory get reported anywhere? `warnUnreadEnvFile` runs once in `Run` before `resolve`, against the parent's `Detected`, and the `EnvErr` this re-detection sets has no other reader, so a `.env` that couldn't be parsed looks the same as no `.env` at all.

internal/workload/up/run.go
```diff
@@ -488,10 +488,18 @@ func load(dir string, opts Options) (Loaded, error) {
...
+	setup, err := runWizardFn(wizard.Options{Dir: dir, Stderr: opts.Stderr})
+	if err != nil {
 		return Loaded{}, err
 	}
 
+	// The wizard's directory question may have moved the project, and the
+	// deploy has to follow the file it wrote: re-searching from where the
+	// command ran walks upward and cannot see a manifest one level down.
+	if setup.Path != "" {
+		return Load(filepath.Dir(setup.Path))
```
wojtekwdr, line 501: [low] Nothing covers this branch. The one test that reaches it has the fake write into `opts.Dir` and return `manifest.Path(opts.Dir)`, so `filepath.Dir(setup.Path)` is just `dir` again, and deleting these three lines leaves the whole `up` package green.

internal/workload/wizard/run.go
wojtekwdr: [low] `warnSuspectDir` is the only warn helper in this file without the `if w == nil { return }` guard that `warnShadowedManifest`, `warnUnreadEnvFile` and `warnLiveSecretLiterals` all carry. No production caller leaves `Stderr` nil today, so it is only about the file staying consistent with itself.

internal/workload/wizard/detect_test.go
```diff
@@ -222,3 +222,44 @@ func TestParseProblem_UnrecognizablePayloadYieldsNoLine(t *testing.T) {
...
+}
+
+// A directory carrying any project file is not suspect, and the subdirectory
+// scan never runs: a directory that looks like a project needs no offer.
+func TestDetect_MarkerBearingDirIsNotSuspect(t *testing.T) {
+	detected := Detect(writeDockerfile(t, t.TempDir(), "FROM scratch\n"))
+
+	assert.False(t, detected.SuspectDir())
+	assert.Contains(t, detected.RootMarkers, "Dockerfile")
+	assert.Empty(t, detected.Candidates)
```
wojtekwdr, line 273: [low] nit: `assert.Empty` can't fail here. The fixture has no subdirectories at all, so it passes whether or not `Detect` skips the scan. Adding a marker-bearing subdir would make it test what the comment above it claims.

wojtekwdr (dismissed):

internal/workload/wizard/model.go
```diff
@@ -968,7 +1012,7 @@ func (f *flow) onEnter(at screen) tea.Cmd {
 		return focus
 
 	case screenBinding, screenName, screenKind, screenA2A, screenSource,
-		screenEntrypoint, screenImage, screenSettings, screenEnv:
+		screenEntrypoint, screenImage, screenSettings, screenEnv, screenDirectory:
```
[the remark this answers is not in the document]
adamalpi: Fixed in 64f870b: `Init` no longer fetches while the directory question is on screen; `acceptDirectory` starts the deferred fetch once the answer is given. Covered by `TestFlow_BindWaitsForTheDirectoryAnswer`.

internal/workload/wizard/model.go
```diff
@@ -671,6 +693,28 @@ func (f *flow) acceptKind() (tea.Cmd, error) {
...
+
+	draft, err := f.answers.draft(f.detected)
+	if err != nil {
+		draft = f.answers.partialDraft(f.detected)
+	}
+
+	f.draft = draft
+
+	return nil, nil
+}
```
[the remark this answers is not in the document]
adamalpi: Fixed in 64f870b: the flow freezes the original `Detected` as `f.offer`, and `directoryOptions()` builds from it, so Escape re-poses the original candidates. Covered by `TestFlow_DirectoryOfferSurvivesEscape`.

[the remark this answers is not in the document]
adamalpi: Fixed in 64f870b: choosing another directory drops the env table along with the draft. Covered by `TestFlow_ChangingDirectoryDropsTheOldEnvTable`.

internal/workload/wizard/detect.go
adamalpi: Added `log.Debug` on the three quiet exits worth tracing: the home-directory skip, an unlistable directory, and the suspect-dir summary (candidate count + truncation). Kept at debug — the user-facing story is the wizard question / headless warning themselves.

internal/workload/wizard/detect.go
adamalpi: Agreed there's overlap, but the two answer different questions today (which package manager vs. does this look like a project root), so I'd rather converge them deliberately than grow this PR — filed as https://datarobot.atlassian.net/browse/RAPTOR-19752.

internal/workload/wizard/model.go
adamalpi: Fixed in 64f870b: `newFlow` skips the loading view and `Init` skips the fetch while the directory question is first; `acceptDirectory` resumes the bind afterwards. Test: `TestFlow_BindWaitsForTheDirectoryAnswer`.

internal/workload/wizard/model.go
adamalpi: Fixed in 64f870b with a frozen `f.offer` that the question always renders from. Test: `TestFlow_DirectoryOfferSurvivesEscape`.

internal/workload/wizard/model.go
adamalpi: Deduped into `Answers.draftOrPartial`, used by both `newFlow` and `acceptDirectory`.

internal/workload/wizard/detect.go
adamalpi: Fixed in 64f870b: `maxDirEntriesScanned = 256` caps the work, `$HOME` is never scanned, and stopping early at either cap is reported (`MoreCandidates`) so the offer says the list is incomplete.

internal/workload/wizard/detect.go
adamalpi: Fixed in 64f870b: `isSubdirectory` follows one symlink level via `os.Stat`. Test: `TestDetect_CandidateBehindSymlinkIsOffered`.

internal/workload/wizard/detect.go
adamalpi: Fixed in 64f870b: a truncated scan sets `MoreCandidates`; the interactive offer's stay-put option says "more candidates exist — pass --dir to name one" and the headless warning appends ", and more exist". Test: `TestDetect_TruncatedOfferReportsMoreCandidates`.

internal/workload/wizard/run.go
adamalpi: Fixed in 64f870b: after `resolve`, a chosen directory that differs from the starting one gets its own `.env` warning pass (same `--skip-env` opt-out).

internal/workload/wizard/run.go
adamalpi: Reverted — the headless resolvers are back to three returns; only `resolve` reports the settled directory.

internal/workload/wizard/run.go
adamalpi: Fixed — it's now an `Options` method with the same nil guard as its siblings.

internal/workload/wizard/run.go
adamalpi: Gated in 64f870b for the headless path: `--build-mode image` no longer warns (test: `TestRun_HeadlessImageModeSkipsTheSuspectDirWarning`). Interactively the build mode isn't known yet when the directory question is asked — reordering the screens for that feels like its own change, so the interactive offer stays unconditional. Happy to file a follow-up if you think it's worth it.

internal/workload/wizard/run.go
adamalpi: Fixed in 64f870b: `config` now prints `Next: dr workload up --dir <chosen>` via the existing `manifest.DirFlag`, so it stays bare exactly when the shell already stands in the project.

internal/workload/wizard/model.go
adamalpi: It didn't — fixed in 64f870b: `Run` re-runs the `.env` warning for the settled directory when it differs from the starting one.

internal/workload/up/run.go
adamalpi: Added `TestRun_WizardRedirectIsFollowed`: the fake wizard writes into a subdirectory and the test asserts the workload id lands in that manifest.

internal/workload/wizard/run.go
adamalpi: Same fix as the sibling thread — `warnSuspectDir` is now an `Options` method with the nil guard.

internal/workload/wizard/detect_test.go
adamalpi: Fixed — the fixture now plants a marker-bearing subdirectory, so the assertion proves the scan was skipped rather than that there was nothing to find.

internal/workload/wizard/detect.go
ajalon1: Happy to see this one fixed; I spent a lot of brain cycles and agent turns on symlinks for workload config 😓

ajalon1 (approved):
This LGTM earlier, anything else can be handled in follow ups

PR #853, opened 2026-08-27
[RAPTOR-19530] feat(workload): wait out a rollout in flight instead of refusing

cdevent:
# RATIONALE

[RAPTOR-19530](https://datarobot.atlassian.net/browse/RAPTOR-19530) — *"if a workload is still settling when I run `dr wl up` it should wait around and tell me it is doing so rather than erroring out."*

Most of that shipped in [RAPTOR-19686](https://datarobot.atlassian.net/browse/RAPTOR-19686) (#836): `awaitSteady` waits out a workload whose *status* is transitional (`submitted`, `provisioning`, `launching`, `stopping`, `unknown`) before the plan is built.

What was left is the one transition that check **structurally cannot see**, and it is the most common one. A workload being replaced reports itself `running` for the whole of the swap, so a rollout in flight never reached `awaitSteady`; it reached `guardRollout` and was refused:

```
Error: workload 68b0…e9f0: a replacement is already in progress
(to artifact 68a0…00a2, status switching); wait for it to settle before starting another
```

That is literally the erroring-out-and-telling-you-to-wait the ticket asks us to stop doing.

## CHANGES

**`lookSettled` asks the replacement route before it builds the plan** (`internal/workload/up/run.go`). New `awaitReplaced` names the swap on stderr, waits it out, and re-reads the workload:

```
  Workload 68b0c1d2e3f4a5b6c7d8e9f0 is being replaced onto artifact 68a0…00a2, status switching.
  ✓ Waiting for the rollout already in progress  2m14s

  Plan for workload gpt-oss-20b-vllm
  Already up to date.
```

Ordering: **rollout first, status second.** A swap that lands leaves the workload coming up, so one pass settles both; the other order would return with the workload still `provisioning` and hand `deployable` a state it has to refuse.

The re-read is the point, not a formality — a completed swap moves the workload onto a different artifact, so a plan built against the pre-swap state would roll a version the platform had already installed, or report drift the swap had just closed.

**A rollout that ends `failed` is noted, and the run continues.** A failed rollout never promotes, so the version that was serving is still serving and the workload is deployable. Refusing there would hand back the same "run it again" dead end this wait exists to remove:

```
  ⚠ That rollout ended as failed, so the workload is still running the version it was.
```

A timeout or an unreadable route stops the run and names where the rollout got to (`"the rollout of workload … was still switching, so nothing was deployed; check 'dr workload status …'"`), because "still switching after 30m" is what decides whether to wait longer or go and look at the platform.

**The guards at the apply sites stay refusals.** A rollout that appears *after* the pre-plan read is somebody deploying concurrently, and waiting there would apply a plan built against state that has since moved — for `--lock`, it would take a one-way lock on the outgoing version. All five `guardRollout` sites are unchanged.

**Flag behaviour mirrors the settling wait**: `--dry-run` says a deploy would wait and changes nothing; `--detach` explains that this wait precedes the deploy.

**Not asked at all** when there is no workload for the route to answer about (unbound, missing, terminated) — the replacement route returns the same 404 for "no such workload" as for "nothing in flight", so the answer could not be read either way.

Help text for `dr workload up` updated to cover the new case.

## TESTING

`task lint` clean across all three GOOS targets; `task test` green (141 packages).

Eight new tests in `internal/workload/up/run_test.go`, all verified red before the change:

| Test | Asserts |
|---|---|
| `RolloutInFlightIsWaitedOutAndThenDeployed` | waits, seeds the wait with the record just read, re-reads — fixture carries drift only on the first read, proving the plan comes from the post-swap state |
| `RolloutThatEndsFailedIsNotedAndTheRunContinues` | no error, note on stderr, the run goes on to do what the file asked |
| `RolloutThatNeverLandsStopsBeforeMutating` | error names the status it was still in, nothing mutated, binding survives |
| `UnreadableRolloutStateStopsTheRun` | `"cannot tell whether"` — a read that failed is not a rollout |
| `DryRunWithARolloutInFlightDoesNotWait` | no wait, says a deploy would |
| `DetachedRunSaysWhyItIsWaitingForTheRollout` | `--detach applies to the deploy` |
| `ARunWithNoLiveWorkloadNeverAsksAboutARollout` | the route is never called on a create |
| `LockWaitsOutARolloutThatWasAlreadyInFlight` | the one-way lock lands on what the swap left serving, not on the pre-swap read |

`TestRun_LockWithNothingToDoWaitsForARolloutInFlight` renamed to `…RefusesARolloutThatStartedLate` — its behaviour is unchanged, but it now documents the late guard as the backstop rather than the ordinary answer.

### Verified live on staging

Same workload, back to back, on `https://staging.datarobot.com`. `--detach` returns the moment the swap is *requested*, which leaves a rollout in flight deterministically — no two-shell race needed.

**Before** — `main` binary, deploying while a rollout was in flight:

```
Error: workload 6a9047702caffb5d64da7c9e: a replacement is already in progress
(to artifact 6a9047d03b68e0a45994233f, status initializing); wait for it to settle before starting another
```

**After** — this branch, same situation:

```
  Workload 6a9047702caffb5d64da7c9e is being replaced onto artifact 6a9047d03b68e0a45994233f, status promoting.
  ✓ Waiting for the rollout already in progress (5.3s)
raptor-19530-probe (6a904770), running

  + artifact   new version, 1 spec change
      containerGroups[default].containers[primary].environmentVars[MARKER].value: changed
  ✓ Creating the new version (200ms)
  ✓ Building the image (19s)
  ✓ Rolling out the new version (200ms)
  ✓ Waiting for the rollout (26.1s)
  ✓ Waiting for the workload to run (600ms)
  ✓ Workload endpoint: https://staging.datarobot.com/api/v2/endpoints/workloads/6a9047702caffb5d64da7c9e/
```

**`--dry-run`** during a rollout returned immediately (checked under a 60s timeout), said what it saw and changed nothing:

```
  Workload 6a9047702caffb5d64da7c9e is being replaced onto artifact 6a9048572caffb5d64da7ce2, status initializing.
  A deploy would wait for this rollout to finish and plan against where it lands.
  ...
Dry run: nothing was changed.
```

Probe workload and all four draft artifacts deleted afterwards; `dr workload list` and `dr artifact list` are clean.

## NOTES

**The live run turned up a fact worth recording.** The two in-flight statuses the platform actually returned were **`initializing`** and **`promoting`**. Neither appears in the documented replacement vocabulary — `internal/workload/replacement.go` names only `completed`, `failed` and `errored`, with a comment noting the non-terminal vocabulary "is not documented anywhere". Both read correctly as in-flight only because `IsTerminalReplacementStatus` treats an unrecognised status as still in progress. That stance is now load-bearing for this feature, and the two observed names are evidence for the open question already flagged in that file's comment.

Cost: one extra `GET /workloads/{id}/replacement/` per deploy onto an existing workload. Against a deploy measured in minutes, it replaces a class of failure that costs a whole re-run.

---

> [!NOTE]
> **Medium Risk**
> Changes the pre-plan deploy path and artifact/lock timing on an existing workload, though apply guards and broad test coverage limit blast radius.
> 
> **Overview**
> **`dr workload up` no longer errors when another rollout is already swapping the workload.** Before building the plan, `lookSettled` now calls **`awaitReplaced`**, which checks the replacement API (workloads stay `running` during a swap, so status-only waits cannot see it), prints what is in flight, waits for completion when appropriate, and **re-reads** live state so planning targets the post-swap artifact.
> 
> Rollout wait runs **before** the existing settling wait. **`--dry-run`** and **`--detach`** behave like the settling path: preview does not block; detach still waits pre-plan but explains why. A rollout that ends **failed** is warned and the run continues; timeouts and unreadable replacement state still abort without mutating. **Apply-time `guardRollout` refusals are unchanged** for concurrent swaps that start after the pre-plan read (including the late `--lock` race). Help text documents the new behavior; tests cover wait, re-read, lock ordering, and flag behavior.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 43650a6b4970940791c870a2ae2a2f3b85029949. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/internal/workload/up/run.go
+++ b/internal/workload/up/run.go
@@ -384,10 +384,25 @@ func awaitReplaced(live Live, opts Options) (Live, error) {
 			live.WorkloadID, err)
 	}
 
-	// A settled record stays readable for a while after the rollout ends, and
-	// waiting on one would block every deploy for as long as it lingers.
-	if active == nil || workload.IsTerminalReplacementStatus(active.Status) {
+	if active == nil {
 		return live, nil
+	}
+
+	// A settled record stays readable for a while after the rollout ends, so
+	// waiting on one would block every deploy for as long as it lingers. It is
+	// still re-read rather than returned as it stands: a terminal record means a
+	// swap landed recently, possibly in the window between the read above and
+	// this one, and the snapshot from before it names the outgoing artifact.
+	// Planning against that would roll a version the platform had just
+	// installed, and a --lock run would make the version being rolled off
+	// permanent, which cannot be undone.
+	//
+	// The residual, stated rather than pretended away: a swap whose record is
+	// collected inside that same window reads as nil above and is not re-read.
+	// Closing it means re-reading on every deploy, which doubles the workload
+	// GET on the quiet path to catch a window narrower than the one this covers.
+	if workload.IsTerminalReplacementStatus(active.Status) {
+		return Look(live.WorkloadID)
 	}
 
 	report := newReporter(opts.Stderr, opts.Spinner)
@@ -397,6 +412,19 @@ func awaitReplaced(live Live, opts Options) (Live, error) {
 	if opts.DryRun {
 		report.say("  %s\n", tui.HintStyle.Render(
 			"A deploy would wait for this rollout to finish and plan against where it lands."))
+
+		// The plan below is computed against a workload the platform is already
+		// moving, so an empty one is not "up to date": the swap decides what
+		// differs, and it has not landed. StateSettling is what says so, and it
+		// is the same answer for the same reason a workload halfway through
+		// stopping gets. Without it the preview prints "Already up to date"
+		// directly beneath the note above, contradicting it.
+		//
+		// Safe to synthesise because it is confined to the preview. Every state
+		// that decides the shape of a plan already groups settling with running
+		// (see actsOnState and creates), so only the verdict moves; and a real
+		// run never arrives here with it, because it waits and re-reads instead.
+		live.State = StateSettling
 
 		return live, nil
 	}
--- a/internal/workload/up/run.go
+++ b/internal/workload/up/run.go
@@ -54,6 +54,7 @@ var (
 	listBuildsFn         = workload.ListArtifactBuilds
 	getCredentialFn      = workload.GetCredential
 	findCredentialFn     = workload.FindCredentialNamed
+	activeReplacementFn  = workload.GetActiveReplacement
 	guardReplacementFn   = workload.RefuseActiveReplacement
 	startReplacementFn   = workload.StartReplacement
 	waitReplacementFn    = workload.WaitForReplacement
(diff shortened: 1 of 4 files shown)


internal/workload/up/run.go
```diff
@@ -282,13 +283,215 @@ func guardRollout(workloadID, consequence string) error {
...
+		// differs, and it has not landed. StateSettling is what says so, and it
+		// is the same answer for the same reason a workload halfway through
+		// stopping gets. Without it the preview prints "Already up to date"
+		// directly beneath the note above, contradicting it.
+		//
+		// Safe to synthesise because it is confined to the preview. Every state
+		// that decides the shape of a plan already groups settling with running
+		// (see actsOnState and creates), so only the verdict moves; and a real
+		// run never arrives here with it, because it waits and re-reads instead.
+		live.State = StateSettling
```
ajalon1, line 456: **Dry-run with a rollout in flight prints the same hint twice.**

When `--dry-run` hits an in-flight rollout, `awaitReplaced` (line 369 above) prints:
```
A deploy would wait for this rollout to finish and plan against where it lands.
```
...then sets `live.State = StateSettling` and returns. Back in `lookSettled`, `awaitSteady(replaced, opts)` runs next, sees `live.State == StateSettling`, and — since `opts.DryRun` is also true — prints its *own* near-identical hint right after:
```
A deploy would wait for this to finish and plan against where it lands.
```
Confirmed live by running the `TestRun_DryRunWithARolloutInFlightDoesNotWait` scenario and dumping stderr — both lines appear back to back:
```
  Workload 68b0c1d2e3f4a5b6c7d8e9f0 is being replaced, status switching.
  A deploy would wait for this rollout to finish and plan against where it lands.
  A deploy would wait for this to finish and plan against where it lands.
```
None of the new tests catch this because they only `assert.Contains`, never checking the message isn't duplicated. Worth either short-circuiting `lookSettled` after the dry-run synthetic-state assignment, or having `awaitSteady` skip its own hint when the settling state came from `awaitReplaced` rather than a real read.

internal/workload/up/run.go
```diff
@@ -282,13 +283,215 @@ func guardRollout(workloadID, consequence string) error {
...
+	}
+
+	active, err := activeReplacementFn(live.WorkloadID)
+	if err != nil {
+		return live, fmt.Errorf(
+			"cannot tell whether workload %s already has a rollout in progress, so nothing was deployed: %w",
+			live.WorkloadID, err)
+	}
+
+	if active == nil {
```
ajalon1, line 399: **Nit: nothing marks the quiet path (no rollout in flight) in the debug log.**

`version.go` in this same package logs decisions the user wouldn't see otherwise (e.g. `log.Debug("artifact repository refused; creating the version in a new one", ...)` at line 220) so a `--debug` run can explain *why* something took the path it did. `awaitReplaced` never touches `internal/log` at all — this branch is the ordinary, silent case (nothing in flight, proceed straight to `awaitSteady`), so probably fine as-is, but worth a `log.Debug("no replacement in flight", "workload_id", live.WorkloadID)` if a support engineer ever needs to confirm from a `--debug` transcript that this route was even checked.

internal/workload/up/run.go
```diff
@@ -282,13 +283,215 @@ func guardRollout(workloadID, consequence string) error {
...
+	// this one, and the snapshot from before it names the outgoing artifact.
+	// Planning against that would roll a version the platform had just
+	// installed, and a --lock run would make the version being rolled off
+	// permanent, which cannot be undone.
+	//
+	// The residual, stated rather than pretended away: a swap whose record is
+	// collected inside that same window reads as nil above and is not re-read.
+	// Closing it means re-reading on every deploy, which doubles the workload
+	// GET on the quiet path to catch a window narrower than the one this covers.
+	if workload.IsTerminalReplacementStatus(active.Status) {
```
ajalon1, line 424: **This branch is the one truly silent decision in the new code — worth a debug log.**

When the replacement record read here is already terminal, the function discards `live` and does a second, unlogged `Look()` instead of waiting — the exact narrow race the PR description calls out (a swap landing between the workload read and the replacement read). Nothing is said to the user (no `report.say`) and nothing goes to `internal/log` either, so this decision is invisible even under `--debug`. Compare to `version.go`'s `fellBack` handling, which is deliberately "said out loud rather than only to the debug log" for the same reason: "nobody goes looking for the cause of that after a run that printed nothing but check marks." Something like:

```go
log.Debug("replacement already settled; re-reading before planning",
    "workload_id", live.WorkloadID, "status", active.Status)
```

would make this traceable if a plan ever looks like it rolled the wrong artifact.

ajalon1:
LGTM, would suggest adding the debug logs, defer to @wojtekwdr and @adamalpi for final approval

chasdr:
took a look through this myself, nothing new to add. the earlier comments (both bugbot ones and aj's three) are all addressed in the follow-up commits and each has a test. gtg from me.

seconding aj on deferring final approval. @datarobot-oss/workload-cli this is ready for your review.

internal/workload/up/run.go
```diff
@@ -327,13 +329,245 @@ func guardRollout(workloadID, consequence string) error {
...
+
+	// A preview that has already said a deploy would wait for the rollout says
+	// nothing further. The settling state it hands back is synthetic — put there
+	// so an empty plan is not called up to date — and awaitSteady would read it
+	// as a second transition and print the same sentence about the same wait.
+	if previewed {
+		return replaced, nil
+	}
+
+	return awaitSteady(replaced, opts)
```
ajalon1, line 359: **The two pre-plan waits don't share the `--poll-timeout` budget.**

`awaitReplaced` can spend the whole `opts.PollTimeout` waiting out the rollout, and then this call hands `awaitSteady` a fresh full copy — `dr workload up --poll-timeout 10m` can block for up to ~20m before a plan is built. That is the exact case `budgetLeft` (run.go:1593) was written for — *"A --poll-timeout of ten minutes meant ten, not twenty"* — and the sibling roll path already budgets its second wait with it (roll.go:346: *"a --poll-timeout the user set is a bound on the deploy, not on each half"*). Realistic trigger: a long rollout that eats most of the budget and lands with the workload still provisioning, which is exactly the state this ordering is there to catch.

```suggestion
	waitFrom := time.Now()

	replaced, previewed, err := awaitReplaced(found, opts)
	if err != nil {
		return replaced, err
	}

	// A preview that has already said a deploy would wait for the rollout says
	// nothing further. The settling state it hands back is synthetic — put there
	// so an empty plan is not called up to date — and awaitSteady would read it
	// as a second transition and print the same sentence about the same wait.
	if previewed {
		return replaced, nil
	}

	return awaitSteady(replaced, budgetLeft(opts, waitFrom))
```

internal/workload/up/run_test.go
```diff
@@ -1187,6 +1200,421 @@ func TestRun_SettlingIntoErroredIsNotReportedAsUpToDate(t *testing.T) {
...
+		) (*workload.Replacement, error) {
+			return &workload.Replacement{ID: "rep-1", Status: workload.ReplacementStatusCompleted}, nil
+		},
+	})
+
+	bound := "workloadId: 68b0c1d2e3f4a5b6c7d8e9f0\n" + boundLiveManifest
+
+	_, stderr, err := runIn(t, bound, Options{NonInteractive: true, Detach: true})
+	require.NoError(t, err)
+	assert.Contains(t, stderr, "--detach applies to the deploy")
```
ajalon1, line 1425: **This test passes even if `--detach` stops waiting.**

The `waitReplace` fake records no side effect and the only assertion is the stderr message, so a regression that prints the detach hint but returns without calling `waitReplacementFn` still passes: the fixture matches the live state field for field, the plan comes up empty, and `require.NoError` holds either way. The sibling `TestRun_RolloutInFlightIsWaitedOutAndThenDeployed` captures `waitedFor` for exactly this reason — worth doing the same here so the contract "--detach still waits before planning" is actually covered.

```suggestion
	var waitedFor string

	install(t, fakes{
		workloadD: func(string) (workload.Document, error) { return doc(t, liveWorkloadJSON), nil },
		artifactD: func(string) (workload.Document, error) { return doc(t, liveArtifactJSON), nil },
		activeReplacement: func(string) (*workload.Replacement, error) {
			return &workload.Replacement{ID: "rep-1", Status: "switching"}, nil
		},
		waitReplace: func(id string, _ *workload.Replacement, _, _ time.Duration,
			_ func(*workload.Replacement),
		) (*workload.Replacement, error) {
			waitedFor = id

			return &workload.Replacement{ID: "rep-1", Status: workload.ReplacementStatusCompleted}, nil
		},
	})

	bound := "workloadId: 68b0c1d2e3f4a5b6c7d8e9f0\n" + boundLiveManifest

	_, stderr, err := runIn(t, bound, Options{NonInteractive: true, Detach: true})
	require.NoError(t, err)
	assert.Contains(t, stderr, "--detach applies to the deploy")
	assert.Equal(t, "68b0c1d2e3f4a5b6c7d8e9f0", waitedFor,
		"--detach still waits before planning; the message alone does not prove it")
```

ajalon1 (commented):
Two follow-ups from re-reviewing the head after the follow-up commits (70c110d6). Everything from the earlier round — the three review comments and both bugbot findings — checks out as fixed in the code, each with a test. These two are new against the current head.

internal/workload/up/run.go
```diff
@@ -327,13 +329,245 @@ func guardRollout(workloadID, consequence string) error {
...
+		// differs, and it has not landed. StateSettling is what says so, and it
+		// is the same answer for the same reason a workload halfway through
+		// stopping gets. Without it the preview prints "Already up to date"
+		// directly beneath the note above, contradicting it.
+		//
+		// Safe to synthesise because it is confined to the preview. Every state
+		// that decides the shape of a plan already groups settling with running
+		// (see actsOnState and creates), so only the verdict moves; and a real
+		// run never arrives here with it, because it waits and re-reads instead.
+		live.State = StateSettling
```
wojtekwdr, line 456: [medium] This also catches a stopped or errored workload, and there it hides the state the plan exists to report. With an errored workload plus a rollout in flight, `--dry-run` now exits 0 and lists the artifact change with no refusal, where it used to return "workload X is errored, so there is nothing safe to deploy onto"; a stopped one loses its "started, having been stopped" line and the envelope's action goes from started to unchanged. The comment above cites actsOnState, but plan.go:146 puts stopped and errored on the true side and settling on the false side, so a `live.State == StateRunning` guard would keep the synthesis and its rationale in agreement.

internal/workload/up/run.go
```diff
@@ -327,13 +329,245 @@ func guardRollout(workloadID, consequence string) error {
...
+
+		return waitErr
+	})
+	if err != nil {
+		if failure := replacedFailed(live, settled, err); failure != nil {
+			return live, false, failure
+		}
+
+		report.say("  %s\n", tui.WarnStyle.Render(fmt.Sprintf(
+			"⚠ That rollout ended as %s, so the workload is still running the version it was.",
```
wojtekwdr, line 487: [low] For a stopped workload this line says it's still running. TestRun_RolloutThatEndsFailedIsNotedAndTheRunContinues builds exactly that case, so the transcript reads "still running the version it was" and then starts it a moment later. "still on the version it was" would be true either way.

PR #854, opened 2026-08-27
[RAPTOR-19519] fix(workload): stop rebuilding the image for runtime-only changes

wojtekwdr:
# RATIONALE

Changing an environment variable on a workload retriggered a full image build. Env vars are runtime configuration: the container reads them when it starts, so the change wants a restart, not a new image.

The rebuild was ours rather than the server's. Any artifact spec edit made the run a roll, a roll minted a fresh artifact, and a fresh artifact has no image, so it always built. A port, a probe or a route cost a build the same way.

## CHANGES

A version that changes only fields a container reads at start is now copied from the one now serving, inheriting its image, the code it was built into and its repository, with the file's spec written over the copy. Everything else still builds. An env-var deploy goes from about 45 seconds to about 25, with no build at all, and the same holds for locked production and for a workload that was stopped.

Three calls worth arguing about.

The runtime fields are an allow-list rather than a list of build inputs to exclude. The compiler passes keys it does not recognize straight through, so a field the platform grows after this release ships would read as runtime under a deny-list, inherit a stale image and deploy clean. Under an allow-list it reads as a rebuild, which costs a build and is always survivable.

Project state gains `lastBuiltVersionId`. Nothing on the platform records which code an image was built from: a build row does not name its catalog pair, and an artifact's build pointer is cleared both by a re-sync and by being copied, so the two cannot be told apart. Without it, `dr artifact code sync` between deploys would let a later env-var change inherit an image built from the code before that sync and report success. Absent, which it is for every existing project until its next build, the deploy behaves as it did before.

A copy inherits artifact-spec fields the file never mentions instead of dropping them as a create would. That is the one-directional contract `Subset` already implements, but it is now a difference between the two tracks.

Verified against staging across twenty scenarios, including locked production, a stopped workload, an out-of-band `dr artifact code sync`, `--detach`, `--force-build`, a service-to-agent type change and a manifest bound by artifact id.

---

> [!NOTE]
> **Medium Risk**
> Changes the core roll/deploy path (artifact minting, build skipping, and production/locked behavior) with new platform API dependencies; mitigated by conservative runtime allow-list, read-back validation, and broad integration tests.
> 
> **Overview**
> **`dr workload up` no longer triggers a full image build when a roll only changes runtime configuration** (environment variables, ports, probes, routes, and similar fields read at container start). Those deploys **clone the running artifact**, **PATCH the manifest spec onto the copy** (preserving image and code reference), and **roll without calling the builder**.
> 
> The deploy planner gains **`RebuildsImage` / `InheritsImage`** using an **allow-list of runtime fields** and structured **`Change.Keys`** so rebuild vs copy is not inferred from rendered paths. **`Look`** now surfaces the live **image URI** and **catalog version**; project **`lastBuiltVersionId`** records which code the last successful build used so env-only rolls still rebuild after **`dr artifact code sync`**. New workload API helpers **`CloneArtifact`**, **`UpdateArtifactSpec`**, and **`SpecWithPrimaryCodeRef`** support the copy path, with **fallback to create-and-build** when clone is unsupported or the copy cannot be updated. Plans and JSON output expose **`keepsImage`** / dry-run messaging when no rebuild is expected.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 7dc420d513dccd67dd27908b06cf107d1c772feb. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/internal/workload/artifact.go
+++ b/internal/workload/artifact.go
@@ -373,6 +373,45 @@ func CreateArtifact(payload any) (*Artifact, error) {
 	return &artifact, nil
 }
 
+// CloneArtifact copies an artifact into a new draft named name.
+func CloneArtifact(artifactID, name string) (*Artifact, error) {
+	url, err := config.GetEndpointURL("/api/v2/artifacts/" + escapeID(artifactID) + "/clone")
+	if err != nil {
+		return nil, err
+	}
+
+	body := map[string]string{"name": name}
+
+	var artifact Artifact
+
+	if err := drapi.PostJSON(url, "artifact clone", body, &artifact); err != nil {
+		return nil, err
+	}
+
+	return &artifact, nil
+}
+
+// UpdateArtifactSpec writes a draft artifact's spec.
+//
+// containerGroups is replaced wholesale rather than merged key by key. The
+// platform re-injects what it owns on a build-from-source container, so a spec
+// that omits imageUri keeps the running image; codeRef is not re-injected,
+// which is why the copy path puts it back before writing.
+func UpdateArtifactSpec(artifactID string, spec json.RawMessage) error {
+	if len(spec) == 0 {
+		return errors.New("no artifact spec to update")
+	}
+
+	url, err := config.GetEndpointURL("/api/v2/artifacts/" + escapeID(artifactID) + "/")
+	if err != nil {
+		return err
+	}
+
+	body := map[string]json.RawMessage{"spec": spec}
+
+	return drapi.PatchJSON(url, "artifact", body, nil)
+}
+
 func PatchArtifactCodeRef(artifactID, catalogID, catalogVersionID string) error {
 	url, err := config.GetEndpointURL("/api/v2/artifacts/" + escapeID(artifactID) + "/")
 	if err != nil {
@@ -394,6 +433,29 @@ func PatchArtifactCodeRef(artifactID, catalogID, catalogVersionID string) error 
 	return drapi.PatchJSON(url, "artifact", body, nil)
 }
 
+// SpecWithPrimaryCodeRef returns spec with the primary container pointed at the
+// catalog pair given. An update replaces a container wholesale, so the file's
+// spec alone would erase the reference.
+func SpecWithPrimaryCodeRef(spec json.RawMessage, catalogID, catalogVersionID string) (json.RawMessage, error) {
+	var decoded map[string]any
+
+	if err := json.Unmarshal(spec, &decoded); err != nil {
+		return nil, fmt.Errorf("cannot read the artifact spec: %w", err)
+	}
+
+	// The map is shared rather than copied, so its edits are marshalled back.
+	if err := setPrimaryCodeRefInRawArtifact(map[string]any{"spec": decoded}, catalogID, catalogVersionID); err != nil {
+		return nil, err
+	}
+
+	updated, err := json.Marshal(decoded)
+	if err != nil {
+		return nil, fmt.Errorf("cannot convert the artifact spec to JSON: %w", err)
+	}
+
+	return updated, nil
+}
+
 func setPrimaryCodeRefInRawArtifact(raw map[string]any, catalogID, catalogVersionID string) error {
 	spec, ok := raw["spec"].(map[string]any)
 	if !ok {
@@ -405,76 +467,140 @@ func setPrimaryCodeRefInRawArtifact(raw map[string]any, catalogID, catalogVersio
 		return errors.New("artifact: spec.containerGroups missing or empty")
 	}
 
-	codeRef := map[string]any{
-		"datarobot": map[string]any{
-			"catalogId":        catalogID,
-			"catalogVersionId": catalogVersionID,
-		},
+	container := primaryContainerInGroups(groups)
+	if container == nil {
+		return noPrimaryContainer(groups)
 	}
 
-	if found := assignToPrimaryContainer(groups, codeRef); found {
+	setImageBuildConfigCodeRef(container, map[string]any{
+		keyRawDatarobot: map[string]any{
+			keyRawCatalogID:        catalogID,
+			keyRawCatalogVersionID: catalogVersionID,
+		},
+	})
+
+	return nil
+}
+
+// noPrimaryContainer says why the search came back empty. Asked only after it
+// has failed: checking groups[0] first would refuse an artifact whose flagged
+// primary sits in a later group.
+func noPrimaryContainer(groups []any) error {
+	group, ok := groups[0].(map[string]any)
+	if !ok {
+		return errors.New("artifact: spec.containerGroups[0] missing or wrong type")
+	}
+
+	if len(containersOfGroup(group)) == 0 {
+		return errors.New("artifact: spec.containerGroups[0].containers missing or empty")
+	}
+
+	return errors.New("artifact: spec.containerGroups[0].containers[0] missing or wrong type")
+}
+
+// Keys along the code reference inside a decoded container, spelled once so
+// the reader and the writer below cannot drift apart.
+const (
+	keyRawImageBuildConfig = "imageBuildConfig"
+	keyRawCodeRef          = "codeRef"
+	keyRawDatarobot        = "datarobot"
+	keyRawCatalogID        = "catalogId"
+	keyRawCatalogVersionID = "catalogVersionId"
+)
+
+// CodeRefInContainer reads the catalog pair a decoded container points at, nil
+// when it names none. The read half of setPrimaryCodeRefInRawArtifact.
+func CodeRefInContainer(container map[string]any) *DatarobotCodeRef {
+	build, _ := container[keyRawImageBuildConfig].(map[string]any)
+	ref, _ := build[keyRawCodeRef].(map[string]any)
+
+	datarobot, ok := ref[keyRawDatarobot].(map[string]any)
+	if !ok {
 		return nil
 	}
 
-	// Mirror ExtractCodeRef's [0][0] fallback when no container is flagged primary.
-	return assignToFirstContainer(groups, codeRef)
+	catalogID, _ := datarobot[keyRawCatalogID].(string)
+	catalogVersionID, _ := datarobot[keyRawCatalogVersionID].(string)
+
+	return &DatarobotCodeRef{CatalogID: catalogID, CatalogVersionID: catalogVersionID}
 }
 
-func assignToPrimaryContainer(groups []any, codeRef map[string]any) bool {
+// PrimaryContainerInDocument is primaryContainer for an artifact still in the
+// shape the server sent it. Two readings of which container is primary
+// disagreeing is how a plan promises an image the run cannot find.
+func PrimaryContainerInDocument(raw map[string]any) map[string]any {
+	spec, ok := raw["spec"].(map[string]any)
+	if !ok {
+		return nil
+	}
+
+	groups, ok := spec["containerGroups"].([]any)
+	if !ok {
+		return nil
+	}
+
+	return primaryContainerInGroups(groups)
+}
+
+// primaryContainerInGroups is primaryContainer's rule for a decoded document:
+// the container flagged "primary", else the first.
+func primaryContainerInGroups(groups []any) map[string]any {
+	if flagged := flaggedPrimaryContainer(groups); flagged != nil {
+		return flagged
+	}
+
+	return firstContainerInGroups(groups)
+}
+
+// flaggedPrimaryContainer is the container that claims to be primary, if any.
+func flaggedPrimaryContainer(groups []any) map[string]any {
 	for _, g := range groups {
 		group, ok := g.(map[string]any)
 		if !ok {
 			continue
 		}
 
-		containers, ok := group["containers"].([]any)
-		if !ok {
-			continue
-		}
-
-		for _, c := range containers {
-			container, ok := c.(map[string]any)
-			if !ok {
-				continue
-			}
-
-			if isPrimaryContainer(container) {
-				setImageBuildConfigCodeRef(container, codeRef)
-
-				return true
+		for _, c := range containersOfGroup(group) {
+			if container, ok := c.(map[string]any); ok && isPrimaryContainer(container) {
+				return container
 			}
 		}
 	}
-
-	return false
-}
-
-func assignToFirstContainer(groups []any, codeRef map[string]any) error {
-	firstGroup, ok := groups[0].(map[string]any)
-	if !ok {
-		return errors.New("artifact: spec.containerGroups[0] missing or wrong type")
-	}
-
-	containers, ok := firstGroup["containers"].([]any)
-	if !ok || len(containers) == 0 {
-		return errors.New("artifact: spec.containerGroups[0].containers missing or empty")
-	}
-
-	firstContainer, ok := containers[0].(map[string]any)
-	if !ok {
-		return errors.New("artifact: spec.containerGroups[0].containers[0] missing or wrong type")
-	}
-
-	setImageBuildConfigCodeRef(firstContainer, codeRef)
 
 	return nil
+}
+
+func firstContainerInGroups(groups []any) map[string]any {
+	if len(groups) == 0 {
+		return nil
+	}
+
+	group, ok := groups[0].(map[string]any)
+	if !ok {
+		return nil
+	}
+
+	containers := containersOfGroup(group)
+	if len(containers) == 0 {
+		return nil
+	}
+
+	first, _ := containers[0].(map[string]any)
+
+	return first
+}
+
+func containersOfGroup(group map[string]any) []any {
+	containers, _ := group["containers"].([]any)
+
+	return containers
 }
 
 // setImageBuildConfigCodeRef preserves any existing dockerfile config and
 // seeds a "provided" Dockerfile default when imageBuildConfig is absent
 // (server requires a dockerfile on the imageBuildConfig).
 func setImageBuildConfigCodeRef(container map[string]any, codeRef map[string]any) {
-	ibc, ok := container["imageBuildConfig"].(map[string]any)
+	ibc, ok := container[keyRawImageBuildConfig].(map[string]any)
 	if !ok || ibc == nil {
 		ibc = map[string]any{
 			"dockerfile": map[string]any{
@@ -483,8 +609,8 @@ func setImageBuildConfigCodeRef(container map[string]any, codeRef map[string]any
 		}
 	}
 
-	ibc["codeRef"] = codeRef
-	container["imageBuildConfig"] = ibc
+	ibc[keyRawCodeRef] = codeRef
+	container[keyRawImageBuildConfig] = ibc
 }
 
 func isPrimaryContainer(container map[string]any) bool {
--- a/internal/workload/up/build_test.go
+++ b/internal/workload/up/build_test.go
@@ -1,0 +1,173 @@
+// Copyright 2026 DataRobot, Inc. and its affiliates.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package up
+
+import (
+	"errors"
+	"testing"
+
+	"github.com/datarobot/cli/internal/workload"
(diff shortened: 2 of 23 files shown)


internal/workload/up/run.go
```diff
@@ -1076,13 +1080,32 @@ func defaultCodeChange(loaded Loaded, _ Live) (change CodeChange, err error) {
...
+// the artifact claims without touching the working tree, so a deploy changing
+// only an environment variable would otherwise inherit an image built before
+// that sync and report success.
+func imageStale(projectDir string, live Live) bool {
+	if live.CodeVersionID == "" || !projectLinkedFn(projectDir) {
+		return false
+	}
+
+	cfg, err := loadProjectFn(projectDir)
+	if err != nil || cfg.LastBuiltVersionID == nil {
```
damjandr, line 1297: `imageStale` returns `false` when `LastBuiltVersionID` is nil, so an env-only deploy **inherits** the image. That is the opposite of the PR description's "absent … the deploy behaves as it did before": every project predating this release is on the fast path with zero staleness protection until its next build.

damjandr (commented):
Looks good, put some minor comments.

A couple of them (the nil `lastBuiltVersionId` default, the untested `primaryCodeVersionID`) are worth a look before merge — the rest are cosmetic.

internal/workload/up/look_test.go
```diff
@@ -149,6 +149,7 @@ func TestLook_ReadsBothDocuments(t *testing.T) {
 	assert.Equal(t, "gpt-oss-20b-vllm-artifact", live.ArtifactName)
 	assert.Equal(t, "https://app.datarobot.com/workloads/68b0/", live.Endpoint)
 	assert.True(t, live.Locked)
+	assert.Equal(t, "registry.internal/built-by-the-server:sha-abc123", live.ImageURI)
```
damjandr, line 153: `ImageURI` gets an assertion here but `CodeVersionID` gets none, and `primaryCodeVersionID` has no test anywhere. A wrong key path there makes `imageStale` silently always false, which is exactly the hole above.

internal/workload/wapi/config.go
```diff
@@ -30,11 +30,16 @@ import (
...
-	CreatedAt           time.Time `json:"createdAt" validate:"required"`
-	CLIVersion          string    `json:"cliVersion" validate:"required"`
+	ArtifactID          string  `json:"artifactId" validate:"required,dr_id"`
+	CatalogID           *string `json:"catalogId" validate:"omitempty,dr_nonempty_ptr,dr_id"`
+	LastSyncedVersionID *string `json:"lastSyncedVersionId" validate:"omitempty,dr_nonempty_ptr,dr_id"`
+
+	// LastBuiltVersionID is the code version the last image was built from, nil
+	// when none has been recorded. Nothing on the platform holds it, so without
+	// it a deploy that inherits an image cannot tell whether the code moved.
+	LastBuiltVersionID *string   `json:"lastBuiltVersionId" validate:"omitempty,dr_nonempty_ptr,dr_id"`
```
damjandr, line 40: The server already computes `imageOutdated` per container (`manifest/schema.go:91`, called "unambiguous" in `manifest/live.go:977`). Did you check whether it answers the staleness question before adding local project state for it?

internal/workload/up/version.go
```diff
@@ -82,38 +89,200 @@ func buildVersion(
...
 
+// inheritsImage reports whether this run may copy the running version rather
+// than create one, which decides whether it spends a build. Every condition is
+// something the copy would otherwise get wrong.
+func inheritsImage(live Live, plan Plan, opts Options, kind, artifactName string) bool {
+	return plan.RollsArtifact() &&
+		plan.Code.Applies &&
+		!opts.ForceBuild &&
+		!plan.RebuildsImage() &&
+		live.ImageURI != "" &&
```
damjandr, line 143: `live.ImageURI != ""` is taken as proof of a built image, but `internal/workload/artifact.go:91` documents `imageUri` as a *placeholder* until a build completes. The guard needs to tell the two apart, not just test for emptiness.

internal/workload/up/version.go
```diff
@@ -82,38 +89,200 @@ func buildVersion(
...
+			manifest.ArtifactTypeOrDefault(kind),
+			manifest.ArtifactTypeOrDefault(live.ArtifactType),
+		) &&
+		artifactName != ""
+}
+
+// copiedVersion is the version a runtime-only change rolls onto: a copy of the
+// running one, written over with what the file says. What the write leaves
+// behind is read back, since an update replaces a container wholesale.
+func copiedVersion(loaded Loaded, live Live, report *reporter) (version, error) {
```
damjandr, line 151: The copy path never compares the clone's `artifactRepositoryId` with `live.ArtifactRepositoryID`. Every other version path guards the lineage (`sameRepository`, `joinable`); a clone landing elsewhere forks it permanently and silently.

internal/workload/up/version.go
```diff
@@ -82,38 +89,200 @@ func buildVersion(
...
+		return made, nil
+	}
+
+	if plan.InheritsImage {
+		made, err := copiedVersion(loaded, live, report)
 
-		return version{ID: id}, nil
+		// A copy that exists is never abandoned for a create, which would
+		// strand it.
+		if err == nil || made.ID != "" || !copyUnsupported(err) {
```
damjandr: Fallback to create-and-build fires only for clone 404/405/501. A 500 on the new clone endpoint — or any failure after the clone — now aborts a deploy that every earlier release completed; the fallback is always correct, so consider widening it.

internal/workload/up/version.go
```diff
@@ -82,38 +89,200 @@ func buildVersion(
...
+			return specErr
+		}
+
+		copied, cloneErr := copyArtifactFn(live.ArtifactID, loaded.Compiled.ArtifactName)
+		if cloneErr != nil {
+			return cloneErr
+		}
+
+		// Recorded before the update, so a failure below can name it.
+		made = version{ID: copied.ID, Fresh: true}
```
damjandr, line 166: `copied.Status` is discarded here. `matchLock` (`roll.go:204`) skips the already-locked check when `Fresh`, on the stated grounds that "one created by this run is always a draft" — `Fresh` now also covers copies, whose status is the platform's to decide.

internal/workload/up/version.go
```diff
@@ -82,38 +89,200 @@ func buildVersion(
...
+		report.say("  This platform cannot copy an artifact, so the new version is built after all.\n")
 	}
 
 	return createVersion(loaded, repositoryID, labelNewVersion, report)
 }
 
+// inheritsImage reports whether this run may copy the running version rather
+// than create one, which decides whether it spends a build. Every condition is
+// something the copy would otherwise get wrong.
+func inheritsImage(live Live, plan Plan, opts Options, kind, artifactName string) bool {
```
damjandr, line 138: This re-implements `sameArtifactType(loaded, live)` rather than calling it. Consequence today: an agent manifest that omits `type` defaults to `service`, never matches, and never gets the fast path — while `Build` itself reports the type as no drift (`TestBuild_NoTypeInTheFileIsNotDrift`).

internal/workload/up/roll.go
```diff
@@ -98,8 +105,11 @@ func candidateArtifact(
 
 	repository := sameRepository(loaded, live)
 
-	if mode := loaded.Manifest.BuildMode(); mode == manifest.BuildModeDockerfile || mode == manifest.BuildModeGenerated {
-		return buildVersion(loaded, live, code, repository, opts, report)
+	// The same reading the plan used, rather than a second one off the parse
+	// tree: two answers to "does the platform build this image" coming apart is
+	// how a plan describes a deploy that does not happen.
+	if plan.Code.Applies {
```
damjandr, line 111: Swapping `loaded.Manifest.BuildMode()` for `plan.Code.Applies` makes the build/create branch depend on a value the `codeChangeFn` seam supplies. A dockerfile project arriving with `Applies: false` would deploy without ever building — worth asserting the two cannot diverge.

internal/workload/artifact.go
```diff
@@ -404,69 +461,101 @@ func setPrimaryCodeRefInRawArtifact(raw map[string]any, catalogID, catalogVersio
 		return errors.New("artifact: spec.containerGroups missing or empty")
 	}
 
-	codeRef := map[string]any{
+	// Named separately: a group that is not an object and a containers list
```
damjandr: `containerGroups[0]` is now validated *before* the flagged primary is looked for. An artifact whose primary sits in a later group with an empty `groups[0]` now errors where it previously succeeded.

internal/workload/up/version.go
```diff
@@ -82,38 +89,200 @@ func buildVersion(
...
+
+	return build[keyCodeRef] != nil, uri, matched
+}
+
+// keepCodeRef puts the copy's own code reference into the spec about to be
+// written over it. The manifest states none, so the file's spec alone would
+// take it away, and the platform refuses to deploy a container that says how to
+// build itself but not from what. This is the one deploy that never builds, so
+// nothing would put it back.
+func keepCodeRef(spec json.RawMessage, copied *workload.Artifact) (json.RawMessage, error) {
```
damjandr, line 264: The whole fast path rests on `PATCH /artifacts/{id}/` shallow-merging container keys — `imageUri` survives, `codeRef` does not, which is why this exists. That contract currently lives only in a test fake; worth stating it where the code depends on it.

internal/workload/up/plan.go
```diff
@@ -197,6 +209,70 @@ func (p Plan) RollsArtifact() bool {
...
+	}
+
+	return false
+}
+
+// runtimeFields are the fields inside a container this release knows are read
+// when it starts. An allow-list, not a deny-list: the compiler passes unknown
+// keys straight through, so a field the platform grows later would otherwise
+// inherit a stale image and deploy clean.
+var runtimeFields = []string{
```
damjandr, line 232: Raw string literals sitting next to the `keyContainerGroups`/`keyContainers` constants this PR just added, and `a2aEnabled` already exists as `manifest.keyA2AEnabled`.

internal/workload/up/look.go
```diff
@@ -182,9 +198,34 @@ func Look(workloadID string) (Live, error) {
...
 		Endpoint:             workloadDoc.String(keyEndpoint),
 		Locked:               isLocked(artifactDoc.String(keyStatus)),
+		ImageURI:             primaryImageURI(artifactDoc),
+		CodeVersionID:        primaryCodeVersionID(artifactDoc),
 	}, nil
 }
 
+// primaryCodeVersionID reads the catalog version the primary container points
+// at. It is what an inherited image is measured against.
+func primaryCodeVersionID(artifactDoc workload.Document) string {
```
damjandr: `PrimaryContainerInDocument` is walked twice per `Look`. Also the `datarobot`/`catalogVersionId` path is now spelled in a third place, alongside `ExtractCodeRef` and `setPrimaryCodeRefInRawArtifact`.

internal/workload/up/roll_test.go
```diff
@@ -1383,3 +1393,351 @@ func TestRun_FailedBuildOnARollLeavesTheOldVersionServing(t *testing.T) {
...
+func updateSpecStep(tr *track) func(string, json.RawMessage) error {
+	return func(artifactID string, spec json.RawMessage) error {
+		tr.steps = append(tr.steps, "update-spec:"+artifactID)
+		tr.updatedSpec = spec
+
+		return nil
+	}
+}
+
+// readBackStep answers the read the copy path makes once it has written the
```
damjandr: Two doc comments on `readBackStep`; the first ("It records no step…") is a leftover and contradicts the second.

internal/workload/up/build_test.go
```diff
@@ -0,0 +1,151 @@
...
+			assert.Equal(t, c.want, imageStale("/project", Live{CodeVersionID: c.live}))
+		})
+	}
+}
+
+// The record is what the next deploy compares against, and it is the code the
+// sync last pushed. A write that fails costs the next deploy a build it might
+// have skipped, which is what every earlier release did anyway, so it must not
+// fail this one.
+func TestRecordBuiltVersion(t *testing.T) {
```
damjandr, line 125: The loop over `nil`/`error` writes asserts the same thing both times, so the error arm proves nothing beyond not panicking.

internal/workload/up/roll_test.go
```diff
@@ -1383,3 +1393,351 @@ func TestRun_FailedBuildOnARollLeavesTheOldVersionServing(t *testing.T) {
...
+
+			copied := copiedArtifact()
+			f := runtimeOnlyRollOf(&tr, copied)
+			c.breaks(&f, &tr, copied)
+
+			install(t, f)
+
+			result, _, err := runIn(t, envDrift(), Options{NonInteractive: true})
+
+			if c.wantErr != "" || c.wantBuild == "" && len(c.wantAbsent) > 0 {
```
damjandr: `c.wantErr != "" || c.wantBuild == "" && len(c.wantAbsent) > 0` mixes `&&`/`||` precedence; an explicit `wantErr bool` per case would read.

internal/workload/up/run.go
wojtekwdr: That was backwards, and it is flipped now: every answer `imageStale` cannot establish is stale. No link, unreadable project state, no record, and no code reference on the running version all return true. So the first deploy after this release still builds, and the fast path opens once there is a record to compare against.

internal/workload/up/look_test.go
wojtekwdr: Added. `TestLook_ReadsTheImageAndCodeOffThePrimaryContainer` now asserts both fields off a primary that is not the first container, with a sidecar beside it carrying a different image, so a walk falling through to `[0][0]` fails. `TestLook_NoCodeReferenceReadsAsNone` covers the empty case.

internal/workload/wapi/config.go
wojtekwdr: Checked, and it answers a different question. `image_outdated` is true when `imageUri` is set and the latest build row is pending or failed (`schemas/containers.py:449`). The case this guards is a completed build whose code has since moved: `dr artifact code sync` repoints `codeRef` without building, so the last build is still COMPLETED and the flag reads false while the image genuinely predates the code. Nothing server-side records which code an image was built from, which is why the project does.

internal/workload/up/version.go
wojtekwdr: Worth walking through, because I tried the obvious fix first and it was wrong.

I gated on the running artifact having a completed build row. That breaks the feature on the second deploy: the platform keeps build rows on the source when it clones, so a version that is itself a copy has none, and every runtime-only deploy after the first went back to building. It passed the whole suite and only showed up on staging.

For a build-from-source container, non-empty is the provenance. Create nulls `imageUri` on those containers (`services/artifact.py:213`) and the server writes it back only once a build reports COMPLETED (`_reconcile_spec_build_states`), so one that never built carries no image at all. What remains is pre-contract legacy data, which the platform's own lock guard rejects, and that guard is tenant-scoped by URI rather than artifact-scoped for exactly the clone reason above. The comment on the guard says this now.

internal/workload/up/version.go
wojtekwdr: Added `sameLineage`. The copy's repository is compared with the running version's, and a mismatch takes the copy back and falls through to create-and-build, which costs a build but lands in the right lineage. An empty repository on either side is not refused, so the field going missing would not close the path for everybody.

internal/workload/up/version.go
wojtekwdr: Widened. The status no longer decides anything: any copy failure that left nothing behind falls back to create-and-build. The single case that stops the run is a copy that exists and could not be deleted, since creating past that would strand an artifact nothing points at, and the error names it.

internal/workload/up/version.go
wojtekwdr: Went and checked the endpoint. A clone always answers DRAFT: `artifact.status = ArtifactStatus.DRAFT  # Cloned artifact always starts as DRAFT`.

I did add a `Locked` field carrying the status through to `matchLock`, then took it out again. The branch cannot be reached, and a guard whose comment asserts something the platform contradicts is worse than no guard. `matchLock` now says a create and a copy are both drafts, so only a leftover has to be asked about.

internal/workload/up/version.go
wojtekwdr: Fixed, and the consequence was real. There is one `sameArtifactTypeAs` now, used by both, folding case and defaulting only the live side so a file that omits `type` has no opinion rather than meaning service. The four type cases are pinned in a test.

internal/workload/up/roll.go
wojtekwdr: Added `TestDefaultCodeChange_AppliesMatchesTheFilesBuildMode`, which runs the real seam over the manifest shapes and asserts `Applies` equals `BuildMode() == BuildModeBuild` for each, so the two cannot drift apart without a failure.

internal/workload/artifact.go
wojtekwdr: Good catch, that would have broken code sync. The validation runs only after the search for a flagged primary has failed, so a primary sitting in a later group beside an empty `groups[0]` still resolves. The three messages stay separate, and there is a test for that shape.

internal/workload/up/version.go
wojtekwdr: Stated it on `UpdateArtifactSpec`. The contract is a little different from shallow merge: `containerGroups` is replaced wholesale, and the platform re-injects what it owns on a build-from-source container, `imageUri` always and the build pointer only when `codeRef` has not moved (`_restore_server_container_fields`). `codeRef` is not re-injected, which is exactly why the copy path puts it back before writing.

internal/workload/up/plan.go
wojtekwdr: Replaced with constants, in look.go beside the other spec keys this file reads. The manifest package keeps its own: that side writes the file and this side reads the platform's spec, so the duplication is deliberate and the comment says so.

internal/workload/up/look.go
wojtekwdr: Both fixed. `Look` walks the primary once and hands the container to the two readers, so they cannot disagree about which one it is. The catalog path moved into `workload.CodeRefInContainer`, beside the writer that puts it there, leaving two spellings rather than three.

internal/workload/up/roll_test.go
wojtekwdr: Removed.

internal/workload/up/build_test.go
wojtekwdr: Restructured into a table that also covers the two states with nothing to record. The failing-write row does still assert the same value, and deliberately: the claim is that the write is attempted and its answer swallowed, so the config has to be captured before the error is returned. `recordBuiltVersion` returns nothing, so there is no stronger assertion available short of asserting the caller keeps going, which the roll-level test does. Happy to change it if you had something else in mind.

internal/workload/up/roll_test.go
wojtekwdr: Gone. The case struct carries `wantErr string` now and the branch is `if c.wantErr != ""`, so there is no precedence to get wrong.

damjandr (approved):

ajalon1 (approved):
LGTM — no issues found.

<details>
The PR implements a significant optimization: runtime-only changes (env vars, ports, probes, routes) no longer trigger a full image rebuild. Instead, the CLI clones the running artifact version and writes the file's spec over the copy, inheriting the image.

### Key verification points

- **Nil safety**: `PrimaryContainerInDocument` can return nil; all callers handle it safely (Go nil-map reads return zero values, comma-ok type assertions return false). `imageStale` and `recordBuiltVersion` both check `*string` pointers for nil before dereferencing.
- **Copy error handling**: `copiedVersion` records the copy ID before the spec update so failures can name and delete the orphan. `discardCopy` handles the case where deletion also fails, returning a "left behind" error. `copyUnsupported` correctly distinguishes 404/405/501 (old platform, fall back to create) from other errors (don't fall back).
- **Read-back verification**: `carried` reads the artifact back after the spec write to verify the change landed and the image survived. A failed read is treated conservatively (no code, no image, matches=true) which triggers a safe rebuild.
- **Code reference preservation**: `keepCodeRef` re-injects the copy's own `codeRef` into the spec before writing, because the manifest states none and the platform refuses a container that says how to build itself but not from what.
- **`runtimeOnly` allow-list**: Uses structured `Keys` segments (not rendered `Path` strings) to identify container fields, correctly handling container names containing brackets. The allow-list approach means unknown fields default to rebuild — safe for future platform additions.
- **`descend` slice isolation**: `slices.Clone` + append prevents sibling key slices from sharing backing arrays (verified by `TestSubset_SiblingKeysDoNotShareStorage`).
- **`InheritsImage` correction**: After execution, `roll.go` corrects `result.Plan.InheritsImage` to `false` if a build happened or an error occurred, so the JSON envelope reports what actually happened rather than what was planned.

All tests pass across all 10 workload packages. Five parallel review passes covered all 23 changed files.
</details>

PR #856, opened 2026-08-27
[RAPTOR-19755] feat(workload): pull an existing workload's code before up deploys from an empty dir

adamalpi:
# RATIONALE

Binding to an existing source-built workload has no download step. `dr workload config` writes the live spec into `.datarobot.yaml`, but never the *code* the artifact was built from — so `dr workload up` from a directory that doesn't already hold that code rolls a new version from whatever *is* there. An empty directory becomes an empty artifact that fails the build with "cannot detect project runtime", minutes in and after the workload has already been started. Found live (RAPTOR-19755): a deploy synced a single `.drignore` and 422'd at build time.

Ticket: [RAPTOR-19755](https://datarobot.atlassian.net/browse/RAPTOR-19755)

## CHANGES

- On a version-minting `up` (a roll, or a create from source — `plan.MintsVersion()`), and only then, `up` looks at the project directory before it reads the working tree:
  - **Empty directory** bound to a workload whose current artifact has code → **pull that code down** into the directory, so the deploy builds from the real thing instead of an empty tree.
  - **Already looks like a project** (has recognised markers) → untouched; the deploy reads it as it always has.
  - **Files, but no recognised project and not linked** → **refused**, naming the fix, rather than allowed to overwrite the workload's code with whatever is there.
- Retune, start-only, and unchanged runs never reach this — they don't push the tree, so they don't read the live artifact or touch the directory.
- Reuses the existing `filesapi` download + `ExtractCodeRef`; no new API surface.

## Scope / follow-up

Pull-only for now. After a pull, the immediately-following deploy rebuilds an identical version once (the working tree matches what's already running). Seeding the sync base to skip that redundant rebuild is a deliberate follow-up — it means reaching into the sync engine's three-way state, which is not worth perturbing in the same change. Not done here: an `up`-level integration test that drives a full source *roll* — the roll fixtures are all published-image based, and the seed logic itself is covered by six unit tests over the decision tree, with the `MintsVersion` gate proven by the existing roll/retune suite passing unchanged.

---

> [!NOTE]
> **Medium Risk**
> Changes the core `up` apply path and writes to the project directory from remote code before deploy; mistakes could block deploys or pull the wrong tree, though scope is narrow and failures happen before workload mutation.
> 
> **Overview**
> Fixes **RAPTOR-19755**: binding with `dr workload config` only writes the manifest, so `dr workload up` from an empty or wrong directory could mint a version from that tree and fail late (or overwrite live code).
> 
> On **version-minting** runs only (`plan.MintsVersion()`), `apply` now calls **`seedIfMinting`** before starts, rolls, or other mutations. For **source-built**, **bound**, **unlinked** projects it resolves the live artifact’s code catalog via existing `filesapi` / `ExtractCodeRef`, then either **downloads** into a directory that looks empty (CLI residue like `.datarobot.yaml` / `.drignore` still counts as empty), **leaves** a directory that already looks like a real project, or **refuses** with a clear error when unrelated files would replace the workload’s code. Retune, start-only, published-image, linked, and artifact-without-code cases are no-ops.
> 
> Adds **`seed.go`** and six unit tests over that decision tree.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 60374aac3e4bfa5b1d1739dbe8b4f6d359a7318d. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/internal/workload/up/seed.go
+++ b/internal/workload/up/seed.go
@@ -1,0 +1,244 @@
+// Copyright 2026 DataRobot, Inc. and its affiliates.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package up
+
+import (
+	"errors"
+	"fmt"
+	"os"
+	"path/filepath"
+	"sort"
+
+	"github.com/datarobot/cli/internal/drapi/filesapi"
+	"github.com/datarobot/cli/internal/workload"
+	"github.com/datarobot/cli/internal/workload/fileops"
+	"github.com/datarobot/cli/internal/workload/ignore"
+	"github.com/datarobot/cli/internal/workload/manifest"
+	"github.com/datarobot/cli/internal/workload/wapi"
+	"github.com/datarobot/cli/internal/workload/wizard"
+)
+
+// filesClientFn builds the client that pulls an existing workload's code. A
+// seam so tests can serve files without a network.
+var filesClientFn = filesapi.New
+
+// seedIfMinting pulls the workload's code when this run is one that pushes the
+// working tree. Only a version-minting run (a roll, or a create from source)
+// reads the tree; a retune, a start, or an unchanged run leaves it alone, and
+// none of them should read the live artifact or touch the directory.
+func seedIfMinting(loaded Loaded, live Live, plan Plan, opts Options) error {
+	if !plan.MintsVersion() {
+		return nil
+	}
+
+	return seedFromLiveArtifact(loaded, live, opts)
+}
+
+// seedFromLiveArtifact pulls a bound workload's current code into an empty
+// project directory before the deploy reads the working tree.
+//
+// Binding to an existing source-built workload has no download step: `config`
+// writes the live spec into the manifest, but never the code the artifact was
+// built from. So an `up` run from a directory that does not already hold that
+// code would roll a new version from whatever is there — an empty tree becomes
+// an empty artifact that fails the build with "cannot detect project runtime",
+// minutes later and after the workload has already been touched. This closes
+// that gap: when the directory is empty, the code is pulled down so the deploy
+// builds from the real thing; when it holds unrelated files, the run is
+// refused rather than allowed to replace the workload's code with them.
+//
+// It runs only for the one shape that needs it: a source-built workload, bound
+// and already existing, in a directory this CLI has not linked yet. Every
+// other run is left untouched.
+func seedFromLiveArtifact(loaded Loaded, live Live, opts Options) error {
+	if !seedApplies(loaded, live) {
+		return nil
+	}
+
+	codeRef, err := artifactCodeRef(live.ArtifactID)
+	if err != nil || codeRef == nil {
+		// A nil ref (no code catalog: a published image) is nothing to pull.
+		return err
+	}
+
+	files, err := filesClientFn().AllFiles(codeRef.CatalogID, codeRef.CatalogVersionID)
+	if err != nil {
+		return fmt.Errorf("cannot read the code of workload %s to pull it: %w", live.WorkloadID, err)
+	}
+
+	if !hasRealCode(files) {
+		// The workload's own artifact is empty too (only the ignore file, or
+		// nothing). There is nothing to seed, and the build guard elsewhere is
+		// what speaks to a runtime that cannot be detected.
+		return nil
+	}
+
+	return seedLocalDir(loaded, live, codeRef, files, opts)
+}
+
+// seedApplies reports whether this run is the one shape a pull is for: a
+// source-built workload, bound and already existing, in a directory the CLI
+// has not linked yet. A linked project is past first-bind and its incremental
+// sync knows its own base; a published image has no code to pull.
+func seedApplies(loaded Loaded, live Live) bool {
+	mode := loaded.Manifest.BuildMode()
+	if mode != manifest.BuildModeDockerfile && mode != manifest.BuildModeGenerated {
+		return false
+	}
+
+	return live.ArtifactID != "" && !projectLinkedFn(loaded.ProjectDir)
+}
+
+// seedLocalDir decides what to do about the project directory: leave a real
+// project as it is, pull into an empty one, or refuse one that holds files
+// that are not a recognised project and are not this workload's code.
+func seedLocalDir(
+	loaded Loaded,
+	live Live,
+	codeRef *workload.DatarobotCodeRef,
+	files map[string]filesapi.FileMeta,
+	opts Options,
+) error {
+	// A directory that already looks like a project is the normal case: the
+	// user has the code, and the deploy reads it as it always has.
+	if !wizard.Detect(loaded.ProjectDir).SuspectDir() {
+		return nil
+	}
+
+	empty, err := looksEmpty(loaded.ProjectDir)
+	if err != nil {
+		return err
+	}
+
+	if !empty {
+		// Files, but not a recognised project, and not linked to this
+		// workload. Rolling them would replace the workload's code with
+		// whatever this is; deploying its own code needs an empty directory to
+		// pull into, or the project root.
+		return fmt.Errorf(
+			"refusing to deploy: %s has files but no recognised project layout, and workload %s already has "+
+				"code this run would replace. Deploy from the project root, or start from an empty directory "+
+				"to pull the workload's current code first",
+			loaded.ProjectDir, name(loaded, live))
+	}
+
+	return pullCode(loaded.ProjectDir, codeRef, files, live, loaded, opts)
+}
+
+// artifactCodeRef reads the code catalog behind the artifact the workload is
+// running, nil when the artifact is built from a published image rather than
+// from source.
+func artifactCodeRef(artifactID string) (*workload.DatarobotCodeRef, error) {
+	artifact, err := getArtifactFn(artifactID)
+	if err != nil {
+		return nil, fmt.Errorf("cannot read artifact %s to pull its code: %w", artifactID, err)
+	}
+
+	return workload.ExtractCodeRef(*artifact), nil
+}
+
+// hasRealCode reports whether the file set is more than the CLI's own ignore
+// file. An artifact carrying only .drignore has no code worth pulling.
+func hasRealCode(files map[string]filesapi.FileMeta) bool {
+	for path := range files {
+		if path != ignore.FileName {
+			return true
+		}
+	}
+
+	return false
+}
+
+// looksEmpty reports whether the directory holds nothing but the files the CLI
+// writes itself: the state directory, the manifest, and the ignore file. Those
+// are the residue of an earlier bind or a failed run, not the user's code, so a
+// directory holding only them is still a directory to pull into.
+func looksEmpty(dir string) (bool, error) {
+	entries, err := os.ReadDir(dir)
+	if err != nil {
+		return false, fmt.Errorf("cannot read %s: %w", dir, err)
+	}
+
+	for _, entry := range entries {
+		switch entry.Name() {
+		case wapi.RootDirName, manifest.FileName, ignore.FileName, ".git":
+			continue
+		default:
+			return false, nil
+		}
+	}
+
+	return true, nil
+}
+
+// pullCode downloads every file of the workload's current version into the
+// project directory. The directory is empty (looksEmpty gated the call), so
+// the files are written in place rather than staged and swapped.
+func pullCode(
+	dir string,
+	codeRef *workload.DatarobotCodeRef,
+	files map[string]filesapi.FileMeta,
+	live Live,
+	loaded Loaded,
+	opts Options,
+) error {
+	client := filesClientFn()
+
+	paths := make([]string, 0, len(files))
+	for path := range files {
+		paths = append(paths, path)
+	}
+
+	sort.Strings(paths)
+
+	for _, path := range paths {
+		if err := fileops.SafeRelPath(path); err != nil {
+			return fmt.Errorf("the workload's code names an unsafe path %q: %w", path, err)
+		}
+	}
+
+	for _, path := range paths {
+		if err := downloadInto(client, codeRef, dir, path); err != nil {
+			return err
+		}
+	}
+
+	fmt.Fprintf(opts.Stderr, "  Pulled %d file(s) from %s's current version into %s\n",
+		len(paths), name(loaded, live), dir)
+
+	return nil
+}
+
+// downloadInto writes one file of the version into dir, creating parents.
+func downloadInto(client filesapi.Client, codeRef *workload.DatarobotCodeRef, dir, path string) error {
+	dst := filepath.Join(dir, filepath.FromSlash(path))
+
+	if err := os.MkdirAll(filepath.Dir(dst), 0o755); err != nil {
+		return fmt.Errorf("cannot create the directory for %s: %w", path, err)
+	}
+
+	file, err := os.Create(dst)
+	if err != nil {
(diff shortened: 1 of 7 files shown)


internal/workload/up/seed.go
```diff
@@ -0,0 +1,244 @@
...
+	}
+
+	file, err := os.Create(dst)
+	if err != nil {
+		return fmt.Errorf("cannot create %s: %w", path, err)
+	}
+
+	defer func() { _ = file.Close() }()
+
+	if _, _, err := client.DownloadFile(codeRef.CatalogID, codeRef.CatalogVersionID, path, file); err != nil {
```
ajalon1: [P1] Partial download left on disk breaks retries and can corrupt the build

`os.Create` truncates the file, and on a mid-stream `DownloadFile` error the function returns without removing the partial file (the deferred `Close` only closes it). On retry, if the partial is a root marker (e.g. a half-written `pyproject.toml`), `SuspectDir()` returns false, seeding is skipped, and the deploy builds from the corrupt file; if it's not a root marker, `looksEmpty` returns false and the run is refused with a misleading error the user can only resolve by manually deleting files. Fix: `os.Remove(dst)` on the error path, or write to a temp file and rename on success.

👨🏽‍🚀 Also, honestly, couldn't you reuse the code from internal/workload/sync/downloadOne? You'd have to extract it somewhere shared and somehow not use *Engine.

internal/workload/up/seed.go
```diff
@@ -0,0 +1,244 @@
...
+		return nil
+	}
+
+	codeRef, err := artifactCodeRef(live.ArtifactID)
+	if err != nil || codeRef == nil {
+		// A nil ref (no code catalog: a published image) is nothing to pull.
+		return err
+	}
+
+	files, err := filesClientFn().AllFiles(codeRef.CatalogID, codeRef.CatalogVersionID)
```
ajalon1: [P2] Unnecessary `AllFiles` call can block deploys of recognized projects

`seedFromLiveArtifact` calls `artifactCodeRef` (GetArtifact) and `AllFiles` before `seedLocalDir` checks `SuspectDir()`. For a recognized-but-unlinked project (has a Dockerfile/pyproject.toml, not linked), `seedApplies` is true so both network calls fire, then `seedLocalDir` returns nil immediately — the `AllFiles` result is discarded. `AllFiles` is a new network dependency on a path that previously had none; a transient failure blocks a deploy that needed no seeding. Fix: check `SuspectDir()` / `looksEmpty` before the network calls.

ajalon1 (changes requested):
Reviewing the seed logic that pulls an existing workload's code before `up` deploys from an empty directory. Two issues in `seed.go` below; one note on already-merged wizard code (PR #848) that this PR exercises.

**Not inline-anchorable (already-merged PR #848, `model.go`):** `acceptDirectory` rebuilds the draft on a directory change but does not reset `f.nameGiven`. On back-navigation (pick candidate → type name on `screenName` → Escape → pick a different candidate), the new directory's suggested name is shown as a pre-filled value instead of a placeholder, and pressing Enter silently names the workload after the directory (e.g. "src"). The `nameGiven` field exists specifically to prevent this. Fix: `f.nameGiven = f.answers.Name != ""` inside the `chosen != f.detected.Dir` branch. Worth a follow-up since the seed logic relies on the wizard's directory screen.

internal/workload/up/seed.go
ajalon1: Per the 🤖 :

```go
     func DownloadFiles(client filesapi.Client, dir, catalogID, versionID string, files []FileAction) error
     func DownloadOne(client filesapi.Client, dir, catalogID, versionID string, fa FileAction) error
```

internal/workload/up/seed.go
```diff
@@ -0,0 +1,298 @@
...
+// directory holding only them is still a directory to pull into.
+func looksEmpty(dir string) (bool, error) {
+	entries, err := os.ReadDir(dir)
+	if err != nil {
+		return false, fmt.Errorf("cannot read %s: %w", dir, err)
+	}
+
+	for _, entry := range entries {
+		switch entry.Name() {
+		case wapi.RootDirName, manifest.FileName, ignore.FileName, ".git":
```
ajalon1: [P1] Only treat workload state as empty

Please take a look at the `.datarobot` allowance here, especially cases like `.datarobot/cli/state.yaml`. `looksEmpty` currently treats any top-level `.datarobot` directory as CLI residue, but the sync ignore rules only exclude `.datarobot/workload`; sibling state under `.datarobot` may already have internal logic and can be uploaded today. I would recommend checking only `.datarobot/workload` as seed residue, and refusing `.datarobot` when it contains anything else.

ajalon1 (commented):
One P1 finding from review.

ajalon1:
One remaining concern from my first review looks unresolved, but it is not inline-anchorable in this PR because the touched code is in the already-merged wizard flow.

`internal/workload/wizard/model.go::acceptDirectory` still rebuilds `f.detected`, `f.draft`, and `f.envTable` when the user picks a different directory, but it does not reset `f.nameGiven`. On back-navigation, this can keep treating the new directory's suggested name as a typed value instead of a placeholder. I would recommend setting it from the actual answer when the directory changes, e.g. `f.nameGiven = f.answers.Name != ""`, in the same branch that rebuilds the draft.

internal/workload/up/seed.go
adamalpi: Fixed. `looksEmpty` no longer blanket-allows `.datarobot` — a `.datarobot` entry counts as residue only when it holds nothing but `wapi.StateDirName` (`.datarobot/workload`). Anything else under it (`.datarobot/cli/state.yaml`, which the sync uploads) means the directory is not empty and the run is refused rather than seeded over. Added `datarobotHoldsOnlyWorkloadState` + two tests: workload-state-only pulls, `.datarobot/cli` present refuses.

internal/workload/up/seed.go
```diff
@@ -0,0 +1,330 @@
...
+	}
+
+	// The pull is parallel and stops on the first error, but the files that
+	// already landed stay on disk. Left there, a subset that happens to hold a
+	// root marker (pyproject.toml arrived, app.py did not) makes the next run's
+	// SuspectDir() false, so seeding is skipped and the deploy builds from an
+	// incomplete tree — the very thing this pull exists to prevent. The
+	// directory was empty before the pull, so a failure reverts it to empty and
+	// the retry pulls the whole set again rather than inheriting half of it.
+	before, err := topLevelNames(dir)
```
ajalon1, line 315: **Reuse opportunity**: `revertPull`/`topLevelNames` (this file) hand-roll a snapshot-then-diff revert — track top-level directory names before the pull, remove any new ones on failure. `internal/workload/sync/rollback.go` already implements this exact pattern as a tested `Rollback` type (`TrackCreated` + `Restore`), used by the same `sync` package this code's downloader is reused from (see `pullCode`). Worth reusing `Rollback` here instead of a second, independent (and coarser — top-level names only, see the separate finding on partial-overwrite residue) reimplementation, so a future fix to the revert logic doesn't have to land in two places.

👨🏽‍🚀 Reuse your code :)

internal/workload/up/seed.go
```diff
@@ -0,0 +1,330 @@
...
+// source-built workload, bound and already existing, in a directory the CLI
+// has not linked yet. A linked project is past first-bind and its incremental
+// sync knows its own base; a published image has no code to pull.
+func seedApplies(loaded Loaded, live Live) bool {
+	mode := loaded.Manifest.BuildMode()
+	if mode != manifest.BuildModeDockerfile && mode != manifest.BuildModeGenerated {
+		return false
+	}
+
+	return live.ArtifactID != "" && !projectLinkedFn(loaded.ProjectDir)
```
ajalon1, line 143: **Likely bug**: `seedApplies` gates on `!projectLinkedFn(loaded.ProjectDir)`, and in production `projectLinkedFn` is `wapi.Exists` — true the moment `.datarobot/workload` exists at all, regardless of what it contains. That means `seedApplies` returns `false` (and the whole seed path is skipped) in exactly the case the previous commit ("only treat .datarobot/workload as seed residue") set out to *still allow*: a `.datarobot/workload` holding nothing but residue. `looksEmpty`'s tolerance for that residue (below) is never reached on the real path because this check short-circuits first.

`TestSeed_PullsWhenOnlyWorkloadStateIsPresent` passes today only because it stubs `projectLinkedFn` to return `false` directly — it doesn't exercise the real `wapi.Exists` against the `.datarobot/workload` directory the test itself creates on disk. Worth checking whether `projectLinkedFn` needs to distinguish "linked" from "holds only residue," or whether this ordering needs to change so `looksEmpty` runs first.

internal/workload/up/seed.go
```diff
@@ -0,0 +1,330 @@
...
+// directory holding only them is still a directory to pull into.
+func looksEmpty(dir string) (bool, error) {
+	entries, err := os.ReadDir(dir)
+	if err != nil {
+		return false, fmt.Errorf("cannot read %s: %w", dir, err)
+	}
+
+	for _, entry := range entries {
+		switch entry.Name() {
+		case manifest.FileName, ignore.FileName, ".git":
```
ajalon1: `looksEmpty`'s residue allowlist covers `manifest.FileName`, `ignore.FileName`, and `.git`, but not `ignore.LegacyFileName` (`.wapiignore`), which the `ignore` package itself still recognizes as a valid (deprecated) ignore file. A directory whose only residue is a pre-rename `.wapiignore` hits the `default:` branch and gets wrongly refused, even though the equivalent `.drignore` directory would be accepted.

internal/workload/up/seed.go
```diff
@@ -0,0 +1,330 @@
...
+	if err != nil {
+		return nil, fmt.Errorf("cannot read artifact %s to pull its code: %w", artifactID, err)
+	}
+
+	return workload.ExtractCodeRef(*artifact), nil
+}
+
+// hasRealCode reports whether the file set is more than the CLI's own ignore
+// file. An artifact carrying only .drignore has no code worth pulling.
+func hasRealCode(files map[string]filesapi.FileMeta) bool {
```
ajalon1, line 161: `hasRealCode` only excludes `ignore.FileName` when deciding whether the workload's code catalog is worth pulling. `internal/workload/ignore/matcher.go`'s `systemExcludes` doesn't exclude `.datarobot/cli/` from sync, so a catalog built from a project that once carried CLI tool state under `.datarobot/cli/` can include those paths. `hasRealCode` would then treat the catalog as "real code" and pull it down — and the *next* `up` run's `looksEmpty`/`datarobotHoldsOnlyWorkloadState` would see a `.datarobot` holding something other than `wapi.StateDirName` and permanently refuse to deploy until the user manually removes it. Worth widening the exclusion here to match whatever `looksEmpty` tolerates on the way back in.

internal/workload/up/seed.go
```diff
@@ -0,0 +1,330 @@
...
+
+	fmt.Fprintf(opts.Stderr, "  Pulled %d file(s) from %s's current version into %s\n",
+		len(paths), name(loaded, live), dir)
+
+	return nil
+}
+
+// topLevelNames is the set of entries directly in dir, captured before a pull
+// so its leftovers can be told from what was already there.
+func topLevelNames(dir string) (map[string]struct{}, error) {
```
ajalon1, line 334: Minor: `looksEmpty` (line 173) and `topLevelNames` here both call `os.ReadDir` on the same, unmutated directory within the same seeding pass — nothing writes to `dir` in between (the network fetch of `codeRef`/`files` happens between the two calls, not a filesystem write). Not a bug, but `looksEmpty` could hand back the entry set it already read instead of paying for a second `ReadDir`.

ajalon1 (commented):
Follow-up from a deeper /code-review pass — 4 findings, most important first.

internal/workload/up/seed.go
adamalpi: Sharp catch — you are right, and it exposed a wrong framing in my previous commit. `wapi.Exists` keys on `.datarobot/workload` existing, so a directory holding the workload state is *linked*, and seedApplies correctly skips it (the sync engine owns a linked project's base). So "still allow workload-state-only" was never the right goal — that case should skip, not pull. The `datarobotHoldsOnlyWorkloadState` tolerance is now documented as unreachable defense-in-depth; the decision that actually runs on the live path is refusing a `.datarobot` carrying non-workload state (e.g. `.datarobot/cli`). And you nailed the test: `TestSeed_PullsWhenOnlyWorkloadStateIsPresent` only passed by stubbing `projectLinkedFn=false`. Replaced it with `TestSeed_SkipsWhenWorkloadStateExistsOnDisk`, which leaves `projectLinkedFn` as the real `wapi.Exists` against the on-disk `.datarobot/workload` and asserts the directory is skipped.

internal/workload/up/seed.go
adamalpi: Fixed — `looksEmpty` now allows `ignore.LegacyFileName` (`.wapiignore`) alongside `.drignore`, so a directory whose only residue is a pre-rename ignore file is accepted, not refused. Added `TestSeed_PullsWhenOnlyLegacyIgnoreIsPresent`.

internal/workload/up/seed.go
adamalpi: Fixed, and I closed the loop on both ends. `hasRealCode` now treats `.datarobot/`-tree paths (and the legacy ignore file) as residue, so a catalog of only tool state is not pulled. But that alone would still let real-code-plus-`.datarobot/cli` seed the poison, so `pullCode` also filters any `.datarobot/`-prefixed path out of the pull set — seed never writes tool state into the fresh directory, so the next `up` cannot then refuse it. (Root cause is really the sync `systemExcludes` gap that lets `.datarobot/cli` into a catalog at all; this is the seed-side defense.) Tests: `TestSeed_DoesNotPullDatarobotToolStateFromCatalog`, `TestSeed_SkipsWhenCatalogIsOnlyCLIResidue`.

internal/workload/up/seed.go
adamalpi: I looked hard at reusing `sync.Rollback` and it is not a clean drop-in here, for two reasons: (1) `NewRollback` creates its tree under `.datarobot/workload/.rollback` — but seed runs against a fresh, *unlinked* directory that has no `.datarobot/workload`, and creating one would itself make the dir "linked" (`wapi.Exists`), the opposite of what seed wants; (2) the actual write is the reused Engine-free `sync.DownloadFiles`, which does not expose per-file `TrackCreated`, so I could not hand it a `Rollback` without threading one through the shared downloader. `Rollback` is built for overwrite-recovery in an already-linked project; seed`'s shape is "write new files into an empty dir, remove what I added on failure," which the snapshot-diff fits.

You are right that it is coarser, though: `revertPull` restores nothing it did not *create*, so a failed pull that overwrote an allowlisted residue file (e.g. the catalog`'s `.drignore` over an existing one) leaves that file in its partial state. Impact is low — it is not a root marker, and the next successful pull rewrites it — but it is a real gap. If you would like it fully closed I would rather stage into a temp subdir and rename into place (atomic for both the partial-set and the partial-overwrite cases) than adopt `Rollback` with the state-dir coupling above. Happy to do that in a follow-up if you agree it is worth it.

internal/workload/up/seed.go
adamalpi: Agreed it is a redundant `ReadDir`. The two reads live in different functions on different sides of the network gate (`looksEmpty` in `seedFromLiveArtifact` before the artifact/`AllFiles` fetch, `topLevelNames` in `pullCode` after), so folding them means threading the entry set from the first call through `pullLiveCode` into `pullCode`. Given it is one stat of an empty-ish directory and the coupling it would add across three functions, I left it as two reads for now — but happy to thread it if you would prefer the single read.

internal/workload/up/seed.go
ajalon1: got it. file that as a followup then, just thought I'd raise it.

internal/workload/up/seed.go
ajalon1: nah

ajalon1:
Filed as a follow-up since it's outside this PR's diff: [RAPTOR-20092](https://datarobot.atlassian.net/browse/RAPTOR-20092).

ajalon1 (approved):
LGTM. Long road here

ajalon1:
Note for releasebot:  Non-breaking bug fix with a narrow intentional behavior change: `up` now refuses or seeds only in cases that previously produced incorrect or unsafe deploys.

PR #857, opened 2026-08-27
[RAPTOR-19762] feat(workload): highlight the selected picker row with a yellow bar

adamalpi:
# RATIONALE

In the `dr workload up` / `config` wizard, it was hard to see which option was selected. The list pickers (workload binding, base-image selection) marked the cursor row with only a `❯` — the colour cue that was meant to accompany it never showed, because `bubbles/table` colours each cell first and those cells' ANSI resets truncate any style wrapped around the finished row. And the "· in use" tag that marks a bound workload's current choice appeared on the menus but not in the base-image picker.

Ticket: [RAPTOR-19762](https://datarobot.atlassian.net/browse/RAPTOR-19762)

## CHANGES

- **One yellow selection everywhere.** Whatever is under the cursor — a menu option, the advanced-options row, a picker row — is now shown in the same **bold yellow text**, so a selection reads the same across every list the wizard offers. (Previously: menus cyan, pickers/advanced green, and the picker cue invisible.) The yellow uses `GetAdaptiveColor(DrYellow, DrYellowDark)`, so it adapts to the terminal background and stays legible in both **light and dark** themes.

> **Design note — deviation from the ticket.** The ticket describes the selected row as a *full-width black-on-yellow highlight bar*. We built the filled bar and it read as heavy/ugly against the neutral rows, so this PR keeps the lighter **bold-yellow foreground** cue instead. Same intent (an unmistakable, consistent selection colour), lighter treatment. The mechanism the ticket called out is still honoured: the picker's table cells are left uncoloured so the row-level style survives `bubbles/table`'s per-cell resets.
- **The pickers can show it** because their table cells are now left uncoloured; `bubbles/table` renders the cursor row's colour over plain cells cleanly instead of having it swallowed. Non-selected rows go neutral, the conventional look for a selection list.
- **`· in use` in the base-image picker.** The picker now tags the environment a bound workload is already built on and **lifts it to the top** — the list runs long, and the option most likely to be kept should not have to be scrolled for. The tag trails on the right, riding the last cell, and is **plain text**: it lives inside a `bubbles/table` cell, and the table sizes columns and truncates cells by display width with no awareness of ANSI, so a styled tag's escape bytes would inflate the measured column width and be cut mid-escape on a narrow terminal, bleeding colour across the row. The green live-value colour stays on the **menu screens**, which render their own lines and never truncate. Scoped to a bound run: a fresh setup has no live environment, so nothing is tagged and the order is left as it came.
- Removed the now-unused `valueStyle`; folded `selectedRowStyle` into `selectedStyle`.

Verified against the rendered escapes: a selected in-use picker row emits yellow across name/language/id under selection, with non-selected rows plain and no escape sequences embedded in any cell. A forced-TrueColor test guards that the picker cells stay plain text (the no-colour test profile is why an earlier styled tag slipped through), and `columnsFor` now measures display width so its sizing agrees with what the table truncates against.

## Review

Addressed AJ's review ([#857](https://github.com/datarobot-oss/cli/pull/857)):
- **P1** — the in-use tag was a styled string inside a table cell, breaking `columnsFor`'s `len(cell)` sizing and risking a mid-escape truncation by `bubbles/table` (green bleed) on an ~80-column terminal. Fixed by rendering the picker tag as plain text and measuring columns with `lipgloss.Width`; the green tag remains on the menu screens. Added a forced-colour regression test.
- **P3** — the menus rendered the cursor as ASCII `>` while the pickers use `❯`; unified on `❯`.

## PR Automation

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -22,6 +22,7 @@ require (
 	github.com/jeandeaual/go-locale v0.0.0-20250612000132-0ef82f21eade
 	github.com/joho/godotenv v1.5.1
 	github.com/muesli/cancelreader v0.2.2
+	github.com/muesli/termenv v0.16.0
 	github.com/sabhiram/go-gitignore v0.0.0-20210923224102-525f6e181f06
 	github.com/sergi/go-diff v1.4.0
 	github.com/spf13/cobra v1.10.2
@@ -69,7 +70,6 @@ require (
 	github.com/microcosm-cc/bluemonday v1.0.27 // indirect
 	github.com/muesli/ansi v0.0.0-20230316100256-276c6243b2f6 // indirect
 	github.com/muesli/reflow v0.3.0 // indirect
-	github.com/muesli/termenv v0.16.0 // indirect
 	github.com/pelletier/go-toml/v2 v2.4.2 // indirect
 	github.com/rivo/uniseg v0.4.7 // indirect
 	github.com/sagikazarmark/locafero v0.12.0 // indirect
--- a/internal/workload/wizard/choice.go
+++ b/internal/workload/wizard/choice.go
@@ -126,7 +126,9 @@ func (c choice) view() string {
 		line := fmt.Sprintf("%d. %s", i+1, opt.label)
 
 		if i == c.selected {
-			b.WriteString(selectedStyle.Render("> " + line))
+			// The same cursor glyph the pickers use (table.go syncRows), so a
+			// selection reads the same across menus and lists.
+			b.WriteString(selectedStyle.Render("❯ " + line))
 		} else {
 			b.WriteString("  " + line)
 		}
--- a/internal/workload/wizard/choice.go
+++ b/internal/workload/wizard/choice.go
@@ -126,7 +126,7 @@ func (c choice) view() string {
 		line := fmt.Sprintf("%d. %s", i+1, opt.label)
 
 		if i == c.selected {
-			b.WriteString(tui.InfoStyle.Render("> " + line))
+			b.WriteString(selectedStyle.Render("> " + line))
 		} else {
 			b.WriteString("  " + line)
 		}
@@ -140,6 +140,9 @@ func (c choice) view() string {
 			b.WriteString(style.Render("   " + opt.note))
 		}
 
+		// A trailing green tag on the right, its own colour whatever the option
+		// is doing: green means "from the running workload" throughout the
+		// wizard's notes, and the tag is one more live-derived value.
 		if c.liveValue != "" && opt.value == c.liveValue {
 			b.WriteString(liveStyle.Render("   · in use"))
 		}
--- a/internal/workload/wizard/model.go
+++ b/internal/workload/wizard/model.go
@@ -1156,7 +1156,7 @@ func (f flow) execEnvsLoaded(msg execEnvsLoadedMsg) (tea.Model, tea.Cmd) {
 		f.failed = errors.New("no execution environments are available; go back and pick another image source")
 	default:
 		f.execEnvs = msg.environments
-		f.picker = newExecEnvPicker(msg.environments, f.width, f.height)
+		f.picker = newExecEnvPicker(msg.environments, f.liveExecEnvID(), f.width, f.height)
 	}
 
 	return f, nil
--- a/internal/workload/wizard/screens.go
+++ b/internal/workload/wizard/screens.go
@@ -1205,14 +1205,15 @@ func newExecEnvPicker(environments []workload.ExecutionEnvironment, liveID strin
 		// the top: the list runs long, and the option most likely to be kept
 		// should not have to be scrolled for.
 		//
-		// The tag is green and on the last cell, matching the menus' live-value
-		// green and their trailing placement. Last is deliberate: a coloured
-		// cell resets colour where it ends, and only at the row's end does that
-		// reset land past everything, leaving the selection highlight on the
-		// rest of the row intact.
+		// The tag is plain text, not styled. It rides a table cell, and
+		// bubbles/table sizes columns and truncates cells by display width with
+		// no awareness of ANSI: a styled tag's escape bytes would inflate the
+		// column's measured width and, on a narrow terminal, be cut mid-escape,
+		// bleeding colour across the row. The green live-value tag stays on the
+		// menu screens, which render their own lines and never truncate.
 		if liveID != "" && ee.ID == liveID {
 			last := len(row.cells) - 1
-			row.cells[last] += liveStyle.Render(inUseSuffix)
+			row.cells[last] += inUseSuffix
 			inUse, haveInUse = row, true
 
 			continue
--- a/internal/workload/wizard/screens.go
+++ b/internal/workload/wizard/screens.go
@@ -183,7 +183,7 @@ func (f *flow) enterPicker(at screen) {
 		// the list again. On a first visit there is nothing yet, and emptying
 		// the shared table means a failed load cannot leave the previous
 		// screen's workload rows on display under a base-image heading.
-		f.picker = newExecEnvPicker(f.execEnvs, f.width, f.height)
+		f.picker = newExecEnvPicker(f.execEnvs, f.liveExecEnvID(), f.width, f.height)
 
 		if id := f.draft.Build.ExecutionEnvironmentID; id != "" {
 			f.picker.selectValue(func(v any) bool {
@@ -1175,8 +1175,17 @@ func newWorkloadPicker(workloads []workload.Workload, width, height int) rowTabl
 	return newRowTable([]string{"WORKLOAD", "STATUS", "UPDATED"}, rows, true, width, height)
 }
 
-func newExecEnvPicker(environments []workload.ExecutionEnvironment, width, height int) rowTable {
+// inUseSuffix tags the option a bound workload already uses, the same phrase
+// the menu screens append to their live option.
+const inUseSuffix = "  · in use"
+
+func newExecEnvPicker(environments []workload.ExecutionEnvironment, liveID string, width, height int) rowTable {
 	rows := make([]tableRow, 0, len(environments))
+
+	var (
+		inUse     tableRow
+		haveInUse bool
+	)
 
 	for _, ee := range environments {
 		// An environment with nothing built is not offered: picking it would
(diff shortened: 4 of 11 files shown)


ajalon1:
Wonder how this handles light mode vs dark mode.

cc @cdevent

internal/workload/wizard/screens.go
```diff
@@ -1142,15 +1151,48 @@ func newExecEnvPicker(environments []workload.ExecutionEnvironment, width, heigh
...
+		// should not have to be scrolled for.
+		//
+		// The tag is green and on the last cell, matching the menus' live-value
+		// green and their trailing placement. Last is deliberate: a coloured
+		// cell resets colour where it ends, and only at the row's end does that
+		// reset land past everything, leaving the selection highlight on the
+		// rest of the row intact.
+		if liveID != "" && ee.ID == liveID {
+			last := len(row.cells) - 1
+			row.cells[last] += liveStyle.Render(inUseSuffix)
```
ajalon1: [P1] The coloured tag inside this cell breaks column sizing and can be cut mid-escape-sequence by `bubbles/table`.

Two downstream consumers don't know this cell now carries ~24 bytes of SGR codes:

1. `columnsFor` (table.go:158) sizes columns with `len(cell)`, so whenever a bound workload exists the ID column is sized ~20 columns wider than its visible content.
2. Worse: `bubbles/table` truncates every cell with `runewidth.Truncate(value, col.Width, "…")`, and go-runewidth is not ANSI-aware. Verified with a probe against the exact escape bytes lipgloss emits:

```
len(cell)=62  runewidth.StringWidth(cell)=59   // escape interiors counted as visible
Truncate(cell, 40) => "ee-1a2b3c4d5e6f7a8b9c0d1e2f\x1b[38;2;12…"
```

The truncated string ends *inside* the SGR sequence — an unterminated escape that bleeds green over the rest of the row. This fires whenever the overflow branch in `columnsFor` (or the 56-char `maxColumnWidth` cap) shrinks the ID column below the tagged cell's escape-inflated width — i.e. a bound workload on an ~80-column terminal, which is exactly this feature's use case. The tests miss it because they run in the no-colour profile, where `liveStyle.Render` is plain text.

Suggested direction: keep escape sequences out of the measured/truncated path — either measure with an ANSI-aware width consistent with what `bubbles/table` itself sees (`runewidth.StringWidth`, making `Truncate` a no-op) and accept the cosmetic width, or render the picker's suffix as plain text and keep the green tag on the menu screens only.

internal/workload/wizard/choice.go
```diff
@@ -126,7 +126,7 @@ func (c choice) view() string {
 		line := fmt.Sprintf("%d. %s", i+1, opt.label)
 
 		if i == c.selected {
-			b.WriteString(tui.InfoStyle.Render("> " + line))
+			b.WriteString(selectedStyle.Render("> " + line))
```
ajalon1: [P3] Two cursor glyphs inside one wizard.

The PR's stated goal is "one selection look", but the menus render the cursor as ASCII `"> "` here while the pickers render `" ❯"` (table.go `syncRows`). Same semantic, sibling-file inconsistency — worth unifying on `❯` while this PR is already touching both.

ajalon1 (changes requested):
Nice work — the yellow selection now reads consistently across every list in the wizard, and the `· in use` lift in the base-image picker is a real usability win. Thanks for using `GetAdaptiveColor(tui.DrYellow, tui.DrYellowDark)`.

One blocking finding (inline): the coloured tag inside a table cell collides with `columnsFor`'s `len(cell)` sizing and `bubbles/table`'s ANSI-unaware truncation — reproducible with a probe against the real escape bytes. Plus one minor consistency nit on the cursor glyph. Separately, I'd like to have a discussion in the future about making the UI consistent **outside** the workload wizard — today `cmd/templates/list`, `cmd/llm-gateway/select`, `cmd/component/shared` and `tui/hostpicker` each carry their own selection vocabulary (mostly non-adaptive purples) — whether that's adopting this wizard's design (a shared exported `selectedStyle` in `tui`) or something else entirely. Not a blocker for this PR, which is correctly scoped.

ajalon1:
`examples/fastapi/` (app/, app2/, README, Dockerfiles, uv.lock, artifact.json — ~700 lines) doesn't exist on `main` and has nothing to do with this PR's stated fix (picker highlight/ANSI truncation). Looks like local smoke-test scaffolding from the sibling RAPTOR-19755 branch got swept in by a `git add -A`.

It roughly doubles the diff and pulls unreviewed content into the merge. Worth dropping before this goes further so the review can stay on the actual wizard change.

go.mod
```diff
@@ -22,6 +22,7 @@ require (
 	github.com/jeandeaual/go-locale v0.0.0-20250612000132-0ef82f21eade
 	github.com/joho/godotenv v1.5.1
 	github.com/muesli/cancelreader v0.2.2
+	github.com/muesli/termenv v0.16.0
```
ajalon1, line 25: just for the test, right?

ajalon1:
Could you please run the wizard manually and attach a screenshot or terminal capture showing the selected picker row? The ticket asks for a full-width yellow highlight bar, but the current style appears to set only yellow foreground text. A visual proof would make it clear whether the intended selected-row treatment is actually rendering in the TUI.

go.mod
ajalon1, line 25: Non-blocking: if this is only needed to force color in the test, you can avoid promoting `github.com/muesli/termenv` to a direct dependency. A helper can set `CLICOLOR_FORCE=1`, replace Lip Gloss’s default renderer so it re-detects the env, and restore it in cleanup:

```go
func forceColor(t *testing.T) {
    t.Helper()

    prev := lipgloss.DefaultRenderer()

    t.Setenv("CLICOLOR_FORCE", "1")
    lipgloss.SetDefaultRenderer(lipgloss.NewRenderer(io.Discard))

    t.Cleanup(func() {
        lipgloss.SetDefaultRenderer(prev)
    })
}
```

That still makes styles emit ANSI in tests without importing termenv directly.

ajalon1 (commented):
Non-blocking dependency note.

go.mod
adamalpi: Good instinct, but the `CLICOLOR_FORCE` + `SetDefaultRenderer` trick does not work here, and I dug in to be sure why. lipgloss pins each style to the renderer that was default when `NewStyle` ran (`Style{r: r}` in style.go), so the package-level `liveStyle`/`selectedStyle` are bound to the default renderer built at init. `SetDefaultRenderer(NewRenderer(io.Discard))` swaps the *global* default but leaves those styles on the old renderer, whose profile earlier tests in the suite have already lazily cached as `Ascii` — so the tag renders plain and the teeth-check fails. (It passed in isolation only because nothing had cached that renderer's profile yet.) Mutating the pinned renderer's profile in place with `lipgloss.SetColorProfile(...)` is the only thing that reaches them, and naming a profile needs `termenv`. I extracted a `forceColor(t)` helper with this rationale in a comment so it does not look like an accidental promotion. Happy to revisit if there is a lipgloss-native way to name a profile.

ajalon1: Thanks, I'll take something a little more explicit.

adamalpi:
@ajalon1 — thanks for flagging these; you were right that I'd been working the inline review threads and missing the conversation-level comments. Caught up on all three now:

**`examples/fastapi/`** — you nailed it: local smoke-test scaffolding from the sibling RAPTOR-19755 branch that a `git add -A` swept in. Removed it from this branch entirely (rebuilt the fix commit so it's gone from history, not just add-then-removed), so the diff is back to the wizard change only. It's kept locally, and I added `/examples/` to `.gitignore` (on #856, where the scaffolding belongs) so a stray `add` can't pull it in again.

**Light vs dark mode** — the selection colour is `GetAdaptiveColor(tui.DrYellow, tui.DrYellowDark)`, which resolves per the terminal background, so it stays legible in both themes; same for the filter line it matches. Documented that on `selectedStyle`.

**Full-width bar vs the current style** — good catch, and you read the code correctly: it's a bold-yellow *foreground*, not the ticket's black-on-yellow *background* bar. I built the filled bar to compare and it read as heavy/ugly against the neutral rows, so we're deliberately keeping the lighter yellow-text cue. I've updated the PR description with a "deviation from the ticket" note calling this out explicitly. The mechanism the ticket cared about is still honoured — the table cells are uncoloured so the row-level style survives `bubbles/table`'s per-cell resets. If you'd still like a terminal capture of the selected row for the record, happy to attach one.

On the meta-point: fair — I'll read the full conversation, not just the diff comments, from here on.

go.mod
adamalpi: Yes — test-only. `termenv` is imported in exactly one place, `execenv_picker_test.go`, for `lipgloss.SetColorProfile(termenv.TrueColor)`; no production file touches it.

The `go.mod` line isn't a new dependency, just the direct/indirect marker moving: `termenv v0.16.0` was already in the graph via lipgloss, same version on both sides, and `go.sum` is untouched. `go mod tidy` promotes a module out of the indirect block as soon as anything in the main module imports it, test files included — so `task lint` (`go mod tidy -diff`) fails without this hunk.

internal/workload/wizard/model.go
```diff
@@ -1156,7 +1156,7 @@ func (f flow) execEnvsLoaded(msg execEnvsLoadedMsg) (tea.Model, tea.Cmd) {
 		f.failed = errors.New("no execution environments are available; go back and pick another image source")
 	default:
 		f.execEnvs = msg.environments
-		f.picker = newExecEnvPicker(msg.environments, f.width, f.height)
+		f.picker = newExecEnvPicker(msg.environments, f.liveExecEnvID(), f.width, f.height)
 	}
```
ajalon1: **Non-blocking** — filed as [RAPTOR-20231](https://datarobot.atlassian.net/browse/RAPTOR-20231); happy for it to land there instead of here.

`execEnvsLoaded` rebuilds the picker without the `selectValue` re-selection that `enterPicker` applies on re-entry. A flag-driven run (`--build-mode generated --execution-environment X --entrypoint ...`) still passes through this screen, the list loads async, and the cursor lands on row 0 — which is now the live environment wearing the `· in use` tag. An unthinking Enter overwrites the flag's env with the live one. Pre-existing gap, but the lift gives the wrong default new authority. Your own `TestFlow_GeneratedBuildAnswersSurviveBackAndForth` documents this bug class. If you want to take it here:

```suggestion
	default:
		f.execEnvs = msg.environments
		f.picker = newExecEnvPicker(msg.environments, f.liveExecEnvID(), f.width, f.height)

		if id := f.draft.Build.ExecutionEnvironmentID; id != "" {
			f.picker.selectValue(func(v any) bool {
				picked, ok := v.(pickedEnv)

				return ok && picked.id == id
			})
		}
	}
```

ajalon1 (approved):
Approved. Thanks for handling all the previous issues. There's one non-blocking followup. The :robot: tells me that the async picker load doesn't reapply the draft selection, which is a pre-existing issue. Filed as [RAPTOR-20231](https://datarobot.atlassian.net/browse/RAPTOR-20231).

PR #859, opened 2026-08-27
[CFX-7834] refactor(pipeline): validate count flags at parse time

ajalon1:
# RATIONALE
The pipeline list commands did no flag validation at all — the workload family validated, this family didn't, so the same mistake failed differently depending on which command ate it. The latent bugs this closes:

- **`--limit 0` (or negative) silently returned fewer rows than asked.** The old query builder omitted the `limit` param when it was `<= 0`, so the server's default page size applied: a script requesting 100 rows got the default, with no error and nothing in the output saying so.
- **A negative `--offset` silently restarted results from the beginning.** Same omission: the param was dropped, so every page request returned page 1. A paging loop with a computed offset that dipped negative re-fetched duplicate rows forever instead of failing.
- **`pipeline run task logs --tail -5` sent `tail_lines=-5` to the live-log endpoint.** The negative value was forwarded verbatim once the flag was set, producing a server-side failure mid-command — after authentication, deep past the point of feedback.
- **No parse-time gate for CI.** Any of the above burned a full auth + network round trip before surfacing; now cobra rejects them during flag parsing.

Migrating onto `internal/countflags` (from part 1, #858) makes the fix mechanical rather than per-command.

## CHANGES
- `--limit` -> `countflags.PositiveInt` on `dr pipeline list` (default 50) and `version/run/image/input/schedule list` (default 100).
- `--offset` -> `countflags.NonNegativeInt` (default 0) on all six list commands.
- `--tail` on `dr pipeline run task logs` -> `NonNegativeInt`, keeping 0 as "no limit".
- New invalid-limit/invalid-offset reject tests per list command and an invalid-`--tail` test, in the existing `runCmd` style.

## TESTING
`task lint` (all GOOS legs) and `task test` (race + coverage) pass.

## NOTES
Part 2/4 of a stacked review. Based on #858 (base = its branch), so this diff shows only the pipeline changes. Invalid values now error at parse time instead of reaching the API. (Result truncation for large limits is a separate fix, part 4 #861.)

---

> [!NOTE]
> **Low Risk**
> CLI-only flag validation and tests; behavior is stricter for invalid pagination/tail values with no changes to auth or server contracts beyond not sending bad query params.
> 
> **Overview**
> Pipeline **list** commands and **run task logs** now validate pagination and tail flags during Cobra parsing via `internal/countflags`, instead of plain `IntVar` flags that could reach the API with bad values.
> 
> **`--limit`** uses `PositiveInt` (rejects `0` and negatives) on `pipeline list`, and on image/input/run/schedule/version list commands with the same defaults as before. **`--offset`** uses `NonNegativeInt` on those list commands. **`pipeline run task logs --tail`** uses `NonNegativeInt` so negatives fail before auth/network; **`--tail 0`** remains “no limit.”
> 
> Each touched command gains tests that invalid `--limit`, `--offset`, or `--tail` fail at parse time with the expected error messages.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 23c9407abb38a920ac22cd748be0a36d95781651. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/cmd/pipeline/image/list/cmd.go
+++ b/cmd/pipeline/image/list/cmd.go
@@ -18,6 +18,7 @@ import (
 	"fmt"
 
 	"github.com/datarobot/cli/internal/auth"
+	"github.com/datarobot/cli/internal/countflags"
 	"github.com/datarobot/cli/internal/outputformat"
 	"github.com/datarobot/cli/internal/pipeline"
 	"github.com/datarobot/cli/internal/telemetry"
@@ -60,8 +61,8 @@ Example:
 
 	outputformat.AddFlag(cmd, &outputFormat)
 
-	cmd.Flags().IntVar(&offset, "offset", 0, "Pagination offset")
-	cmd.Flags().IntVar(&limit, "limit", 100, "Maximum number of images to return")
+	cmd.Flags().Var(countflags.NonNegativeInt(&offset, 0), "offset", "Pagination offset")
+	cmd.Flags().Var(countflags.PositiveInt(&limit, 100), "limit", "Maximum number of images to return")
 
 	telemetry.TrackWith(cmd, func(_ *cobra.Command, _ []string) map[string]any {
 		return map[string]any{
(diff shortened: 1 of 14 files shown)


sunny2get (approved):
LGTM - thanks for covering this.

PR #860, opened 2026-08-27
[CFX-7834] refactor(task): validate --concurrency at parse time

ajalon1:
# RATIONALE
`dr task run --concurrency` forwarded any value verbatim as `-C <n>` to the task binary, and two of those values did the opposite of what a user would expect:

- **`--concurrency 0` meant unlimited parallelism, not "use the default 2".** go-task treats `-C 0` as "no concurrency limit", so passing 0 — a reasonable reading of "default" — unbounded the fan-out across every task in the graph. With recursive/root Taskfiles that is a quiet resource spike.
- **Negative values went straight through to the child binary** (`-C -1`), leaving its behavior undefined rather than ours.

Both now die at parse time via `internal/countflags.PositiveInt` (from #858): zero concurrency would run nothing, so it is rejected with the standard `invalid argument` message, before the task runner is even looked up.

## CHANGES
- `--concurrency` -> `countflags.PositiveInt` (default 2 unchanged).
- Reject test added in the existing command-test style.

## TESTING
`task lint` (all GOOS legs) and `task test` (race + coverage) pass.

## NOTES
Part 3/4 of a stacked review. Based on #859, so this diff shows only the task change.

### DIFF
--- a/cmd/task/run/cmd.go
+++ b/cmd/task/run/cmd.go
@@ -23,6 +23,7 @@ import (
 	"strings"
 
 	"github.com/datarobot/cli/internal/cli"
+	"github.com/datarobot/cli/internal/countflags"
 	"github.com/datarobot/cli/internal/log"
 	"github.com/datarobot/cli/internal/task"
 	"github.com/datarobot/cli/internal/telemetry"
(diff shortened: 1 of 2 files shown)


chasdr (approved):
👋 chasbot here, chas had me take a pass at this one.

clean. `-C 0` quietly meaning "unlimited" was a real trap, and pushing the check to parse time via `countflags.PositiveInt` kills it before the task runner is even looked up. default 2 is preserved, `Type()` still reports `int` so `GetInt` keeps working, and the reject test needs no task env because parsing dies first. tidy work, you and your droid.

built the two packages and ran the new test on the branch, both green. approving with the stack in mind (this rides on #858/#859).

LGTM 👍

PR #861, opened 2026-08-27
[CFX-7945] refactor(api): unify next-link pagination behind internal/paginate

ajalon1:
Split out of [CFX-7834](https://datarobot.atlassian.net/browse/CFX-7834) and PR stack 858/859/860. While working on that ticket, found two problems with list-API pagination:

1. The pipelines-api list client decodes a single `DataPage` and returns it (`ListPipelines`, `internal/pipeline/pipeline.go:144`), so `--limit 500` silently returns one page — the server enforces a 1..100 page-size ceiling (`maxPageSize = 100`, `internal/workload/workload.go:706`) and larger totals are only reachable via next-links, which the pipeline family never follows.
2. The next-link walk is copy-pasted per endpoint across the workload family (`paginateWorkloads`, `paginateArtifacts`, inline loops in `build.go`, `credential.go`, `execenv.go`, and in `internal/drapi` for templates and llms) instead of sharing one helper.

## Scope

- Add `internal/paginate.Walk`: the one generic next-link loop (accumulate rows, stop at limit or end of pages, return at most limit rows), guarded by the existing `drapi.AssertNextOnSameHost`.
- Convert `ListWorkloads`, `ListArtifacts`, and `ListArtifactBuilds` to share that loop, replacing the per-endpoint `paginate*` copies.
- Pipeline family: pass `offset`/`limit` through unclamped as today, and either adopt the walk or leave it single-page by design — current behavior kept, flagged for the API team re: the 422-on-limit>100 contract.
- Not adopted: llm-gateway list and `ListTaskExecutions` (fetch-all by design).

## Verification

- `task lint` (all GOOS legs) and `task test` (race + coverage) pass.
- Multi-page walk tests cover both the `DataPage` and rows paths; tests pin that out-of-range numerics error instead of clamping.

### DIFF
--- a/cmd/artifact/build/list/cmd.go
+++ b/cmd/artifact/build/list/cmd.go
@@ -15,10 +15,9 @@
 package list
 
 import (
-	"fmt"
-
 	"github.com/datarobot/cli/cmd/artifact/build/internal/buildargs"
 	"github.com/datarobot/cli/internal/auth"
+	"github.com/datarobot/cli/internal/countflags"
 	"github.com/datarobot/cli/internal/outputformat"
 	"github.com/datarobot/cli/internal/telemetry"
 	"github.com/datarobot/cli/internal/workload"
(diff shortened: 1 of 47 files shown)


ajalon1:
Leaving this in draft because I'm not 100% sure on the changes here, unlike for the others in the stack. Still looking for reviews from workload and pipeline, though.

PR #863, opened 2026-08-28
[RAPTOR-19729] fix(workload): recovery never ends at the file, and carries its own exit

cdevent:
# RATIONALE

[RAPTOR-19729](https://datarobot.atlassian.net/browse/RAPTOR-19729). Follow-up to #846, which is already merged and closed two of this ticket's messages. The ticket stayed open because its other half — "`up` must never leave 'delete the file/state' as the way out" — was undone, and because the fix landed on the wrong end of the sequence.

**The message that gets there first was missed.** `deployable()` speaks for the run once the workload is already stuck. `awaitRunning` is what prints at the *moment* it becomes stuck, and it still ended at `"finished as errored; check the logs"` — precisely the dead end the ticket describes, and the reason the folk remedy was hand-deleting `workloadId`, after which the create hit 409 and the conflict advised setting the exact line just deleted.

**Four more refusals ended at deleting the state directory,** because re-pointing a project at another artifact had no command: `dr artifact code init` refused a linked directory and said to delete the directory too. Nothing could be reworded until something existed to reword *onto*.

## CHANGES

**`awaitRunning` carries the same exit as its sibling** — the delete that clears the binding, hedged with `dr workload logs` for `errored` and certain for `terminated`, matching what `nameTaken` and `deployable` already do.

**`dr workload up --recreate`** performs that exit inside the deploy: delete the bound workload, clear the binding, create it again under the same name.
- Acts only on `errored`/`terminated`. Any other state is refused — a workload a deploy can act on is deployed onto, not replaced.
- Asks for the workload name to be typed back (the gesture `confirmLock` already uses for production) unless `--yes`.
- Deletes nothing under `--dry-run`. This is the one ordering hazard: the delete must precede the plan for the plan to describe the create it now is.
- A failed binding-clear does not fail the run — the irreversible half is already done, and the create overwrites the key anyway.

**`dr artifact code init --force`** re-points an already-linked directory in place, via a new `wapi.Relink`. The catalog comes from the artifact named on the command line, exactly as a fresh `init` takes it; the BASE manifest resets, because it describes the tree synced into the artifact being left. The ignore file and `history.log` survive — which is the point, since deleting the directory discards them along with the code catalog.

**Five messages now name a command instead of a path:** `reusedArtifactConflict` (`up`), the locked-artifact refusal in `code sync`, `init`'s own already-linked refusal, and the artifact note `workload delete` prints.

## NOTES

Two sites are **deliberately left**, per scope agreed before implementation: `manifest/write.go`'s `"Delete the file instead"` and `workload config`'s `"Delete it to run setup again"`. Both need `dr workload config --force` before they can name a command. The first is the most literal instance of this ticket's title *and* sits on the supported `dr workload delete --dir` recovery path, so it is the natural next follow-up.

One correction worth flagging: `linkProject` preserves `catalogId`/`lastSyncedVersionId` only because it re-points within the same lineage. An arbitrary relink cannot, so the new messages do not claim it does — what `--force` genuinely buys over a deletion is that it is a command, and that the ignore file and history survive.

This PR also **unbreaks `internal/workload/up`'s tests**, which have not compiled since 6bea35a changed `waitWorkloadFn`'s signature and left one test fake behind. That is unrelated to this ticket but blocked running anything here.

## TESTING

- `task lint` — clean across all three GOOS legs.
- `task test` — green under `-race`. (`internal/plugin`'s `TestDiscoverWithContextSuite` flakes under full-suite load; it passes in isolation and on a clean tree, and that package is untouched here.)
- New: `up/recreate_test.go` (deletes for both dead states, refuses a live one, no-ops unbound, **dry-run deletes nothing**, no-terminal-no-`--yes` refuses, failed clear does not stop the run, failed delete does); `wapi/relink_test.go`; `sync/phase1_locked_test.go`; plus cases in `up/run_test.go`, `init/cmd_test.go`, `init/display_test.go`, `up/cmd_test.go`.
- Regression guards assert `NotContains` on the deletion advice, so these messages cannot quietly revert.
- Verified `--help` and the live already-linked refusal against a real binary; confirmed the refusal leaves `config.json` untouched.
- Live scenario `RAPTOR-19729-E-recovery.sh` added and registered (`task smoke-test-workload`). **Not yet run against a tenant** — it needs a real API token.

---

> [!NOTE]
> **Medium Risk**
> `--recreate` performs irreversible workload deletes on a core deploy path; guards limit it to dead states and require confirmation, but misuse or flag bugs could remove production-adjacent resources.
> 
> **Overview**
> **Workload recovery** adds `dr workload up --recreate`, which deletes a bound workload that is **errored** or **terminated**, clears `workloadId` in `.datarobot.yaml`, and continues the deploy as a fresh create under the same name. It refuses other states, requires typing the workload name (or `--yes`), and under `--dry-run` only prints what would happen—no delete before the plan.
> 
> **Artifact relinking** adds `dr artifact code init --force` and `wapi.Relink` so an already-linked project can point at another artifact without removing `.datarobot/workload/`. Catalog and sync baseline come from the new artifact; ignore file and history are kept, and the previous artifact id is shown for undo.
> 
> **Error and help text** no longer tell users to delete `.datarobot/workload/` or hand-edit bindings. Stuck deploy paths (`awaitRunning`, `deployable`, locked sync, artifact conflicts, `workload delete` notes) now point at `dr workload delete`, `--recreate`, or `init --force`.
> 
> Docs, changelog, unit tests, and live smoke scenario **E** (`RAPTOR-19729-E-recovery.sh`) cover the new behavior.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 853a8724ceb9fc17dd89a3db984fa43317223c87. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


internal/workload/up/recreate.go
```diff
@@ -0,0 +1,151 @@
...
+	workloadName := name(loaded, live)
+
+	if opts.Confirm != nil {
+		return opts.Confirm(fmt.Sprintf(
+			"Workload %s is %s. It will be deleted, which cannot be undone, and created again "+
+				"under the same name.\nType the workload name to recreate it, anything else to stop: ",
+			tui.WarnStyle.Render("`"+workloadName+"`"), live.Status), workloadName)
+	}
+
+	if opts.NonInteractive {
```
wojtekwdr: [high] `Confirm` is nil exactly when `yes || !isStdinTerminalFn()`, and both of those also set `NonInteractive`, so this branch is where a piped-stdin or CI run lands: `dr workload up --recreate < /dev/null` deletes the workload with no prompt and no `--yes`, and the refusal below can't be reached from the command at all. `dr workload delete` decides the same question the other way with `idargs.EnvMayNotConsent`, and the flag help, `docs/commands/workload.md` and the CHANGELOG all promise it asks unless `--yes` is passed.

internal/workload/up/recreate.go
```diff
@@ -0,0 +1,151 @@
...
+// --yes is a caller who has already said so; and no terminal with no --yes is
+// the case that must not proceed silently.
+func confirmRecreate(loaded Loaded, live Live, opts Options) (bool, error) {
+	workloadName := name(loaded, live)
+
+	if opts.Confirm != nil {
+		return opts.Confirm(fmt.Sprintf(
+			"Workload %s is %s. It will be deleted, which cannot be undone, and created again "+
+				"under the same name.\nType the workload name to recreate it, anything else to stop: ",
+			tui.WarnStyle.Render("`"+workloadName+"`"), live.Status), workloadName)
```
wojtekwdr, line 151: [medium] Nothing wires a `Confirm` that returns true, so this `want` argument is never checked. I changed it to a literal "yes" and the whole `up` suite stayed green, which means the type-the-name gesture the flag's safety story rests on isn't asserted anywhere. `confirmRoll` has that test at `roll_test.go:298`.

smoke_test_scripts/workload/RAPTOR-19729-E-recovery.sh
```diff
@@ -0,0 +1,150 @@
...
+
+still="$(wl::dr_jq '.status' workload get "$wl_id" 2>/dev/null || echo gone)"
+[[ "$still" != "gone" ]] \
+    || wl::fail "--dry-run --recreate deleted the workload"
+wl::pass "--dry-run --recreate left the workload alone (status: $still)"
+
+# A preview has to predict the run it previews. Planning against the workload
+# that is about to be deleted described a roll onto something the very next
+# check refuses, while the hint above it said a create would happen.
+previewed="$(printf '%s' "$preview_json" | jq -r '.action')"
```
wojtekwdr: [medium] `workload up --output-format json` prints a `{"up": {...}}` envelope, so `.action` is null here on every run and E.4 fails unconditionally.

```suggestion
previewed="$(printf '%s' "$preview_json" | jq -r '.up.action')"
```

smoke_test_scripts/workload/RAPTOR-19729-E-recovery.sh
```diff
@@ -0,0 +1,150 @@
...
+wl::pass "the dead-deploy message names 'dr workload delete'"
+
+grep -q -- "--recreate" "$WL_SCRATCH/deploy-err.txt" \
+    || wl::fail "the dead-deploy message must name the one-step recovery"
+wl::pass "the dead-deploy message names --recreate"
+
+# The regression that defines this ticket: no message may end at removing a
+# file or a directory the user has to find themselves.
+wl::assert_absent "$WL_SCRATCH/deploy-err.txt" \
+    'delete \.datarobot|delete the file|Delete the file|delete the state' \
```
wojtekwdr: [medium] None of these four alternatives ever appeared in `up`'s output. The strings being guarded against were `delete /abs/path/.datarobot/workload` and `Delete /abs/path/.datarobot/workload to re-init.`, so `delete \.datarobot` can't match with a path sitting between the verb and the dot, and `grep -nE` is case-sensitive besides. Revert the fix and this still passes; something like `[Dd]elete[^ ]*\.datarobot` would catch it.

cmd/artifact/code/init/cmd.go
```diff
@@ -132,6 +148,33 @@ func runInit(cmd *cobra.Command, args []string, outputFormat outputformat.Output
...
+	}
+
+	if err := wapi.Relink(dir, opts); err != nil {
+		return err
+	}
+
+	// The JSON envelope is the same shape a fresh link produces: a caller
+	// parsing it is asking which artifact this directory pushes to, and the
+	// answer does not depend on whether it had a different one a moment ago.
+	if outputFormat == outputformat.OutputFormatJSON {
```
wojtekwdr, line 194: [medium] This returns before `printRelinked` and `initResult` has no field for `previous`, so `--force --output-format json` never emits the id it replaced. The comment a few lines up says that id is the only thing left that remembers where the link came from, and `docs/commands/artifact.md` says the same. A nullable `previousArtifactId` on `initResult` would keep both paths honest.

internal/workload/wapi/relink.go
```diff
@@ -0,0 +1,90 @@
...
+	// Read before write, so a corrupted config is reported as itself rather
+	// than silently replaced by this one. It also carries CreatedAt across.
+	previous, err := LoadConfig(projectDir)
+	if err != nil {
+		return err
+	}
+
+	now := time.Now().UTC()
+
+	cfg := Config{
```
wojtekwdr, line 89: [low] question: `Config` also carries `LastBuiltVersionID` and this literal drops it, so `imageStale` sees nil and the next deploy rebuilds the image. I think that's the right answer after re-pointing at a different artifact, but the comment above settles every other field by name and is silent on this one.

smoke_test_scripts/workload/RAPTOR-19729-E-recovery.sh
```diff
@@ -0,0 +1,150 @@
...
+      containers:
+        - name: primary
+          resourceAllocation: {cpu: 0.5, memory: 512MB}
+YAML
+
+# The deploy is expected to fail: the container exits and the workload lands
+# errored. What is under test is the message it fails with.
+wl::dr_capture workload up --dir "$work" --yes
+if [[ "$WL_RC" -eq 0 ]]; then
+    wl::skip "the workload came up; this scenario needs one that does not"
```
wojtekwdr: [low] `wl::skip` is just an echo and the runner scores `exit 0` as a pass, so a run where busybox happens to come up reports E green having asserted nothing, `--recreate` included. This one also fires before `wl::register_workload`, so the workload the deploy just made is left behind with nothing to clean it up.

wojtekwdr (changes requested):

internal/workload/wapi/relink.go
```diff
@@ -0,0 +1,93 @@
...
+		return err
+	}
+
+	if !Exists(projectDir) {
+		return ErrNotInitialized
+	}
+
+	// Read before write, so a corrupted config is reported as itself rather
+	// than silently replaced by this one. It also carries CreatedAt across.
+	previous, err := LoadConfig(projectDir)
```
chasdr, line 82: does init --force need a guard for the same artifact id? relinking to the artifact already linked resets the BASE manifest and lastSyncedVersionId, so the next sync diffs everything against an empty base and any unsynced local edit becomes a remote-wins conflict with a *.LOCAL copy.

no-op when previous.ArtifactID matches, or is same-id re-init intended?

cdevent: that was a good point added guard in next commit

internal/workload/up/recreate.go
ajalon1: **The piped-stdin half of this is fixed, but the CI env var still consents.**

`dr workload up --recreate < /dev/null` now refuses with `Re-run with --yes`, since `confirmRecreate` reads `Yes` rather than `NonInteractive`. But `run` still wires `Yes: cli.IsNonInteractive(cmd)` (cmd/workload/up/cmd.go:335), and IsNonInteractive merges `DATAROBOT_CLI_NON_INTERACTIVE` into the same signal as the `--yes` flag. A CI user who sets that env var to suppress prompts gets the workload deleted with no typed name and no `--yes`, while `docs/commands/workload.md` and the CHANGELOG both promise it asks for the name to be typed back unless `--yes` is passed.

`dr workload delete` settles this exact question the other way with `idargs.EnvMayNotConsent`, for the reason its confirm test states outright: a CI-wide env var must not consent to an irreversible delete of a workload nobody named. `Yes` here should read the `--yes` flag alone, so the env var lands in `NonInteractive` (no prompts) without doubling as consent to delete.

internal/workload/up/run.go
```diff
@@ -193,7 +227,12 @@ func Run(opts Options) (Result, error) {
 	// while it goes on serving the old value.
 	early := Result{WorkloadID: loaded.WorkloadID(), Env: loaded.Env}
 
-	live, err := lookSettled(loaded.WorkloadID(), opts)
+	// Filled in once, here, because every remedy printed below this line has to
+	// name a command that reaches this project rather than the one the shell
+	// happens to be standing in.
+	opts.dirFlag = dirFlagFor(loaded)
+
+	live, err := lookLive(loaded, opts)
```
ajalon1, line 235: **A run that fails after a successful `--recreate` reports the workload it just deleted.**

`early` (line 228) is seeded from the bound workload before this call, and `lookLive` is where `--recreate` performs the delete and clears the binding. If the delete succeeds but `codeChangeFn` (line 240) or `Build` fails, the run returns `early` as-is, and `reportable` is true because `WorkloadID` is set, so the rendered error envelope names the id of a workload that no longer exists. A JSON caller following up on the failure is pointed at the deleted workload, which is exactly the misleading-recovery shape this PR is closing.

Worth re-seeding `early` from the post-`lookLive` state (the unbound live that follows a recreate carries no id), so a run that has already deleted reports the create-shaped failure it actually is.

ajalon1 (commented):
Follow-ups from re-reviewing the fix commits at bf8be849. Everything else from the earlier round checks out as fixed: woj-2 through woj-7, chasdr's same-id point, and the bugbot dry-run plan each have the fix in the code plus a test. Two P1s: the surviving half of the [high] on the non-interactive consent path is posted as a reply on that thread, and this one is new, on the recreate error path. `go test -race` is green across the affected packages at head.

cmd/workload/del/cmd.go
```diff
@@ -277,5 +279,5 @@ func noteLinkedArtifact(w io.Writer, projectDir string) {
 
 	fmt.Fprintln(w, tui.DimStyle.Render(
 		"This project is still linked to artifact "+cfg.ArtifactID+", which was not deleted with the workload. "+
-			"To unlink it, delete "+wapi.Dir(projectDir)+"."))
+			"To point it at a different one, run 'dr artifact code init --force <artifact-id>'."))
```
wojtekwdr, line 282: [high] This remedy carries no `--dir`, so from a parent directory it names the wrong project. With the manifest in `repo/site` and the shell at `repo`, the pasted command re-points `repo`, and if `repo` isn't linked `runInit` falls through to `wapi.Initialize` and creates a stray `.datarobot/` there instead of failing. The line printed directly above this one already does `manifest.DirFlag(dir)`, and `noteLinkedArtifact` has the project dir in hand. Same omission in `phase1_gather.go:65` and `build.go:134`.

internal/workload/up/recreate.go
```diff
@@ -0,0 +1,161 @@
...
+			"--recreate would delete workload %s (%s) and create it again under the same name. "+
+				"The plan below is the one that would run after it.",
+			live.WorkloadID, live.Status)))
+
+		return Live{State: StateUnbound}, nil
+	}
+
+	agreed, err := confirmRecreate(loaded, live, opts)
+	if err != nil || !agreed {
+		return live, err
```
wojtekwdr, line 84: [medium] Declining the typed-name prompt returns a nil error, so the run carries on with the dead workload and fails a moment later in `deployable` with "...or re-run with `--recreate` to do both in one step", which is the flag just refused. `confirmLock` answers the same gesture with `cancelled: nothing was deployed and workload %s is as it was`.

internal/workload/up/recreate.go
```diff
@@ -0,0 +1,161 @@
...
+		return live, fmt.Errorf("cannot delete workload %s, so it still holds its name: %w", live.WorkloadID, err)
+	}
+
+	clearBinding(loaded, live, report)
+
+	// Unbound rather than missing: the binding is gone from the file, not merely
+	// pointing at something that answers 404, and the render each state produces
+	// is the difference between "will be created" and "is bound to an id that no
+	// longer exists".
+	return Live{State: StateUnbound}, nil
```
wojtekwdr, line 101: [medium] Returning a bare `Live{State: StateUnbound}` leaves `priorBinding` empty, so a `--recreate` run's envelope carries `workloadId: ""` and `plan.priorWorkloadId: ""` and is byte-identical to a first deploy. I checked both: a binding that merely 404s reports `priorWorkloadId: "68b0c1d2e3f4a5b6c7d8e9f0"`, and that's the less destructive of the two paths. `PriorWorkloadID`'s own doc argues the field exists because a stderr line isn't something machine-readable can gate on.

wojtekwdr (changes requested):

PR #867, opened 2026-08-28
[RAPTOR-19525] fix(workload): sync upload integrity — hash-while-streaming, Phase 6 write order, zip overwrite form field

ajalon1:
# RATIONALE

RAPTOR-19525 reported that `dr artifact code sync` silently skips changed files when size and mtime are unchanged. Investigation proved that mechanism has never existed in this codebase (detection is content-hash only; `Diff` receives only hashes, `hashEntries` rehashes every file every run, `fileops.Entry` carries no size or mtime). The investigation did surface real integrity gaps, and this PR fixes them. This is the **fixes** half of a two-PR stack; the follow-up PR adds the `--verify` flag and symlink surfacing on top.

Three distinct defects are fixed here:

1. **Zip-path re-uploads silently stored as `name (2).ext` (the reporter's most likely symptom).** The CLI sent `overwrite=REPLACE` as a URL **query parameter** (`internal/drapi/filesapi/fromfile.go`), but the server's `/files/<catalogID>/fromFile/` route reads `overwrite` only from the **multipart form body** and silently defaults to RENAME when the form field is absent (monorepo `MLData/files/public_api/files/validators.py`). Every re-uploaded file was therefore stored as `name (2).ext`: the original path kept the old bytes while the manifest recorded the new hash, so a following plain `--dry-run` falsely printed `Up to date.` — the reported symptom. Reproduced race-independently (20 of 22 files). Only projects taking the zip path (>20 files or >50 MiB) were affected; the per-file stage route was unaffected because the CLI sends `overwrite` as a JSON body there, which the server does read. Fixed by moving `overwrite` into the multipart form body (fields framed before the file part, Content-Length stays byte-exact).

2. **Torn reads: the manifest could record a hash of bytes the server never received.** Hashing happened at plan time, hours or milliseconds before the upload read; a file rewritten mid-stream (build artifacts, logs) produced a manifest entry describing neither the old nor the new server content. Fixed by hashing while streaming: `io.TeeReader` on the stage path, `io.MultiWriter` on the zip path, so the manifest records exactly the bytes that went over the wire.

3. **Phase 6 wrote config before manifest**, so a crash between the two individually-atomic writes left `config.json` advanced past a stale `manifest.json` — silent, permanent BASE poisoning. Reordered to manifest-before-config so the surviving (non-transactional) window only ever fails safe: an advanced manifest with a stale config makes the next sync detect drift and rebuild BASE from the real remote.

The governing invariant throughout: **BASE (`manifest.json`) must describe the REMOTE, not the local disk.**

Also included: the supporting test infrastructure (a self-consistent fake Files API client that models server state, with fault-injection hooks) and regression suites pinning all of the above, including a codified refutation of the ticket's claimed size/mtime mechanism (a same-size, same-mtime content change via `os.Chtimes` restoration IS detected).

## CHANGES

- `398e2e57` test: rework fakeFilesClient into a self-consistent server model
- `8da0d086` fix: hash uploads while streaming so the manifest records what was sent
- `95de8b9b` test: cover zip uploader hash-while-streaming
- `0b2723d0` test: prove failed/partial uploads never advance persisted state
- `ad7b90c2` fix: write manifest before config in Phase 6
- `4204464d` test: guard sync surroundings against regression
- `7d8ebb5e` test: lock in content-hash-only change detection
- `630e2f40` fix: send zip-path overwrite as a multipart form field

## TESTING

- `task test` (full suite, `-race -shuffle=on`) and `task lint` (GOOS=linux/darwin/windows) green.
- Validated end-to-end against live staging: incremental zip re-uploads now replace in place (22/22 paths hold the new bytes; three-way agreement between disk `shasum -a 256`, manifest, and server `fileChecksum`; zero `name (2).ext` artifacts; a following `--dry-run` truthfully prints `Up to date.`).
- Every new test was confirmed to fail for the right reason against the pre-fix code.

## NOTES

- The server implements its documented OpenAPI contract; the zip defect was on the CLI side. A separate hardening ask to the files-API owners (reject or honor query-param `overwrite` rather than silently defaulting to RENAME) is being filed independently.
- Draft: opened early for team feedback per repo workflow.

---

> [!NOTE]
> **Medium Risk**
> Changes core sync upload and manifest persistence paths (zip overwrite wire format, streamed hashing, write ordering); behavior fixes are high-impact but heavily regression-tested.
> 
> **Overview**
> Fixes three workload sync integrity bugs and adds regression coverage around upload and state persistence.
> 
> **Zip re-uploads (`overwrite`).** `UploadFromZipExisting` no longer sends `overwrite` in the query string (the server ignores it and defaults to RENAME). The value is sent as a multipart form field **before** the file part; streaming multipart framing was extended to support optional form fields with correct `Content-Length`.
> 
> **Manifest matches bytes on the wire.** Stage and zip upload paths hash while streaming (`TeeReader` / `MultiWriter`). Phase 5 stores an `UploadOutcome` with per-path `Sent` hashes; Phase 6 builds `manifest.json` from those streamed hashes and **errors** if any upload path lacks a `Sent` entry (no fallback to plan-time hashes).
> 
> **Safer Phase 6 persistence.** `manifest.json` is written before `config.json` so a partial failure does not advance config ahead of a stale manifest.
> 
> **Tests.** The in-memory Files API fake models server versions, REPLACE merge, zip extraction, pagination, and fault injection; large regression suites cover dry-run, failures, concurrency, classification, and detection behavior.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 630e2f40fe2b786f367ce07d56b3dd08d1bc680a. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/internal/workload/sync/fake_files_client_test.go
+++ b/internal/workload/sync/fake_files_client_test.go
@@ -83,6 +83,11 @@ type fakeFilesClient struct {
 	// follows next links, mirroring the real client's pagination behavior.
 	allFilesPageSize int
 
+	// Downloadable content: versionID → (path → bytes served by DownloadFile).
+	// Opt-in: a test that never registers content gets the loud "not
+	// expected" failure, so unexpected download calls stay visible.
+	downloadContent map[string]map[string][]byte
+
 	// Fault-injection hooks (all opt-in; zero values are no-ops).
 	dropPathFromApply  string // omit this path from the resulting version after apply
 	wrongChecksumPath  string // return wrong checksum for this path in AllFiles
@@ -110,6 +115,23 @@ func (f *fakeFilesClient) withVersion(catalogID, versionID string, files map[str
 
 	f.versions[versionID] = files
 	f.latestVersion[catalogID] = versionID
+
+	return f
+}
+
+// withDownloadable registers the content bytes DownloadFile serves for a
+// version. The registered content must hash to the FileMeta the same test
+// put in the version state — the download path verifies the streamed bytes
+// against the plan's RemoteHash, exactly like the real server relationship.
+func (f *fakeFilesClient) withDownloadable(versionID string, files map[string][]byte) *fakeFilesClient {
+	f.mu.Lock()
+	defer f.mu.Unlock()
+
+	if f.downloadContent == nil {
+		f.downloadContent = make(map[string]map[string][]byte)
+	}
+
+	f.downloadContent[versionID] = files
 
 	return f
 }
@@ -534,8 +556,26 @@ func (f *fakeFilesClient) AllFiles(_, versionID string) (map[string]filesapi.Fil
 	return result, nil
 }
 
-func (f *fakeFilesClient) DownloadFile(_, _, _ string, _ io.Writer) (string, int64, error) {
-	return "", 0, errors.New("fakeFilesClient: DownloadFile not expected")
+// DownloadFile serves content registered via withDownloadable. Without
+// registered content it fails loudly, so a download the test did not plan
+// for surfaces as an error instead of a silent empty file.
+func (f *fakeFilesClient) DownloadFile(_, versionID, path string, w io.Writer) (string, int64, error) {
+	f.mu.Lock()
+	defer f.mu.Unlock()
+
+	files, ok := f.downloadContent[versionID]
+	if !ok {
+		return "", 0, fmt.Errorf("fakeFilesClient: no downloadable content registered for version %s", versionID)
+	}
+
+	data, ok := files[path]
+	if !ok {
+		return "", 0, fmt.Errorf("fakeFilesClient: no content registered for %s in version %s", path, versionID)
+	}
+
+	n, err := w.Write(data)
+
+	return path, int64(n), err
 }
 
 func (f *fakeFilesClient) DeleteFiles(catalogID string, paths []string) (*filesapi.DeleteFilesResp, error) {
--- a/internal/workload/sync/fake_files_client_test.go
+++ b/internal/workload/sync/fake_files_client_test.go
@@ -531,12 +531,21 @@ func (f *fakeFilesClient) AllFiles(_, versionID string) (map[string]filesapi.Fil
 	f.mu.Lock()
 	defer f.mu.Unlock()
 
+	// The counter means "calls received", so it increments even when the
+	// lookup below fails: a test asserting a call count does not want the
+	// total silently lowered by error-path bookkeeping, and the call did
+	// reach the client either way.
 	f.allFilesCalls++
 
-	// Get the files for the requested version from server state.
+	// Get the files for the requested version from server state. An
+	// unregistered version ID is a fixture mistake (usually a typo'd
+	// version string), not an empty remote: the real client surfaces the
+	// server's non-200 as an error here, and returning an empty map would
+	// instead fabricate a plausible-looking plan (every base entry
+	// REMOTE_DELETED, every local file LOCAL_ADDED) that hides the typo.
 	files, ok := f.versions[versionID]
 	if !ok {
-		return make(map[string]filesapi.FileMeta), nil
+		return nil, fmt.Errorf("fakeFilesClient.AllFiles: version %s not found", versionID)
 	}
 
 	// Copy so the caller cannot mutate our state.
--- a/internal/workload/sync/fake_files_client_test.go
+++ b/internal/workload/sync/fake_files_client_test.go
@@ -1,0 +1,638 @@
+// Copyright 2026 DataRobot, Inc. and its affiliates.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package sync
+
+import (
+	"archive/zip"
+	"bytes"
+	"crypto/sha256"
+	"encoding/hex"
+	"errors"
+	"fmt"
+	"io"
+	"sort"
+	stdsync "sync"
+
+	"github.com/datarobot/cli/internal/drapi/filesapi"
+	"github.com/datarobot/cli/internal/workload/fileops"
+)
+
+// fakeFilesClient is a self-consistent in-memory model of the Files API
+// server. Unlike a shallow stub, it records per-path content and its SHA-256
+// so tests can assert "what the server ended up holding" after an upload.
+//
+// Server state is modeled as versions: each ApplyStage (or zip upload) creates
+// a new version whose files are the REPLACE-merge of the staged paths into the
+// prior version. AllFiles serves exactly what was recorded, keyed by version,
+// with configurable pagination via a next-link simulation. ApplyStage returns
+// a numFiles that defaults to the count of ALL files in the resulting version
+// (matching the real API semantics), not just the ones uploaded.
+//
+// Per-method call counters let tests assert that specific network calls were
+// or were not issued. Fault-injection hooks are all opt-in: a test that does
+// not ask for a fault gets a faithful server.
+//
+// Every shared map is guarded by a mutex so -race is clean when the uploader
+// runs concurrently (UploadConcurrency = 4).
+type fakeFilesClient struct {
+	mu stdsync.Mutex
+
+	// Configured IDs returned by the fake server. CreateCatalog returns
+	// catalogID, CreateStage returns stageID, ApplyStage/zip uploads return
+	// versionID.
+	catalogID string
+	stageID   string
+	versionID string
+
+	// Server state: versionID → (path → FileMeta with SHA-256 hash + size).
+	versions map[string]map[string]filesapi.FileMeta
+
+	// Per-catalog latest version ID, for REPLACE-merge semantics.
+	latestVersion map[string]string
+
+	// Current staging area: path → content bytes (stage path only).
+	stagedFiles map[string][]byte
+
+	// Backward-compatible fields: recorded uploaded content and deleted
+	// paths, kept so existing tests that inspect them still work.
+	uploadedFiles map[string][]byte
+	deletedPaths  []string
+
+	// Call counters (guarded by mu).
+	createCatalogCalls int
+	createStageCalls   int
+	uploadToStageCalls int
+	applyStageCalls    int
+	uploadFromZipCalls int
+	allFilesCalls      int
+
+	// Configurable AllFiles page size. 0 = return all at once (no pagination
+	// simulation). When > 0, AllFiles internally splits files into pages and
+	// follows next links, mirroring the real client's pagination behavior.
+	allFilesPageSize int
+
+	// Fault-injection hooks (all opt-in; zero values are no-ops).
+	dropPathFromApply  string // omit this path from the resulting version after apply
+	wrongChecksumPath  string // return wrong checksum for this path in AllFiles
+	wrongChecksumValue string // the wrong checksum to substitute
+	failNthUpload      int    // fail the Nth UploadToStage call (1-indexed; 0 = disabled)
+	failApplyStage     bool   // ApplyStage returns error after staging succeeded
+	numFilesOverride   *int   // if non-nil, ApplyStage returns this numFiles; otherwise real count
+}
+
+// --- Builder methods for test setup (chainable, mutex-safe) ---
+
+// withVersion pre-populates a version in the fake's server state. Useful for
+// testing REPLACE-merge semantics without going through the upload flow.
+func (f *fakeFilesClient) withVersion(catalogID, versionID string, files map[string]filesapi.FileMeta) *fakeFilesClient {
+	f.mu.Lock()
+	defer f.mu.Unlock()
+
+	if f.versions == nil {
+		f.versions = make(map[string]map[string]filesapi.FileMeta)
+	}
+
+	if f.latestVersion == nil {
+		f.latestVersion = make(map[string]string)
+	}
+
+	f.versions[versionID] = files
+	f.latestVersion[catalogID] = versionID
+
+	return f
+}
+
+// withPageSize sets the AllFiles page size for pagination simulation.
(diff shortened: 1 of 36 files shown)


internal/workload/sync/phase6_state.go
```diff
@@ -48,22 +67,46 @@ func phase6State(e *Engine) error {
...
+	//   - SaveManifest fails: config has not been advanced yet, so the next
+	//     sync sees the old version in config, detects drift, fetches
+	//     AllFiles, and rebuilds BASE from the remote — safe and
+	//     self-healing. The manifest write is retried by that same sync.
+	//
+	//   - SaveConfig fails: the manifest is already advanced while config
+	//     still names the old version. The version mismatch makes the next
+	//     sync detect drift and fetch the remote; BASE (the advanced
+	//     manifest) truthfully describes that remote, so the plan is empty
+	//     and the run merely converges config. The rollback dir is already
```
wojtekwdr: [medium] This clause doesn't hold. If SaveConfig fails, the next sync's plan is empty, and an empty plan returns before Execute in `Run` (engine.go:207) and in both codesync paths (cmd.go:233 and :290), so Phase 6 never runs and config keeps the old version until some later sync has real work to do. `TestSaveConfigFailure_ManifestAdvanced_NextSyncResyncs` only shows convergence because it calls `Execute(plan)` directly right after asserting the plan is empty, which no production caller does.

internal/workload/sync/surroundings_regression_test.go
```diff
@@ -0,0 +1,970 @@
...
+	msg := fileops.FormatCaseCollisions(collisions)
+	assert.Contains(t, msg, "case-only path collisions")
+	assert.Contains(t, msg, "Config.yaml vs config.yaml")
+}
+
+// TestCaseCollision_ErrorIsBeforeUpload verifies that the case-collision check
+// runs in Phase 2 (manifests), before Phase 5 (execute) where uploads happen.
+// We cannot create case-colliding files on macOS, so we verify the check
+// position by confirming that a Plan with no case collisions succeeds and
+// the fake's upload counters are zero (Plan does not execute).
```
wojtekwdr: [medium] This doesn't test what the name says. I deleted the case-collision check from phase2_manifests.go and the whole sync package still passed, this test included. Zero upload calls follows from Plan never calling Execute, which is true for every plan, so nothing here pins where the check runs.

internal/workload/sync/surroundings_regression_test.go
```diff
@@ -0,0 +1,970 @@
...
+	plan, err := e.Plan()
+	require.NoError(t, err)
+
+	paths := uploadPathsOf(plan)
+
+	assert.NotContains(t, paths, ".Datarobot/workload/secret.json",
+		"case-folded system exclude must still be excluded")
+}
+
+// TestCaseFoldedIgnorePatterns verifies that case-folding of ignore patterns
```
wojtekwdr: [low] The header says a `.drignore` pattern `BUILD/` must exclude a directory named `build/`, but the comment inside the body and the assertion at the end say the opposite: user patterns are case-sensitive and `*.TMP` doesn't match `scratch.tmp`. Worth fixing the header, otherwise someone grepping for case-folding coverage lands on the test that guarantees its absence.

internal/workload/sync/phase6_state_test.go
```diff
@@ -0,0 +1,305 @@
...
+	require.NotNil(t, result)
+
+	// Verify manifest.json schema.
+	manifest, err := wapi.LoadManifest(dir)
+	require.NoError(t, err)
+
+	assert.Equal(t, 1, manifest.Version, "manifest version must stay 1")
+
+	// Verify the manifest has exactly the expected top-level keys by reading
+	// the raw JSON and checking no unexpected fields were added.
```
wojtekwdr: [low] `assert.Contains` can't catch an added field, so this checks the opposite of what the comment claims. Adding `"newField": 1` to either file leaves all six assertions passing. Unmarshalling into a `map[string]any` and comparing key sets would actually pin the schema.

internal/workload/sync/fake_files_client_test.go
```diff
@@ -0,0 +1,678 @@
...
+
+func (f *fakeFilesClient) AllFiles(_, versionID string) (map[string]filesapi.FileMeta, error) {
+	f.mu.Lock()
+	defer f.mu.Unlock()
+
+	f.allFilesCalls++
+
+	// Get the files for the requested version from server state.
+	files, ok := f.versions[versionID]
+	if !ok {
```
wojtekwdr, line 547: [low] An unregistered version reads as an empty server here, so a fixture typo in a version ID produces a plausible-looking plan (every base entry REMOTE_DELETED, every local file LOCAL_ADDED) instead of an error. `DownloadFile` twenty lines down already fails loudly for exactly this reason. I tried making this return an error and nothing in the package cared, so it looks free.

internal/workload/sync/rollback_discard_test.go
```diff
@@ -0,0 +1,457 @@
...
+// At entry nothing has been persisted, so un-advanced state plus the
+// surviving rollback dir is exactly the recoverable mid-Phase-5 outcome:
+// the next run's stale-rollback restore puts back pre-sync bytes that the
+// un-advanced manifest still matches, and the next diff schedules downloads,
+// not false uploads. Swallowing the error instead (the pre-fix behavior)
+// strands the rollback dir next to ADVANCED state, and that same stale
+// restore resurrects pre-sync bytes as phantom local edits which the next
+// sync silently re-uploads over the remote.
+func TestPhase6DiscardFailure_AbortsBeforeStateWrites(t *testing.T) {
+	if runtime.GOOS == "windows" {
```
wojtekwdr: [low] nit: the chmod fault is inert as root, so this fails rather than skips in a root container. The repo's three other chmod-based fault tests pair the windows guard with `os.Geteuid() == 0` (fsutil_test.go:82, wapi/migrate_test.go:107, workload/del/cmd_test.go:255).

wojtekwdr (approved):
LGTM with some comments

internal/workload/sync/surroundings_regression_test.go
ajalon1: I really need to use a better model for validation. :D

internal/workload/sync/rollback_discard_test.go
ajalon1: nice, good call.

internal/workload/sync/phase6_state.go
ajalon1: fixed. however

Follow-up design observation that fell out of addressing this: after a SaveConfig failure, if the project's disk contents never change again, every subsequent sync pays a drift-triggered AllFiles round-trip and config.json stays stale *indefinitely* — not just "until some later sync has real work". It is safe (BASE advanced + stale config re-triggers drift detection every run and self-heals at the first sync with real work; the reverse write order would be silent permanent poisoning), so we left the behavior as-is and corrected the comment/tests to describe it honestly.

A narrowly-scoped alternative would be a config-only Phase 6 reconciliation when drift was detected AND the plan is empty — but that cuts against the pinned invariant that an empty plan makes no state writes, so it felt like a team decision rather than something to fold into this PR. Happy to file a follow-up ticket if you think the perpetual round-trip is worth closing.

internal/workload/sync/phase6_state_test.go
```diff
@@ -0,0 +1,348 @@
...
+
+	assertTopLevelKeys(t, rawManifest, "manifest.json",
+		[]string{"version", "syncedAt", "syncedVersionId", "files"})
+
+	// Same pin for config.json, transcribed from wapi.Config's json tags.
+	rawConfig, err := os.ReadFile(dir + "/.datarobot/workload/config.json")
+	require.NoError(t, err)
+
+	assertTopLevelKeys(t, rawConfig, "config.json",
+		[]string{"artifactId", "catalogId", "lastSyncedVersionId", "createdAt", "cliVersion"})
```
chasdr, line 296: this is the CI red on both ubuntu and windows. wapi.Config has a 6th field, `LastBuiltVersionID`, and its json tag has no omitempty (only the validate tag does), so config.json always serializes `lastBuiltVersionId` (null when nil). the allow-list here is missing it. add it?

```suggestion
		[]string{"artifactId", "catalogId", "lastSyncedVersionId", "lastBuiltVersionId", "createdAt", "cliVersion"})
```

chasdr (commented):
👋 chasbot here, chas had me take a pass while he's off doing chas things. compliments first, since they're earned, then the one red standing between this and merge.

**the good stuff** (genuinely careful work):
- the three fixes all land where the writeup says. moving `overwrite` into the multipart body matches the server binding it from the parsed form only, and framing the form fields *before* the file part so a streaming parser collects them without buffering a multi-GiB body is the right instinct. Content-Length stays byte-exact.
- hash-while-streaming via TeeReader / MultiWriter, with size taken from the open handle and io.Copy's return instead of the plan-time size. the manifest records the bytes that actually crossed the wire.
- `buildNewBaseManifest` refusing outright on a missing `Sent` entry instead of quietly falling back to the planned hash. resisting that fallback is the whole fix, and you named it in the comment so nobody re-adds it. nice.
- the parallel-upload rework is deadlock-free: resCh buffered to len(files), errored/panicked workers skip the send, drain only on the success path.

**the concern:**
you wrote a test to catch schema drift and it caught some. its own. `TestManifestSchemaUnchanged` is red on ubuntu *and* windows because the config.json allow-list is missing `lastBuiltVersionId`, a field already on wapi.Config whose json tag has no omitempty, so it always serializes. one-line fix, left it inline. the approval up top predates the red, so worth a re-glance before this lands.

also it conflicts with main now. main landed the `--accept-remote` conflict work that overlaps your phase5/phase6 rework (engine.go, phase5_execute.go, phase6_state.go, del/cmd_test.go). rebase, fix the allow-list, re-run CI and this is good.

droid-to-droid nit, non-blocking: the phase6 comment blocks are carrying real ordering/poisoning why, so they stay, but yours and mine could probably land it in half the lines.

internal/workload/sync/phase6_state_test.go
ajalon1: no longer an issue!

ajalon1:
Merged in latest main. Since #898 has already merged and shipped, all the work in here related to zip-overwrite can go.

No longer need `conflictCopies` since that's superseded by `localBackups`.

In the meantime, I've addressed all of Woj and Chas' review comments.

internal/workload/sync/phase6_state.go
ajalon1: Fixed in 9cc35465, 70bb9289, and c6ca6369. The convergence test now asserts the real path through `Run` (a later sync with actual work), the rollback-discard test no longer claims empty-plan convergence, and the Phase 6 comment states the honest config-convergence story.

internal/workload/sync/surroundings_regression_test.go
ajalon1: Fixed in 4b27b8fd. `TestCaseCollision_Phase2StopsBeforeRemoteLoad` pins the check at its Phase 2 call site, so deleting the check now fails the suite.

internal/workload/sync/surroundings_regression_test.go
ajalon1: Fixed in f453cbe1. The header now says what the test pins: user ignore patterns are case-sensitive and only system excludes fold case.

internal/workload/sync/phase6_state_test.go
ajalon1: Fixed in f453cbe1 and 724bf638. Both schemas are now pinned by comparing the parsed document's top-level key sets against an allow-list — which promptly caught the missing `lastBuiltVersionId`.

internal/workload/sync/fake_files_client_test.go
ajalon1: Fixed in 714cfbe8. The fake's `AllFiles` now errors on a version ID with no registered state, matching `DownloadFile`, so a fixture typo fails the test instead of fabricating a plausible plan.

internal/workload/sync/rollback_discard_test.go
ajalon1: Fixed in bb0f29a7 and 64944698. Added `testutil.SkipIfRoot`/`SkipIfWindows` with documented reasons and adopted them at every chmod-fault site, so root containers skip instead of failing setup.

internal/workload/sync/phase6_state.go
```diff
@@ -72,17 +89,34 @@ func phase6State(e *Engine) error {
 	return nil
 }
 
-// buildNewBaseManifest computes NEW_BASE = REMOTE + uploads (local hashes)
-// - deletes, with conflicts resolved as remote-wins.
-func buildNewBaseManifest(e *Engine, syncedVersionID string, syncedAt time.Time) wapi.Manifest {
```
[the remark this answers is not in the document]
ajalon1: Fixed in 05c31434 and 341a4272. The rollback is now discarded at Phase 6 entry, before any state write, so a failed save can no longer strand the rollback dir for Phase 0 to restore; rollback_discard_test.go pins that the failure paths schedule no restore and no false uploads.

chasdr:
no, it should be phase 7 On Sep 23, 2026, at 8:06 PM, AJ Alon ***@***.***> wrote:﻿
@ajalon1 commented on this pull request.



In internal/workload/sync/phase6_state.go:
> @@ -72,17 +89,34 @@ func phase6State(e *Engine) error {
 	return nil
 }
 
-// buildNewBaseManifest computes NEW_BASE = REMOTE + uploads (local hashes)
-// - deletes, with conflicts resolved as remote-wins.
-func buildNewBaseManifest(e *Engine, syncedVersionID string, syncedAt time.Time) wapi.Manifest {

Fixed in 05c3143 and 341a427. The rollback is now discarded at Phase 6 entry, before any state write, so a failed save can no longer strand the rollback dir for Phase 0 to restore; rollback_discard_test.go pins that the failure paths schedule no restore and no false uploads.

ajalon1:
Reproduction findings for [RAPTOR-19525](https://datarobot.atlassian.net/browse/RAPTOR-19525) are on the ticket (comments from 2026-09-24), including both live-repro scripts and staging logs.

TL;DR: the reported size+mtime detection bug does **not** reproduce — at the reported version (`a203525c`) change detection is already content-hash based, which this PR pins with `detection_regression_test.go`. The impact described in the ticket matches the zip re-upload overwrite bug instead; a live repro of that one is pending staging stability and will be added to the ticket.

ajalon1:
> no, it should be phase 7

Great catch @chasdr ! I've made what I think is a minimal set of change to address this, but I want to tackle phase numbering separately, as this PR is already too large in scope for my taste.

PR #868, opened 2026-08-28
[RAPTOR-19525] feat(workload): add --verify flag and surface skipped symlinks

ajalon1:
# RATIONALE

Stacked on the integrity-fixes PR (base: `aj/raptor-19525-integrity-fixes`). That PR fixed how the CLI writes state; this one answers "how do I find out my manifest is already lying to me?" and "why did my file never make it to the server?"

**`dr artifact code sync --verify`** forces a remote listing (`AllFiles`), reports any path where BASE diverges from REMOTE (`hash_mismatch` / `base_only` / `remote_only`) as precise per-path stderr notices, repairs the manifest from the real remote (including the empty-plan case, where the plan has nothing to do but the manifest still needs rewriting), and — after applying — re-fetches the new version and compares each uploaded path's server checksum against the bytes actually streamed, failing before any state is persisted on mismatch. Exit code stays 0 when divergence is found and repaired; diagnostics never fail a sync. `--verify` is deliberately opt-in (one or two extra `AllFiles` GETs) and never folds into `--dry-run`/`--diff` semantics; nothing is persisted that would distinguish a verified run.

**Skipped-symlink surfacing**: symlinks have always been skipped silently, so a file absent from the remote was invisible. The walker now reports each skipped symlink with its kind (file vs directory, resolved via `os.Stat`; dangling links reported with empty target), filtered through the same ignore matcher and system-exclude set as regular files (a `.drignore`-matched or `.git` symlink is neither uploaded nor announced). Phase 2 emits a bounded, deterministic stderr notice (first 5, then "and N more"; distinct file vs directory wording; directory wording conveys the whole subtree is omitted), and `PlanJSON` gains an always-present `skippedSymlinks` field (explicitly `[]` when empty, mirroring `locked: false`). Human output for a project with no symlinks is byte-identical to before.

## CHANGES

- `d53db1ec` test: model downloads in the fake client and pin the `.wapiignore` shadow warning
- `d33e4e25` feat: add `--verify` flag, carried into engine options and telemetry
- `15f1be5f` feat: detect BASE-vs-REMOTE divergence under `--verify`
- `3316d3ea` feat: repair a poisoned manifest from the real remote even when the plan is empty
- `43bdf65c` feat: verify after apply that the server holds what was uploaded
- `42874320` feat: report symlink kind and filter ignored symlinks at collection site
- `3cb231d5` feat: surface skipped symlinks via stderr and PlanJSON
- `ecad32bf` test: prove the three fixes compose in combined integration scenarios
- `9ec96a0c` test: add Windows unit coverage for SHA-256 parity, path formatting, and no-symlink sync
- `2f80eac2` test: harden test quality where scrutiny flagged weaker-than-claimed assertions
- `e7baa205` test: pin sync temp cleanup and checkout/versions surroundings
- `51b07b8a` refactor: drop the vestigial query parameter from newStreamingMultipartRequest
- `9a35a9c0` docs: state the shallow-copy invariant at the Phase-6 config copy site
- `05fb1263` test: sweep the actionable symlink-visibility scrutiny nits
- `2723cb15` fix: make `--verify` divergence prose honest about what the run writes
- `68a420e4` docs: document sync overwrite semantics, `--verify`, and symlink notices

## TESTING

- `task test` and `task lint` green; the `--verify` half was validated end-to-end against live staging (26/26 behavioral assertions), including: flagship poisoned-manifest scenario reproduced (`Up to date.` lie) then caught and repaired by `--verify` with exit 0; idempotent repair (second `--verify` silent, manifest byte-identical); post-apply verification cost pinned at exactly two `AllFiles` GETs for a non-drifted run with uploads; JSON mode pure on stdout to EOF with all prose on stderr.
- Combined integration tests prove hash-while-streaming, `--verify` repair, and symlink surfacing compose in a single run (including a 22-file zip-path project with a divergence and both symlink kinds).

## NOTES

- Draft for team feedback. The full stack has now been through the final validation pass: the symlink-visibility milestone validated 25/25 behavioral assertions against live staging plus engine-level evidence, and 89/89 assertions pass mission-wide. Windows unit coverage, test-quality hardening, and divergence-prose polish have all landed.
- Locking an artifact is one-way/irreversible on staging, so the locked-artifact interaction is covered at the engine level rather than against a live locked artifact.

---

> [!NOTE]
> **Medium Risk**
> `--verify` changes when sync executes and can fail runs on post-apply checksum mismatch or rewrite manifest state; default sync behavior is intended unchanged, but the new repair and verification paths touch core sync state handling.
> 
> **Overview**
> Adds **`dr artifact code sync --verify`**, an opt-in integrity mode that forces a remote listing on non-drifted artifacts, reports **BASE (manifest) vs REMOTE** mismatches on stderr and in plan JSON (`divergence`), and still **applies** when needed—including **empty plans** that only repair a poisoned manifest via Phase 6 instead of stopping at “Up to date.” Human and JSON output keep diagnostics on **stderr**; JSON plan documents always include explicit `divergence` and `skippedSymlinks` arrays (empty when clean). Telemetry always emits `verify`.
> 
> **Skipped symlinks** are no longer silent: the walker passes **file vs directory** (and dangling links) into the sync engine, stderr gets a bounded, sorted summary, and symlinks ignored by `.drignore`/system excludes are not announced.
> 
> Smaller surrounding changes: **`versions`** text mode shows **Local synced to:** when known; **checkout** gains a regression test that **`.tmp-*` staging dirs** do not linger; **multipart upload** drops an unused URL-query parameter on `newStreamingMultipartRequest`; **docs** cover overwrite semantics, `--verify`, and symlink behavior. Large **integration/unit test** coverage for verify, display composition, downloads, and cross-fix scenarios.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 68a420e42f2a32f97a38b713b6e6458de64ba792. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


internal/workload/sync/divergence.go
```diff
@@ -0,0 +1,129 @@
...
+// deterministic across runs.
+func detectDivergence(base, remote BaseManifest) []Divergence {
+	var out []Divergence
+
+	for path, b := range base {
+		r, ok := remote[path]
+
+		switch {
+		case !ok:
+			out = append(out, Divergence{Path: path, Kind: DivergenceBaseOnly, BaseHash: b.Hash})
```
wojtekwdr, line 69: [high] A `base_only` finding says BASE is wrong about this path, but `Classify` still trusts BASE for it: non-drifted artifact, local matching BASE, path missing from the remote gives `ClsRemoteDeleted`, and Phase 5 removes the user's only copy with no conflict copy and exit 0. I reproduced it with the fake's own `withDropPath`: one plain sync against a server that drops an upload, then `sync --verify`, and the file is gone. `divergence_test.go:373` pins it as intended, so is "remote wins" really right when the run has just proved the remote never held it?

cmd/artifact/code/codesync/cmd.go
```diff
@@ -205,18 +234,104 @@ func runSync(cmd *cobra.Command, outputFormat outputformat.OutputFormat, deps De
...
+
+	switch {
+	case flags.DryRun:
+		return head + " This was a preview; nothing was written. Run without --dry-run to reconcile."
+	case flags.Diff:
+		return head + " This was a preview; nothing was written. Run without --diff to reconcile."
+	case plan.IsEmpty():
+		return head + " The plan is empty, but manifest.json is being rewritten from the server's state to repair them."
+	default:
+		return head + " The plan reconciles them."
```
wojtekwdr, line 302: [medium] This is printed in `runSync` before the conflict prompt, so quitting at the prompt, or JSON mode returning early on conflicts without --yes, leaves "The plan reconciles them." on stderr for a run that executed nothing. The dry-run, diff and empty-plan branches are all careful about exactly this; the conflict-abort mode is the one that got missed.

internal/workload/sync/log_capture_test.go
```diff
@@ -0,0 +1,69 @@
...
+// Tests using this helper must not call t.Parallel: the os.Stderr swap is
+// process-global while fn runs.
+func captureWarnLog(t *testing.T, fn func()) string {
+	t.Helper()
+
+	r, w, err := os.Pipe()
+	require.NoError(t, err)
+
+	origStderr := os.Stderr
+	os.Stderr = w
```
wojtekwdr, line 49: [low] `os.Stderr = w` has no defer, and every call site passes a closure containing `require.*`. The first failure Goexits past the restore, past `w.Close()` and past the `t.Cleanup(log.StopStderr)` registration, so the rest of the package logs into an orphaned pipe and loses its warn output exactly when you need it.

cmd/artifact/code/codesync/combined_integration_test.go
```diff
@@ -0,0 +1,332 @@
...
+	require.NoError(t, err)
+
+	outText := stdout.String()
+	errText := stderr.String()
+
+	// stdout carries only the plan (the upload row).
+	assert.Contains(t, outText, "a.py", "stdout must carry the plan")
+	assert.NotContains(t, outText, "Moved local state", "no migration notice on stdout")
+	assert.NotContains(t, outText, ".wapiignore", "no ignore notice on stdout")
+	assert.NotContains(t, outText, "was not uploaded", "no symlink notice on stdout")
```
wojtekwdr: [low] These two negatives cannot fail. "was not uploaded" and "divergence:" are the phase wording from `symlink.go` and `divergence.go`, which reach the real stderr via `log.Warn` and never appear at all under `fakeEngine`; the cmd summaries say "were not uploaded or synced" and "--verify found N divergence(s)". Routing `format.StateNotice` to stdout leaves both green. The "Moved local state" negative beside them does have teeth.

internal/workload/fileops/walk.go
```diff
@@ -95,13 +101,29 @@ func walkVisit(
...
 
 	target, _ := os.Readlink(absPath)
-	onSymlink(relPath, target)
+
+	// Stat follows the link. A failure means the target does not exist
+	// (dangling) or is inaccessible: report an empty target and isDir
+	// false so the caller can label it without a walk error.
+	info, err := os.Stat(absPath)
+	if err != nil {
+		onSymlink(relPath, "", false)
```
wojtekwdr: [low] A directory symlink whose `os.Stat` fails reports isDir false, so phase 2 matches it as a file and announces a link the user ignored: `node_modules/` only matches with isDir true, so a dangling `node_modules` link gets warned about anyway, and with the "not a regular file" wording. Splitting `fs.ErrNotExist` from the other Stat errors would keep the file-vs-directory contract intact.

internal/workload/sync/upload_zip.go
```diff
@@ -75,17 +76,47 @@ func (ZipUploader) ApplyUploads(e *Engine, files []FileAction) (UploadOutcome, e
...
+		err = closeErr
+	}
 
-	defer func() { _ = zw.Close() }()
+	if err != nil {
+		// The remove itself can fail — a Windows sharing violation was the
+		// original leak — so a discarded error here would hide exactly the
+		// recurrence this cleanup exists to prevent. Log it instead.
+		if rmErr := os.Remove(tmp.Name()); rmErr != nil {
+			log.Debug("zip temp cleanup failed; the archive may be stranded in the system temp dir",
```
wojtekwdr: [low] The comment just above says a discarded error here would hide the recurrence this cleanup exists to prevent, then it logs at `log.Debug`, which is off at the default level for both the stderr and the file logger. `engine.go:313` uses `log.Warn` for the comparable lock-release cleanup failure.

internal/workload/sync/divergence.go
```diff
@@ -0,0 +1,129 @@
...
+// the phase via log.Warn: the log writes stderr immediately, so the findings
+// survive a later phase failing and never touch stdout in JSON mode.
+func maybeDetectDivergence(e *Engine) {
+	if e.drifted || !e.opts.Verify {
+		return
+	}
+
+	e.divergences = detectDivergence(e.base, e.remote)
+
+	for _, d := range e.divergences {
```
wojtekwdr: [low] nit: one `log.Warn` per divergence with no bound, and `divergenceSummaryNotice` joins every path into a single line, while the symlink notices beside them cap at `SymlinkNoticeBound` with an "and N more" tail. A wholesale-poisoned manifest is the case --verify exists for, and it is the case that prints a line per file plus one very long one.

cmd/artifact/code/codesync/cmd.go
```diff
@@ -118,8 +126,11 @@ versioned step.
 
 Use --dry-run to preview the plan without writing anything; --diff to
 also print per-file unified diffs. Both modes exit before any remote
-write. --yes auto-confirms the post-plan prompt and skips any
-interactive directory prompt.
+write. --verify forces a remote round-trip and post-apply verification
+of what was uploaded, catching server-side divergence the ordinary fast
+path cannot see; it composes with --dry-run and --diff. --yes
```
wojtekwdr: [low] "post-apply verification of what was uploaded ... it composes with --dry-run and --diff" reads as though the post-apply half runs under a preview, but `skipsExecute` returns true there so `verifyPostApplyUploads` is never reached. Worth splitting the two halves: the round-trip composes, the post-apply check only happens on an applying run.

wojtekwdr (changes requested):

PR #871, opened 2026-08-31
[RAPTOR-19916] fix(workload): update dr workload commands wording

wojtekwdr:
# RATIONALE

A QA pass over `dr workload config` and `up` against staging found four user-facing strings where the behaviour is right and the sentence describing it is not. None is a behaviour bug. A fifth turned up while reviewing the fix.

Two are worth more than wording. When a binding no longer resolves, `up` treats it as drift and creates a new workload, warning that `.datarobot.yaml is bound to 68b00000, which no longer exists`. Comparing that id against the file is the only check that catches a token resolving to a different instance or organisation, and eight characters is not enough to do it with. Separately, the refusals that tell you to delete a dead workload and deploy again said it comes back "under the same name": true, and it hides that the replacement gets a new id and therefore a new endpoint URL, permanently.

## CHANGES

**Where an ambient id came from.** `Stop workload <id>, named by .datarobot.yaml?` meant *designated by*, but the manifest carries a real `name:` key, so it read just as easily as "its name comes from this file". It now says `The id is specified in .datarobot.yaml rather than on the command line.`, which is also the half that says why a mutating command is asking. `stop`, `start` and `delete` share one prompt skeleton so they cannot drift apart again, and the clause is empty for an id you typed.

**The full workload id** in the recreate warning. `shortID` stays for the plan header, which only labels something rather than asking you to recognise it.

**What recreating costs.** All three delete-and-redeploy remedies now say the name survives but the id and the endpoint URL do not, so anything calling the old URL has to be repointed.

**One path in the manifest-above warning.** It named the directory you passed and then the ancestor the search was handed, reading as though it were about two places. Fixed at the root: `config` was the only command in the family not routing `--dir` through the shared guard, so a typo produced that warning, then a warning about missing project files, then an error about a missing Dockerfile, none of which said the directory was not there. It now refuses by name like `up` and `delete` already did. This is the only behaviour change in the PR.

**The locked-version confirmation** no longer claims the version is production. Locking makes an artifact immutable; why it was locked is not something the CLI knows.

Help text and docs moved with the code so one concept is not described two ways. That picked up a `--yes` section in `docs/commands/workload.md` claiming the prompt is skipped when stdin is not a terminal, which it is not: the command fails with `confirmation required: pass --yes`.

## PR Automation

> [!IMPORTANT]
>

---

> [!NOTE]
> **Low Risk**
> Mostly CLI messaging and shared helpers; the sole functional change is earlier `--dir` validation on `dr workload config`, with no auth, API, or deploy-logic changes.
> 
> **Overview**
> Improves **user-facing copy** across workload commands so manifest-resolved ids are described as *specified in* `.datarobot.yaml` rather than *named by* it, and **stop/start/delete** share one confirmation shape via `idargs.Prompt` with optional consequence text.
> 
> **`dr workload up`** now prints the **full bound workload id** when a stale binding triggers recreate (so users can compare against the manifest), adds a shared **`recreateNote`** on errored/terminated/name-conflict refusals explaining that delete-and-redeploy keeps the name but yields a **new id and endpoint URL**, and drops **“production”** from the locked-version roll prompt (lock ≠ production).
> 
> The only **behavior change**: **`dr workload config --dir`** routes through **`idargs.ProjectDir`**, so invalid `--dir` paths fail immediately like `up`/`delete` instead of cascading into misleading manifest/Dockerfile errors. The config wizard also **suppresses** the “manifest above” warning when the parent walk cannot start (`ErrNotADirectory`).
> 
> Docs, CHANGELOG, and tests track the wording and validation updates.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 83712d75e701778aca4d169f668ea5d6c0da94ae. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


chasdr (approved):
LGTM 👍

cmd/workload/config/cmd_test.go
```diff
@@ -332,3 +332,21 @@ func TestCmd_UnparseableEnvFileLeaksNoValues(t *testing.T) {
 	assert.NotContains(t, stderr.String(), "sk-live-do-not-print")
 	assert.NotContains(t, stderr.String(), "hunter2")
 }
+
+// A --dir that is not a directory is refused by name, the way 'up' and 'delete'
+// already refuse it. This command used to walk on into a path that was not
+// there and report the missing Dockerfile instead, three messages after the one
+// fact that explains them, with a manifest-search warning in between naming an
+// ancestor the user never typed.
```
ajalon1: Appreciate docs explaining why a test exists. Most of these are self explanatory, but future us will definitely forget one day and just remove them thinking they aren't necessary.

ajalon1:
Since you're doing a thorough pass on documentation, might if I ask @smagee-robot to take a look?

cmd/workload/del/cmd_test.go
```diff
@@ -29,6 +29,10 @@ import (
 	"github.com/stretchr/testify/require"
 )
 
+// boundID is the id a manifest binds, matching the fixture the idargs and stop
+// packages keep under the same name.
+const boundID = "68b0c1d2e3f4a5b6c7d8e9f0"
```
ajalon1: I'd suggest making this an obviously fake value, otherwise it'll trigger a bunch of security bots

PR #872, opened 2026-08-31
[RAPTOR-19348] fix(artifact): don't prompt in code sync --dry-run and emit one JSON document

adamalpi:
# RATIONALE

Two of the three defects reported in [RAPTOR-19348](https://datarobot.atlassian.net/browse/RAPTOR-19348) against `dr artifact code sync` — the two quick, self-contained ones. The third (bidirectional remote-wins overwrite / no push-only mode — the data-loss headline) is an engine change with open product decisions and is being handled separately.

### D2 — `--dry-run` no longer hangs on a prompt

`runSync` resolved the project directory with `dirprompt.ResolveDir(dirFlag, flags.Yes, …)`, which prompts whenever `--dir` is absent and `--yes` is not set — it never looked at the preview flags. So `dr artifact code sync --dry-run` (no `--dir`/`--yes`) **blocked on `Project directory [.]:` before it ever built the plan**, for an operation that by definition writes nothing. The reporter hit a 2-minute timeout and had to discover `--dry-run --yes`.

- A preview (`--dry-run` or `--diff`) now resolves the directory **non-interactively** (`runFlags.Preview()`), defaulting to `.` without asking.
- The post-plan confirmation was already skipped for previews, so this closes the last prompt on that path.
- The directory prompt is injected via `Deps.AskDir` so a test can assert the preview modes never reach it, with an interactive control proving the seam still fires on a normal run.

### D3 — `--output-format json` emits one document

`finishJSON` wrote the plan and then the result as **two concatenated top-level JSON documents**, so `json.loads` failed and every consumer needed a custom splitter.

- `finishJSON` now emits **exactly one document** via `display.RenderSyncJSON`: the plan at the top level (**shape unchanged** — `uploads[]`, `downloads[]`, `deletes[]`, `conflicts[]`, `stats`, `locked`) with the executed result nested under a new `"result"` key, present only when an `Execute` actually ran.
- `RenderPlanJSON`/`RenderResultJSON` are folded into `planJSON`/`resultJSON` builders behind it.

The top-level plan shape is preserved, so the reporter's documented workaround (reading `conflicts[]`/`downloads[]` at the root) keeps working — it just no longer needs a splitter.

## OUT OF SCOPE

- **D1** (data loss: bidirectional, remote-wins, no `--push-only`) — tracked separately; it is an engine change (the plan's `Downloads`/`REMOTE_DELETED`/`Conflicts` all write the working tree, and `Downloads` do so with no prompt and no `.LOCAL` backup) with flag/opt-in decisions still open.

## TESTING

- Preview modes don't prompt for the directory (`--dry-run`, `--diff`), with an interactive control that they otherwise do.
- JSON output is one `Unmarshal`-able document with the result nested under `"result"`; `assertOnlyJSON` now requires **exactly one** document; display-level `RenderSyncJSON` shape tests.
- `task lint` clean on all GOOS; `go test -race` green.

---

> [!NOTE]
> **Low Risk**
> Scoped CLI UX and JSON stdout contract changes in `artifact code sync`; sync execution and auth paths are unchanged, with added tests.
> 
> **Overview**
> Fixes **RAPTOR-19348** for `dr artifact code sync`: preview runs no longer block on input, and JSON mode stdout is a single parseable object.
> 
> **`--dry-run` / `--diff`** are treated as **preview** (`runFlags.Preview()`): directory resolution uses the same non-interactive path as `--yes`, so unattended previews default the project dir without `Project directory [.]:`. `Deps.AskDir` is injectable for tests; help text notes previews never prompt.
> 
> **`--output-format json`** now writes **one** document via `display.RenderSyncJSON` instead of concatenating plan + result. Top-level plan fields are unchanged; when `Execute` runs, outcome lives under **`result`**. Plan-only paths omit `result`.
> 
> Tests cover no directory prompt in preview modes, interactive control, exactly-one-JSON assertions, and `RenderSyncJSON` shape.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 7bb85b84548f68cfc67b0e20d0c5acd00d923680. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/cmd/artifact/code/codesync/cmd.go
+++ b/cmd/artifact/code/codesync/cmd.go
@@ -313,6 +313,13 @@ func finishJSON(engine engineRunner, plan *sync.SyncPlan, out io.Writer, flags r
 		return display.RenderSyncJSON(out, plan, nil, locked)
 	}
 
+	// The document is written only after Execute succeeds, so a failed sync
+	// leaves stdout empty and reports the error on stderr with a non-zero exit.
+	// This differs from the old two-document output, which emitted the plan
+	// before Execute and so left a plan on stdout even when the sync failed:
+	// with a single document, stdout carries a result only when there is one,
+	// and a consumer keys off the exit status rather than parsing a document
+	// that describes what was attempted rather than what happened.
 	result, err := engine.Execute(plan)
 	if err != nil {
 		return err
--- a/cmd/artifact/code/codesync/cmd.go
+++ b/cmd/artifact/code/codesync/cmd.go
@@ -67,6 +67,10 @@ func (r realEngine) Fetcher() display.ContentFetcher { return r.Engine.Fetcher()
 type Deps struct {
 	NewEngine func(dir string, opts sync.Options) (engineRunner, error)
 	ReadLine  func() (string, error)
+	// AskDir prompts for the project directory when neither --dir nor a
+	// non-interactive mode supplies one. A seam so tests can assert the
+	// preview modes never reach it.
+	AskDir func(label, defaultVal string) (string, error)
 }
 
 // runFlags is the parsed view of the boolean flags that gate
@@ -76,6 +80,14 @@ type runFlags struct {
 	DryRun bool
 	Diff   bool
 	Yes    bool
+}
+
+// Preview reports whether this run only shows a plan and writes nothing, on
+// either the local or the remote side. Such a run must never block on a
+// prompt: an operation that by definition changes nothing has no confirmation
+// to ask for (RAPTOR-19348).
+func (f runFlags) Preview() bool {
+	return f.DryRun || f.Diff
 }
 
 func defaultDeps() Deps {
@@ -89,6 +101,7 @@ func defaultDeps() Deps {
 			return realEngine{e}, nil
 		},
 		ReadLine: reader.ReadString,
+		AskDir:   dirprompt.AskWithDefault,
 	}
 }
 
@@ -118,8 +131,9 @@ versioned step.
 
 Use --dry-run to preview the plan without writing anything; --diff to
 also print per-file unified diffs. Both modes exit before any remote
-write. --yes auto-confirms the post-plan prompt and skips any
-interactive directory prompt.
+write and never prompt, so they are safe to run unattended. --yes
+auto-confirms the post-plan prompt and skips any interactive directory
+prompt.
 
 Run 'dr artifact code init <artifact-id>' first to link a project
 directory to an artifact.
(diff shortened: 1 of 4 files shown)


cmd/artifact/code/codesync/cmd.go
```diff
@@ -275,30 +296,27 @@ func shouldPromptConflicts(plan *sync.SyncPlan, yes bool) bool {
...
+		return display.RenderSyncJSON(out, plan, nil, locked)
 	}
 
 	result, err := engine.Execute(plan)
 	if err != nil {
 		return err
 	}
 
-	return display.RenderResultJSON(out, result)
+	return display.RenderSyncJSON(out, plan, result, locked)
```
wojtekwdr, line 328: [low] question: was dropping the plan on a failed Execute deliberate? main rendered it before calling Execute, so a JSON run that failed mid-sync still left the plan on stdout; now stdout is empty (I checked against both trees: 322 bytes there, 0 here). Empty plus a non-zero exit may well be the cleaner contract, it just isn't in the description.

internal/workload/sync/display/json.go
```diff
@@ -58,16 +58,27 @@ type PlanStatsJSON struct {
...
+// emits. The plan's fields sit at the top level (its shape unchanged), and the
+// result of a version-writing run is nested under "result" when one ran.
+//
+// One document per invocation: emitting the plan and the result as two
+// concatenated top-level objects made stdout unparseable by a plain json.loads
+// and forced every consumer to write a splitter (RAPTOR-19348).
+type SyncJSON struct {
+	PlanJSON
 
+	Result *ResultJSON `json:"result,omitempty"`
```
wojtekwdr, line 77: [low] non-blocking: `dr workload up` does this the other way round, scalars at the top level with the plan nested under `"plan"` (cmd/workload/up/cmd.go). Two opposite envelope shapes in one CLI is worth knowing about, even though this one has to stay as it is to keep the root `conflicts[]`/`downloads[]` the reporter relies on.

wojtekwdr (approved):
LGTM

cmd/artifact/code/codesync/cmd.go
adamalpi: Deliberate — and you called the intended contract exactly. On a failed Execute, stdout stays empty and the error goes to stderr with a non-zero exit, so stdout carries a document only when there is a result to report and a consumer keys off the exit status. Made it self-documenting in `finishJSON` and added a regression test (`TestRunE_JSONOutput_ExecuteError_LeavesStdoutEmpty`) so it stays that way. Thanks for the byte-level check.

internal/workload/sync/display/json.go
adamalpi: Good catch on the inconsistency — you're right that it has to stay this way here (root `uploads[]`/`downloads[]`/`conflicts[]` are what the reporter's consumers read). Added a note to the `SyncJSON` doc comment flagging the deliberate difference from `dr workload up`, so the next person sees the two shapes are intentional rather than an oversight.

PR #873, opened 2026-08-31
[RAPTOR-19721] feat(workload): import the .env variables the manifest is missing

wojtekwdr:
# RATIONALE

`.env` is read exactly once, when setup writes `.datarobot.yaml`. A variable added after that never reaches the container and nothing says so, so the only route is a hand edit of the manifest plus a hand-made credential for anything secret. This is the first of three PRs that close that gap; it adds the names, and leaves values alone.

## CHANGES

- `dr workload config --import-env` adds the `.env` variables the manifest does not declare yet. Additive and one-directional: a declared name keeps its value, a name dropped from `.env` is left alone. Secrets are stored as credentials the way setup stores them.
- The manifest is edited in place through the node tree, so comments, key order and hand-tuned keys survive.
- The edit refuses, before minting anything, a manifest with nowhere to put the result: an artifact named by id, a spec shared through an anchor, alias or merge key, an `environmentVars` that is not a list, or more than one YAML document. The refusal is the edit itself, rendered and thrown away, so it cannot disagree with the write it guards.
- A plain `dr workload config` over an existing manifest now names the variables `.env` defines and the file does not. Names only, never values.
- `--output-format json` gains `envKeysListed` and `envSecretsPending`, so a headless run can tell a manifest that cannot deploy yet from one that can.

## TESTING

`task test` and `task lint` pass. Exercised against a live instance: setup with a secret mints the credential and writes the reference, a variable added afterwards is named by a plain run and added by `--import-env`, and the deploy carries it into the container. Hostile values (`on`, `12:30`, `[a,b]`, `he said "hi"`, emoji, a 100KB value) round-trip byte-exact through both the create and the import paths, and stay strings under a YAML 1.1 reader.

## NOTES

First of a stack of three. The second adds `--update-env`, the third takes both flags to `dr workload up`.


wojtekwdr:
Reopening from the upstream branch so the stack sits in this repo with the right bases.

PR #876, opened 2026-08-31
[RAPTOR-19721] feat(workload): take the .env flags on up as well as config

wojtekwdr:
# RATIONALE

`dr workload up` already reads `.env`: with no manifest it runs the wizard, which classifies the file, mints the credentials and writes them in. The second run was the only one that could not, so the file a deploy reads was something you had to leave the command to change.

## CHANGES

- `dr workload up --import-env` and `dr workload up --update-env` do the same two edits as part of a deploy, so adding a variable or rotating a key is one command rather than two.
- Both stay opt-in, which keeps a deploy a function of the committed repo: without them the run reads nothing but the manifest, so a fresh CI clone with no `.env` deploys exactly what a working copy deploys.
- The edit runs before the plan, so the plan and the deploy carry the file as edited. `--dry-run` previews it, writes nothing, and says the plan below is for the manifest as it stands.
- A deploy that only re-sent secrets replaces no container, so it says the workload still serves the old value and names the restart, instead of reporting it as up to date and leaving it there.
- The `up` envelope gains an `env` block: `keysAdded`, `valuesUpdated`, `secretsRotated`, `secretsNotRotated`, `secretsPending`.

## TESTING

`task test` and `task lint` pass. Exercised against a live instance: `up --import-env` minted the credential, added both variables, planned them as spec changes that keep the running image, rolled out with no rebuild in about 90 seconds, and the container logs show both. `up --update-env` with nothing else to do re-sends the secret and prints the restart advice rather than a bare "Already up to date".

## NOTES

Third of a stack of three. Merge after the `--update-env` PR.

---

> [!NOTE]
> **Medium Risk**
> Opt-in flags alter the deploy load path (manifest edits and credential re-sends before planning), but logic is delegated to the existing wizard rather than new platform behavior.
> 
> **Overview**
> **`dr workload up`** now accepts **`--import-env`** and **`--update-env`**, mirroring **`dr workload config`**, so syncing `.env` into the manifest and deploying can be a single command.
> 
> The deploy path runs those edits **before** planning via the existing wizard (manifest directory, not the caller’s cwd). **`--dry-run`** previews env changes without writing the manifest; the plan still reflects the file on disk. Default **`up`** behavior is unchanged—no `.env` read unless a flag is passed.
> 
> JSON output adds an **`env`** object (`keysAdded`, `valuesUpdated`, `secretsRotated`, etc.). When a run only re-sends secrets and the deploy is otherwise unchanged, stderr warns that containers still serve the old values and prints **`stop`/`start`** with **`--dir`**.
> 
> **`internal/workload/up`** carries **`EnvEdit`** on **`Loaded`**/**`Result`**; **`load`** calls **`editEnv`** after a manifest is found. Tests cover import-before-read, rotation counts, no wizard without flags, and dry-run preview.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit faad349d0713c161974611ec3398ac304e45552d. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


internal/workload/up/run.go
```diff
@@ -508,6 +545,115 @@ func load(dir string, opts Options) (Loaded, error) {
...
+
+// reportEnvEdit says what the flags did to the manifest, before the plan that
+// was computed from the result of it.
+//
+// The wizard has already reported what it sent to the credential store, which
+// is the half that happens off this machine. What is left is the half that
+// happened to the file, and on a dry run the fact that the plan below knows
+// nothing about it: nothing was written, so the plan describes the manifest as
+// it stands rather than as the flags would leave it.
+func (o Options) reportEnvEdit(edit wizard.Result) {
```
chasdr, line 619: hey, small one. config --import-env prints the "Values written in the clear: X. Anything secret among them belongs in a credential" line when it writes a plaintext literal into the committed manifest (warnLiterals in the config cmd), but up --import-env / --update-env writing the same value only prints the counts here. same act, same file headed for git, so the disclosure lands on one path and not the other.

intentional, or want up to carry it too?

chasdr (approved):
lgtm 👍 one small question inline on the up-path disclosure, non-blocking.

internal/workload/up/run.go
```diff
@@ -138,6 +151,22 @@ type Result struct {
...
+	Env EnvEdit
+}
+
+// EnvEdit is the .env re-entry's side of a deploy.
+type EnvEdit struct {
+	KeysAdded      int
+	ValuesUpdated  int
+	SecretsRotated int
+	SecretsFailed  int
+	SecretsPending int
```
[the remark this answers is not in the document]
ajalon1: Confirmed — `load()` re-sends the secrets, then the `lookSettled`/`codeChangeFn`/`Build` returns drop `Env`, so a run that wrote to the credential store fails without reporting it and the retry of a bare `up` reports `Already up to date` while the old value keeps serving. Carrying `Env: loaded.Env` on those returns closes it.

ajalon1 (approved):
LGTM pending three small fixes: carrying `Env` on the early-failure returns (cursorbot's thread), moving `warnSecretStillServing` above the JSON return in `render()` so the restart advice also prints under `--output-format json`, and restating `doc.go`'s `never reads .env` paragraph the ticket asked for.

cmd/workload/up/cmd.go
```diff
@@ -411,9 +475,18 @@ func render(cmd *cobra.Command, f flags, format outputformat.OutputFormat, resul
...
+				KeysAdded:      result.Env.KeysAdded,
+				ValuesUpdated:  result.Env.ValuesUpdated,
+				SecretsRotated: result.Env.SecretsRotated,
+				SecretsFailed:  result.Env.SecretsFailed,
+				SecretsPending: result.Env.SecretsPending,
+			},
 		})
 	}
 
+	warnSecretStillServing(cmd.ErrOrStderr(), result, f)
```
ajalon1: This call never runs under `--output-format json`: `render()` returns after the envelope, so a secrets-only rotation prints `✓ Already up to date` on stderr and never the restart advice, while the workload keeps serving the old value. Moving it above the JSON return closes it — it writes only to stderr, so stdout purity holds.

Until fixed, don't demo `up --update-env --output-format json` :)

internal/workload/up/run.go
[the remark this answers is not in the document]
wojtekwdr: Confirmed and fixed. The three returns between `load` and the plan now share one result value carrying the workload id and the `.env` counts, so a rotation that already reached the credential store survives the failure instead of being dropped, and the retry of a bare `up` no longer calls the workload up to date while the old value is still serving.

internal/workload/up/run.go
wojtekwdr: Not intentional, an oversight. `up` carries it now.

Rather than copy the line, I pulled `literalNames`/`warnLiterals` out of the config command into the wizard package, so there is one implementation and the two commands cannot drift apart again. The `up` envelope also gained `env.literals` to match config's `envLiterals`, for the same reason: same edit, same committed file, so the JSON consumer should not have to care which command made it either.

cmd/workload/up/cmd.go
wojtekwdr: Moved above the JSON return, so both formats print it. It only writes to stderr, so stdout is still the envelope and nothing else. Added a test for each format so they are known to agree rather than assumed to.

PR #878, opened 2026-08-31
[RAPTOR-19348] fix(sync): back up local files before remote wins, refuse unattended overwrite

adamalpi:
# RATIONALE

D1 — the data-loss headline — of [RAPTOR-19348](https://datarobot.atlassian.net/browse/RAPTOR-19348).

## The problem, in plain English

`dr artifact code sync` sounds like it does one thing — **"push my code up to the platform."** It actually syncs **both directions**: it also pulls the platform's files *down* onto your folder, and when your file and the platform's disagree, **the platform wins** and overwrites yours.

Three things made that dangerous:

1. **You never asked for downloads.** The command reads as "push," so people run it expecting to *send* code, not *receive* it.
2. **It didn't warn you.** For a file that was simply "different on the platform," there was **no prompt** — it just silently overwrote your copy.
3. **There was often no undo.** A `.LOCAL` backup of your version was kept only when *both* sides had changed the same file. For the silent-overwrite case there was **no backup at all** — your version was just gone.

One person lost a working copy this way (they only recovered because the files happened to be committed in git).

## The fix

**Never destroy local work without a backup — and only interrupt the user for the case that can actually lose work.**

- Before the platform replaces or deletes *any* local file, your version is saved as `<file>.LOCAL.<timestamp>` — so there's always a recoverable copy.
- When a file changed **both locally and on the remote (a conflict)** — the only case with unsaved local work at stake — the sync **asks first** (default: No), or run non-interactively (`--yes`/CI) it **refuses** unless you pass `--accept-remote`, so an automated sync fails loudly instead of overwriting your changes.
- A **plain pull** of a remote change you hadn't touched (byte-identical to your last sync) just applies — still backed up to `.LOCAL`, but no prompt — so routine "pull a teammate's change" isn't interrupted and CI doesn't reflexively paste `--accept-remote`.

### Engine — never destroy local work without a recoverable copy
- `executePlan` runs **`backupOverwrittenLocals`** first: every local file a remote-side action would replace or delete — `REMOTE_MODIFIED` downloads, `REMOTE_DELETED` deletes, and conflicts — is renamed to `<path>.LOCAL.<stamp>` **before** the action runs.
- `applyDownloads` **`rb.Backup`s every pull target** before writing, so a `REMOTE_ADDED` path that unexpectedly holds a file (a `.drignore`d file, a symlink the walk skipped) is restored on a failed sync instead of deleted.
- `SyncPlan.OverwrittenLocalPaths` names the backup set; the ignore matcher excludes `*.LOCAL.<stamp>` (`ignore.IsBackupCopy`) so a backup is never re-uploaded.
- `Engine.conflictCopies` → `localBackups`; `Result.ConflictCopies` now carries every backup (JSON key unchanged).

### Command — ask/refuse on conflicts, not on fast-forward pulls
- Gate fires on **`HasConflicts()`** only. Interactive → prompt (names the files, default No). Non-interactive → refuse unless `--accept-remote`. Fast-forward pulls run unattended (still backed up).
- **JSON mode**: a conflicting plan without `--yes` emits `"refused": true` (no `result`), so a consumer sees nothing was applied — uploads included; with `--yes` but no `--accept-remote` it refuses loudly; one document throughout. A failed `Execute` leaves stdout empty (error to stderr, non-zero exit).

## OUT OF SCOPE

- **`--push-only`** (never pull at all) — deferred follow-up.

## TESTING

- Engine: full-set backups; rollback restores a pre-existing `REMOTE_ADDED` file; path-safety guards; ignore exclusion.
- Command: conflict prompt/refuse + `--accept-remote`; fast-forward pull runs without prompting; JSON `refused` flag; JSON `Execute`-error leaves stdout empty.
- `task lint` clean on all GOOS; `go test -race` green.


internal/workload/sync/phase5_execute.go
```diff
@@ -101,105 +102,95 @@ func executePlan(e *Engine, rb *Rollback) error {
...
+//
+// The rename frees the original path: applyDownloads then writes the remote
+// bytes there for downloads and conflicts, while REMOTE_DELETED has no download
+// and so simply stays gone, which is what the remote deletion asked for.
+func backupOverwrittenLocals(e *Engine, rb *Rollback) error {
+	stamp := e.nowFn().UTC().Format(backupStampFormat)
+
+	// OverwrittenLocalPaths is exactly the set of local files a remote-side
+	// action would replace or delete, so backing up that set is the whole job.
+	for _, path := range e.plan.OverwrittenLocalPaths() {
```
wojtekwdr, line 125: [high] `applyDownloads` still calls `rb.TrackCreated` on every pull, but this loop only backs up `ActDownloadModify`, so an `ActDownloadAdd` path is now tracked as created with no `rb.Backup` behind it. If a real file exists there and the sync fails, `Restore` deletes it and has nothing to put back; on main `backupDownloadTargets` backed up every `plan.Downloads` entry and the file came back. Reachable whenever something on disk is missing from the local manifest: a `.drignore`d path a teammate uploaded, or a symlink, since the walk skips those. Keeping an unconditional `rb.Backup` over the `pullList` targets alongside the new `.LOCAL` logic would cover it.

internal/workload/sync/plan.go
```diff
@@ -113,3 +113,41 @@ func (p *SyncPlan) ConflictPaths() []string {
...
+// OverwrittenLocalPaths lists, sorted, the local files applying this plan would
+// replace or delete: REMOTE_MODIFIED downloads (remote bytes land over yours),
+// REMOTE_DELETED deletes (your file is removed), and every conflict (remote
+// wins). REMOTE_ADDED and EDIT_DEL are excluded — they have no local file to
+// lose.
+func (p *SyncPlan) OverwrittenLocalPaths() []string {
+	var out []string
+
+	for _, fa := range p.Downloads {
+		if fa.Action == ActDownloadModify {
```
wojtekwdr, line 131: [medium] `ClsRemoteModified` is only reached when `localHash == baseHash`, so an `ActDownloadModify` file is byte-identical to the last sync and has no unsaved work to lose. Counting it here means a plain fast-forward pull of a teammate's change blocks on the prompt, and `dr artifact code sync --yes` in CI now exits non-zero on a merge that used to be a no-risk pull. My worry is that everyone pastes `--accept-remote` into their CI config within a week, and then the gate protects the conflict case zero percent of the time. Gating on `Conflicts` while keeping the backups on the full set would keep those two decisions separate.

cmd/artifact/code/codesync/cmd.go
```diff
@@ -269,36 +334,89 @@ func renderHumanPlan(cmd *cobra.Command, engine engineRunner, plan *sync.SyncPla
...
+		return display.RenderSyncJSON(out, plan, nil, locked)
+	}
+
+	if plan.OverwritesLocal() {
+		// Without --yes there is no confirmation channel in JSON mode, so the
+		// plan is emitted and nothing runs: a script inspects it and re-invokes
+		// with --yes --accept-remote to proceed. With --yes but no opt-in, the
+		// run is refused loudly (non-zero exit) rather than silently rewriting
+		// local files (RAPTOR-19348).
+		if !flags.Yes {
```
wojtekwdr, line 411: [medium] This branch used to be `!yes && plan.HasConflicts()`; now any `REMOTE_MODIFIED` download takes it, and the plan's uploads don't run either. `cli.IsNonInteractive` is env var and flag only with no TTY check, so a pipeline running `--output-format json` without `--yes` exits 0 having pushed nothing, and the absent `result` key is the only signal it didn't. The `--yes` branch just below fails loudly for the same plan. Could this one carry a `refused` field, or at least a line on stderr?

wojtekwdr (changes requested):

internal/workload/sync/phase5_execute.go
adamalpi: Fixed — good catch, this was a real regression. `applyDownloads` now `rb.Backup`s every pull target before writing (a no-op for the already-renamed REMOTE_MODIFIED/conflict paths, but it captures a REMOTE_ADDED path that unexpectedly holds a file). Added `TestApplyDownloads_RollbackRestoresAPreexistingRemoteAddedFile`: a `.drignore`d file under a REMOTE_ADDED path is now restored on `rb.Restore()` instead of deleted.

internal/workload/sync/plan.go
adamalpi: Agreed, and adopted. The gate now fires on `Conflicts` only; REMOTE_MODIFIED/REMOTE_DELETED are byte-identical to the last sync (I confirmed via `classifyBothPresentWithBase`: `!localChanged && remoteChanged`), so they apply without prompting — but they still get a `.LOCAL` backup, so the "never lose work without a copy" guarantee holds even if the baseline is stale. Backups stay on the full set, gate on conflicts. This keeps routine pulls unblocked and the gate meaningful for the case that matters.

cmd/artifact/code/codesync/cmd.go
adamalpi: Added a `refused` field. A conflicting plan under `--output-format json` without `--yes` now emits `"refused": true` (no `result`), so a consumer sees nothing was applied — uploads included — rather than inferring it from an absent key. Went with the machine-readable field over a stderr line since JSON mode is for scripts. (With the conflict-only gate from your other comment, a plain REMOTE_MODIFIED no longer hits this branch at all — it just runs.)

wojtekwdr (approved):

PR #882, opened 2026-09-01
[RAPTOR-19939] fix(workload): report .env drift on up, and after an update

wojtekwdr:
# RATIONALE

`dr workload up` never read `.env` unless you passed `--import-env` or `--update-env`, so a deploy whose `.env` had just been edited answered `✓ Already up to date`. That is a claim about being in sync, made without looking at the file the reader had just edited, and the only command that would have told them is `dr workload config`, which nothing on the deploy path points at.

The rule the two flags were built around does not change here: a deploy stays a function of the committed repo. Reading the file to say the two have parted is not deploying from it, and nothing is applied without a flag that asks for it by name.

## CHANGES

`up` reports the drift, through the comparison `dr workload config` already runs, so the two commands cannot reach different conclusions about the same pair of files. Names only, never values, matching what the deploy plan already redacts.

The remedy is a flag on the command you are already running rather than a trip to another one, and it carries the `--dir` this run was given, because `config` looks only where it is pointed while `up` walks upward for the manifest.

Only what has moved. A credential's value, which nothing can compare; a name the classifier read as local-only; an entry still naming the placeholder; a manifest whose shape no flag can edit. None of those can be settled by anything the reader is about to run, so on a deploy each would print on every run for the life of the project, and a line that always prints is one you stop reading along with the drift beside it. `dr workload config` still says all of it.

`--update-env` also owed a notice it was not giving. It named the variables the manifest does not declare only on the run that found nothing to do, so reconciling every value you could left the file reading as a clean bill of health while a variable it had never carried went on reaching nothing.

Silent where there is no `.env`, which is the ordinary CI case, and silent where the two files agree.

---

> [!NOTE]
> **Medium Risk**
> Changes the default `dr workload up` load path to read `.env` for drift messaging and shares wizard reporting between deploy and config; behavior stays non-mutating without flags but affects a core deploy workflow.
> 
> **Overview**
> **`dr workload up` now warns when `.env` and the manifest disagree**, instead of printing “already up to date” without reading `.env`. The comparison reuses the wizard’s drift logic; **nothing is written or applied** unless `--import-env` / `--update-env` are passed.
> 
> Deploy notices are **narrower than `dr workload config`**: only actionable drift (missing names, changed literals) with remedies pointing at **`dr workload up … --import-env` / `--update-env`**, including the run’s **`--dir`** (with **shell-safe quoting** for paths with spaces). Standing issues (secrets, local-only vars, placeholders, uneditable manifest shapes) stay on `config` so deploy output does not repeat every run.
> 
> **`--update-env` on `up` and `config`** now still lists manifest-undeclared `.env` variables **after** a successful value sync. Wizard gains **`Remedy`**, **`DriftOnly`**, **`ReportEnvDrift`**, and a lightweight **`detectEnv`** read; deploy indents drift prose via a new **`indent` writer**.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 27867bee0a3a36a38d895de0e6e3b2a2fa07a6e3. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


internal/workload/manifest/manifest.go
```diff
@@ -168,10 +168,28 @@ func DirFlag(dir string) string {
 	// absolute otherwise, which is what a project on another branch of the tree
 	// gets rather than a ../../.. chain.
 	if rel, err := filepath.Rel(cwd, project); err == nil && rel != "" && !strings.HasPrefix(rel, "..") {
-		return " --dir " + rel
+		return " --dir " + shellArg(rel)
 	}
 
-	return " --dir " + project
+	return " --dir " + shellArg(project)
```
chasdr: hey, the windows test job is red on this (TestDirFlag_LeavesAnOrdinaryPathBare, and QuotesAPathWithASpace too). DirFlag runs filepath.Rel, which hands back `services\api` on windows, so the remedy carries a backslash path and shellArg single-quotes it. the tests assert `services/api`.

dr takes forward slashes on windows anyway (filepath.Abs normalizes them), so ToSlash before shellArg greens both without touching the tests:

```suggestion
		return " --dir " + shellArg(filepath.ToSlash(rel))
	}

	return " --dir " + shellArg(filepath.ToSlash(project))
```

wojtekwdr: yep, that's it. ToSlash on both returns, tests untouched. it is also the only spelling worth printing: cmd.exe takes it, and a backslash pasted into a posix shell is an escape rather than a separator.

damjandr (approved):

PR #885, opened 2026-09-02
#826 Added support for multiple site configurations

### LINKED ISSUE #826: Proposal: Add support for profiles
I use more than a few DataRobot installations. I'd like to add support for specifying which one should be used so that I'm not having to erase and start over, or manually manage what config or environment variables are in use. My thought would be to follow a similar example as the aws CLI. Something like $XDG_CONFIG/datarobot/profiles.yaml that looks something like:

```yaml
eu-mtsaas:
  endpoint: <endpoint>
  token: <token>
jp-mtsaas:
  endpoint: <endpoint>
  token: <token>
```
Or alternatively, add a:
```yaml
profiles:
  eu-mtsaas:
    endpoint: <endpoint>
    token: <token>
```

and then add support for `--profile eu-mtsaas` to commands both for `dr auth login` and any other account.

Additionally, it would accept an environment variable to be set for all subcommands ala `DATAROBOT_CLI_PROFILE=eu-mtsaas`

Thoughts? Preferences for one config versus many? Desire to support a `$XDG_CONFIG/datarobot/profiles.d/<profile_name.yml` file?

carsongee:
# RATIONALE
Closes #826 

### Demo

https://github.com/user-attachments/assets/20af35bc-4668-40cb-a889-7bd7905b001a

This pull request introduces a new `profile` command group to the DataRobot CLI, allowing users to inspect and manage multiple named credential profiles for working with different DataRobot installations. It also improves documentation and user guidance for multi-profile usage and enhances the CLI's flexibility for enterprise and multi-environment scenarios.

The most important changes are:

### New Profile Inspection Commands

* Added a new `auth profile` command group with `list` and `show` subcommands for read-only inspection of named profiles stored in `drconfig.yaml`. These commands allow users to view all profiles, see which is active, and inspect the resolved settings for each profile, including inherited values. (`cmd/auth/profile/cmd.go`, `cmd/auth/profile/list/cmd.go`, `cmd/auth/profile/show/cmd.go`) [[1]](diffhunk://#diff-6da584c269ad82a74d5f7b3503932845a494db6f9f0de607dbe4795a0a241170R1-R51) [[2]](diffhunk://#diff-84ea79e66fe98a1a34dbd0c6848fddbc197d71f04deb9b916434584a7b2a70acR1-R128) [[3]](diffhunk://#diff-862cfcb5edc9aeaf1178f4bba7818c57294a7f28fa01a87650e3e954a173e8b0R1-R217)
* Added comprehensive tests for the new profile commands, ensuring correct output in both table and JSON formats, proper marking of the active profile, and correct handling of inherited values and error cases. (`cmd/auth/profile/list/cmd_test.go`, `cmd/auth/profile/show/cmd_test.go`) [[1]](diffhunk://#diff-09ab5bdc1948c1131891d6f4e860b1ae77c0e195ae808ab68901e0d606d1a30cR1-R173) [[2]](diffhunk://#diff-b5e5548f6b5fc8b5d57083e1b88f070d2033496ec58107927a44844e6432dd9cR1-R171)

### CLI Command Integration and Documentation

* Registered the new `profile` command group in the main `auth` command, updated the help text to document multi-profile usage, and provided user guidance for selecting profiles via `--profile` or `DATAROBOT_CLI_PROFILE`. (`cmd/auth/cmd.go`) [[1]](diffhunk://#diff-a676fdd037696dd99accef043bfa72a4dcfef30ca19392a1d573e8a3aa25354bR22) [[2]](diffhunk://#diff-a676fdd037696dd99accef043bfa72a4dcfef30ca19392a1d573e8a3aa25354bR40-R54)
* Added the `--profile` flag to the CLI root command, enabling users to select a named profile for any command invocation. (`cmd/root_factory.go`)

### Usability and Consistency Improvements

* Updated the `auth login` and `auth set-url` commands to annotate that they can create a new profile if the specified name does not exist, improving internal consistency and future extensibility. (`cmd/auth/login/cmd.go`, `cmd/auth/seturl/cmd.go`) [[1]](diffhunk://#diff-a34b6e89a1242864e497c9df0f3368f0eab61157c449351e042f0d24f2fd0691R126-R129) [[2]](diffhunk://#diff-d5415c83f5d4996bc70b4a55891ac205367017ac9d149882d35584d1c963ca78R37-R40)
* Improved error messages and code comments for clarity and accuracy regarding credential evaluation and profile terminology. (`cmd/auth/check/cmd.go`)
* Minor: Fixed missing error import in `cmd/root_factory.go`.om the code and
the review may happen while you are asleep / otherwise not able to respond quickly.
-->

## PR Automation

> [!IMPORTANT]
>

---

> [!NOTE]
> **Medium Risk**
> Changes authentication config load/save and credential precedence across the CLI, though behavior is heavily tested and env-based auth still overrides profiles.
> 
> **Overview**
> Adds **named profiles** in `drconfig.yaml` so one config file can hold credentials for multiple DataRobot installations. Any command can select a profile with **`--profile`** or **`DATAROBOT_CLI_PROFILE`**; the flag is universal and forwarded to plugins.
> 
> Config loading **merges** the active profile’s `endpoint`, `token`, `ca-cert`, and `ssl_verify` over the default section (env vars and flags still win). Writes go under `profiles.<name>` when a profile is active; **`auth login`** and **`auth set-url`** can target a **new** profile name (credentials cleared first so the default instance isn’t reused). New read-only **`dr auth profile list`** and **`dr auth profile show`** expose profiles without printing tokens.
> 
> Docs and help text are updated for multi-install workflows; debug output redacts nested profile tokens.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 564bda96e8853e286d8b61dbca7d7af10341596a. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


ajalon1 (commented):
Nice feature, and the case-sensitivity fix in the last commit was a good instinct. Went deep on this one and found three real bugs worth fixing before merge (left as inline comments), plus two scope questions I'd like your call on.

**Three correctness bugs, see inline comments:**
- `show/cmd.go`'s `resolveDetail()` overwrites a profile's own endpoint with the default's endpoint whenever the profile hasn't set a token yet, without clearing the "own" flag either.
- A brand-new profile can never get `--ca-cert`/`--ssl_verify` persisted into its own section, because the "does this profile already own this key" check reads stale in-memory viper state instead of the file that was just written.
- Related: when a new profile is created, only `endpoint`/`token` get cleared before the write, not `ca-cert`/`ssl_verify`. A new profile can silently inherit the default's CA cert for its first connection.

**Two things I'd like your call on:**

1. **SDK/tooling compat.** pysdk, rsdk, and dr-agent-cli all read this same `drconfig.yaml`, and none of them know about `profiles:` or `DATAROBOT_CLI_PROFILE` (checked directly against dr-agent-cli's `dr_config.py`, which just does `config.get("endpoint")` / `config.get("token")` flat). Switching `--profile` has zero effect on any of them when run standalone. Worth at least a documented limitation, and ideally exporting `DATAROBOT_ENDPOINT`/`DATAROBOT_API_TOKEN` (rsdk wants `DATAROBOT_API_ENDPOINT`) so a shell session can hand these to child tools.

2. **Active-profile switching.** Visibility is already solid: `list` marks the active row and `show` says "(active)", both tested. But there's no persisted "current profile" pointer or `use` command, only the per-invocation flag/env var. gcloud, docker, and kubectl all ship that as core scope, not a follow-up, precisely because silently running against the wrong instance is the most common failure mode of this kind of feature. Since `ActiveProfile()` would need a real precedence-layer change to support it later, seems worth deciding now rather than by omission.

Two smaller gaps also worth a look (inline): `dr start` (and every other command whose `PreRunE` is `EnsureAuthenticatedE`) fails hard on an unknown `--profile` instead of reaching the auto-login flow, unlike a brand-new default profile; and a hand-edited `profiles.default` section is never rejected, so it can silently create a dead, unreachable duplicate row in `list`.

**Smaller things, non-blocking:**
- `LoadProfiles()`'s `errors.As(err, &viper.ConfigFileNotFoundError{})` can't actually match, since `SetConfigFile` bypasses the code path that produces that error. A deleted config file would surface a raw OS error instead of "no profiles configured."
- Profile-awareness now lives inside the generic `UpdateConfigFile` write path, which is why `auth.go` needs two ordered calls with an explanatory comment. An isolated wrapper around it would keep the shared primitive profile-agnostic.
- `list`'s inline token-status string duplicates `show`'s `tokenStatus()` helper, worth hoisting up to the parent package.
- `resolveDetail`'s four field-resolution blocks are near-identical hand-copied code. A small data-driven resolver would make the endpoint bug above structurally hard to reintroduce.
- Read/write asymmetry on profile-name casing (reads always lowercase, writes preserve existing mixed case) isn't documented or tested.

Happy to pair on any of these.

cmd/auth/profile/show/cmd.go
```diff
@@ -0,0 +1,218 @@
...
+
+	if !detail.EndpointOwn {
+		detail.Endpoint = def.Endpoint
+	}
+
+	detail.HasTokenOwn = isDefault || own.HasToken
+	detail.HasToken = own.HasToken
+
+	if !detail.HasTokenOwn {
+		detail.Endpoint = def.Endpoint
```
ajalon1, line 145: This overwrites `detail.Endpoint` with `def.Endpoint` whenever the token isn't the profile's own, even though the endpoint above was already correctly set to the profile's own value two lines up. `EndpointOwn` never gets reset either, so the output shows the default's endpoint labeled as if it were this profile's own.

Trigger: `dr --profile staging auth set-url <url>` (writes only endpoint), then `dr auth profile show staging` before running `auth login`. You'll see the default profile's endpoint, not staging's, with no "(inherited)" tag. A resolver that handles each field independently instead of four hand-copied blocks would make this class of bug structurally harder to write.

internal/config/write.go
```diff
@@ -152,26 +152,125 @@ func readYAMLNode(path string) (*yaml.Node, error) {
...
+
+	if activeProfile != "" {
+		owned, _ = profileSection(activeProfile)
+	}
+
+	for key := range PersistableKeys {
+		_, scoped := ProfileScopedKeys[key]
+		_, owns := owned[key]
+
+		if scoped && activeProfile != "" && !owns {
```
ajalon1: `owned` here comes from `profileSection(activeProfile)`, which reads process-global viper state, not the file `UpdateConfigFile` just wrote to disk moments ago in the prior call. For a brand-new profile, ca-cert/ssl_verify can never be considered "owned," so they get dropped from the bare sweep permanently.

Concretely: `dr --profile onprem --ca-cert /etc/ssl/onprem-ca.pem auth set-url <url>` (your own docs' example) never writes `profiles.onprem.ca-cert`, which contradicts the comment in auth.go claiming ca-cert persists across invocations. Might be worth resolving ownership from the freshly-written node instead of from viper here.

cmd/root_factory.go
```diff
@@ -663,11 +666,33 @@ func defaultConfigInitializer(cmd *cobra.Command) error {
...
 
-	return nil
+	// Commands that create a profile (auth login, auth set-url) are allowed
+	// to name one that doesn't exist yet. Clear any inherited credentials so
+	// the new profile doesn't silently authenticate against the default
+	// profile's instance; the profile's section is created on first write.
+	if cmd.Annotations[config.ProfileCreateAnnotationKey] != "" {
+		log.Debugf("profile %q does not exist yet; %s is expected to create it", unknownProfile.Name, cmd.CommandPath())
+		viperx.Set(config.DataRobotURL, "")
+		viperx.Set(config.DataRobotAPIKey, "")
```
ajalon1, line 690: This clears `DataRobotURL`/`DataRobotAPIKey` explicitly, but `ProfileScopedKeys` also has `ca-cert`/`ssl_verify`, and neither gets cleared here. A new profile can end up silently using the default profile's CA cert or ssl_verify setting for its first connection, which looks like exactly what this block is trying to prevent for the other two keys.

Might be cleanest to iterate `config.ProfileScopedKeys` here instead of naming two of its four members.

cmd/root_factory.go
```diff
@@ -663,11 +666,33 @@ func defaultConfigInitializer(cmd *cobra.Command) error {
...
+	if !errors.As(err, &unknownProfile) {
 		return fmt.Errorf("failed to read config file: %w", err)
 	}
 
-	return nil
+	// Commands that create a profile (auth login, auth set-url) are allowed
+	// to name one that doesn't exist yet. Clear any inherited credentials so
+	// the new profile doesn't silently authenticate against the default
+	// profile's instance; the profile's section is created on first write.
+	if cmd.Annotations[config.ProfileCreateAnnotationKey] != "" {
```
ajalon1, line 687: `ProfileCreateAnnotationKey` is only set on `auth login`/`auth set-url`. Every other command whose `PreRunE` is `EnsureAuthenticatedE` (`dr start` included) fails with "unknown profile" right here, before ever reaching the auto-login flow described two lines up.

Worth deciding if that's intentional (profiles really can only ever be created via login/set-url) or if `dr start` specifically should carry the same annotation, since it's the advertised first command for a new user.

internal/config/profile.go
```diff
@@ -0,0 +1,270 @@
...
+
+	// GetStringMap, not AllSettings()[profilesKey]: AllSettings silently
+	// drops a profile section that defines nothing of its own (an empty
+	// YAML mapping), which would make a bare `profiles.<name>: {}` section
+	// vanish from `dr auth profile list` while still resolving correctly
+	// via applyProfile (which uses the same GetStringMap access pattern).
+	rawProfiles := v.GetStringMap(profilesKey)
+
+	names := make([]string, 0, len(rawProfiles))
+	for name := range rawProfiles {
```
ajalon1: This enumerates every key under `profiles:` with no check against the reserved "default" name, even though `ValidateProfileName` rejects `--profile default` elsewhere. A hand-edited `profiles.default: {...}` section (which write.go is designed to tolerate for arbitrary user edits) would render as a second "default" row in `dr auth profile list`, permanently unreachable, with nothing ever flagging it.

internal/config/profile.go
```diff
@@ -0,0 +1,270 @@
...
+	if configFile == "" {
+		return ProfileInfo{}, nil, nil
+	}
+
+	v := viper.New()
+	v.SetConfigFile(configFile)
+	v.SetConfigType("yaml")
+
+	err := v.ReadInConfig()
+	if err != nil && errors.As(err, &viper.ConfigFileNotFoundError{}) {
```
ajalon1: `v.SetConfigFile(configFile)` two lines up bypasses viper's `findConfigFile()`, the only path that actually produces `ConfigFileNotFoundError`. So this check can't match in practice.

If the file gets deleted mid-session, `dr auth profile show`/`list` will surface a raw OS error instead of gracefully falling back to "no profiles configured" the way this is clearly meant to.

cmd/auth/profile/show/cmd.go
```diff
@@ -0,0 +1,218 @@
...
+	if d.Active {
+		title += " (active)"
+	}
+
+	fmt.Println(tui.SubTitleStyle.Render(title))
+
+	printField("endpoint", valueOrDash(d.Endpoint), d.EndpointOwn)
+	printField("token", tokenStatus(d.HasToken), d.HasTokenOwn)
+	printField("ca-cert", valueOrDash(d.CACert), d.CACertOwn)
+	printField("ssl_verify", sslVerifyStatus(d.SSLVerify), d.SSLVerifyOwn)
```
chasdr: does the CLI actually read ssl_verify anywhere? profile show prints it as a resolved TLS setting and the group help lists it as a profile knob, but setupTLS in cmd/tls.go only reads skip-certificate-check and ca-cert, and i can't find any reader for ssl_verify (no RegisterAlias either).

if it's inert, showing it here reads as "this gates the CLI's TLS" when the real knob is -k / --skip-certificate-check. worth either wiring it in or dropping it from the show output + ProfileScopedKeys and keeping it as a preserved passthrough?

carsongee: 🤦 good catch. Droppig it for now. I'll take a look at properly wiring it in a follow-up

carsongee:
It should be ready for another look!

cmd/root_factory.go
chasdr, line 690: non-blocking: this clears endpoint and token but not ca-cert. the write path is safe now (profileOwnsKey won't persist the default's cert into the new profile's section), so this is only the first login connection running TLS with the default profile's ca-cert. credentials-only on purpose, or iterate `config.ProfileScopedKeys` here so ca-cert clears too?

chasdr (approved):
👋 chasbot here. chas is buried in other things so i pulled the re-review shift. came in ready to nitpick and the fixes just hold, which is frankly rude of you.

re-reviewed at HEAD. everything from last round lands:

- resolveDetail no longer shows the default's endpoint wearing a profile's name. credOwn couples endpoint+token now, and there's a regression test on it. 👍
- new-profile `--ca-cert` actually persists into its own section. profileOwnsKey reads the freshly-written file instead of the startup viper snapshot, and the eu-mtsaas test covers it.
- hand-edited `profiles.default` stops showing up as a phantom second "default" row.
- LoadProfiles falls back clean on a deleted config (os.ErrNotExist).
- ssl_verify from my last pass: dropping it as a profile knob and keeping it a preserved passthrough is the right call. nothing in the CLI reads it anyway.

left one non-blocking question inline on the new-profile clear (ca-cert). the two scope calls @ajalon1 raised (other SDKs/plugins not reading `profiles:`, and no `use` / persisted active profile) are still yours and his to make, not blockers.

approving. your bugbot can stand down.

cmd/auth/profile/show/cmd.go
ajalon1: Fixed in ae059fa4 — resolveDetail now computes one atomic `credOwn` flag instead of the two-branch overwrite.

internal/config/write.go
ajalon1: Fixed in ae059fa4 — profileOwnsKey now reads ownership from the on-disk YAML node instead of the stale process-global viper.

internal/config/profile.go
ajalon1: Fixed in ae059fa4 — namedProfileKeys now skips a `profiles.default` section explicitly, with a debug log.

internal/config/profile.go
ajalon1: Fixed in ae059fa4 — replaced with `errors.Is(err, os.ErrNotExist)`, matching what SetConfigFile actually surfaces.

ajalon1 (approved):
LGTM — did an architecture pass on the profiles schema too; it's additive-safe going forward. Filed some non-blocking fast-follows (CFX-8191, CFX-8192, CFX-8193, CFX-8194) rather than block on them. Good to merge.

cmd/root_factory.go
ajalon1: CFX-8194

cmd/root_factory.go
ajalon1: CFX-8193

PR #888, opened 2026-09-02
[RAPTOR-19939] feat(workload): one --sync-env flag, with a table you approve

wojtekwdr:
# RATIONALE

Stacked on the drift notice below it.

`--import-env` and `--update-env` were two flags for one intent. The first added names the manifest lacked and never touched a value; the second moved values and never added a name. Reconciling a `.env` took two runs, and the reader had to work out which half applied to what.

## CHANGES

One flag reconciles both ways. A name the manifest does not declare is added, a literal whose value has moved is rewritten, and the credential behind a secret is re-sent.

Collapsing them makes a single word rewrite a committed file and overwrite values on a tenant-wide store, so the run prints what it would do to every variable, and why, and asks. The table names the action and the consequence per row, so a rewrite and a re-send are never confused for each other.

`--yes` is the only consent that counts. `DATAROBOT_CLI_NON_INTERACTIVE` suppresses wizards in continuous integration and is not agreement to overwrite a tenant secret, which is the line `dr workload delete` already draws, so a run with no terminal and no `--yes` refuses and names the flag. A machine-readable run refuses and says the format is why. A dry run prints the table and applies nothing. Refusing is a decision rather than a failure: nothing is written, nothing reaches the tenant, and the deploy carries on with the file as it stands.

A re-sent secret reaches the credential store and no file, so a run with nothing else to do still says the container will not read it until the workload restarts, and names the stop and start that would. Performing that restart is the fourth PR in this stack.

---

> [!NOTE]
> **Medium Risk**
> The change rewrites committed manifests and overwrites tenant credentials, now behind a new confirmation path, but mis-automation with `--yes` or credential re-send failures could still leave workloads serving stale secrets until restart.
> 
> **Overview**
> **`--import-env` and `--update-env` are replaced by a single `--sync-env`** on `dr workload config` and `dr workload up`. One run can add missing names, rewrite changed literals, and re-send secrets, with drift notices and JSON telemetry updated to point at the new flag.
> 
> **Before anything is written or sent**, the wizard prints a per-variable table (add / update / re-send / skip and why) and asks for confirmation. Shared **`envconfirm`** wiring means only explicit **`--yes`** counts as consent—`DATAROBOT_CLI_NON_INTERACTIVE` does not—and non-interactive runs without `--yes` error instead of silently updating the tenant credential store. Refusing the prompt is a no-op, not a failure; deploys can continue with the existing manifest.
> 
> **Deploy UX** drops the old stderr nudge to `stop`/`start` after secret rotation when the deploy actually changes the workload; empty plans can still note that re-sent secrets need a restart to reach containers. **`manifest.SyncEnvVars`** is added (with tests) as a fuller reconcile editor; the CLI path keeps the prior **additive** rule—variables dropped from `.env` are not removed from the manifest—and the preview table states that explicitly.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 5bc26106819fb69e39ef5dfaa4d21c149176212a. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


internal/workload/manifest/importenv.go
```diff
@@ -712,3 +712,254 @@ func rewritable(entry *yaml.Node, wanted map[string]EnvVar) (EnvVar, *yaml.Node,
...
+//
+// Anything shared through an anchor or a merge key is refused rather than
+// edited, the way every other edit in this file refuses it: rewriting or
+// dropping a node another container reads would change that container too, in
+// a place no diff of this one would show.
+//
+// Secrets arrive already stored: vars carries the credential id the caller
+// minted, and this only writes the reference. That split is what keeps a
+// credential from being created for a file this edit then refuses.
+func SyncEnvVars(path string, vars []EnvVar, preview bool) (SyncChanges, []byte, error) {
```
chasdr: SyncEnvVars and its test land here but nothing calls it until #889 (editEnv), and reconcileEnvVars inside it removes entries, which this PR's table says never happens. reads as a sync editor that deletes vars the command doesn't. move it down to #889?

cmd/workload/up/cmd.go
```diff
@@ -273,13 +239,11 @@ func addFlags(cmd *cobra.Command, f *flags, poll *pollflags.Set) {
...
-			"Secrets are stored as credentials, the same way setup stores them.")
-	cmd.Flags().BoolVar(&f.updateEnv, "update-env", false,
-		"Before deploying, bring the variables the manifest already declares back in line with .env: "+
-			"a literal is rewritten and the credential behind a secret is re-sent. A re-sent secret reaches "+
-			"the containers this deploy replaces; if the deploy has nothing else to do, it will not replace them.")
+	cmd.Flags().BoolVar(&f.syncEnv, "sync-env", false,
+		"Before deploying, bring the manifest into line with .env: add the variables it does not declare, "+
+			"rewrite a literal whose value has moved, and re-send the credential behind a secret. Prints what "+
+			"it would do and asks first. Nothing is removed. A re-sent secret reaches the containers this "+
+			"deploy replaces, and the deploy restarts the workload when it would otherwise replace none.")
```
chasdr: the --sync-env help says "the deploy restarts the workload when it would otherwise replace none", but this layer does no restart (commit says that's the next PR) and it dropped warnSecretStillServing, so a rotation-only up --sync-env prints "needs a restart" with no command while --help claims it auto-restarts. #889 closes it, but if this merges alone the help and behavior disagree. keep the stop/start line here, or move that clause of the help to #889?

internal/workload/manifest/importenv.go
wojtekwdr: agreed, moved down to #889 with its tests. nothing here references it now.

cmd/workload/up/cmd.go
wojtekwdr: did both, in a way: the restart clause is out of this PR's help and the stop/start line is back where it was, so a rotation-only run prints the two commands again here. #890 removes the line with the restart and puts the clause in the help there. the render comment that spoke of a restart below is fixed too.

cdevent (approved):

PR #889, opened 2026-09-02
[RAPTOR-19939] feat(workload): let .env win on existence too

wojtekwdr:
# RATIONALE

Stacked on the `--sync-env` flag below it.

The reconcile settled a value and a missing name but never a name you had deleted. A variable dropped from `.env` went on reaching the container, and the table said so by name on every run, which is a lot of ceremony for something it refused to do.

## CHANGES

`.env` wins on all three questions: the value, the form the entry takes, and whether the entry exists at all. A literal that now reads as a secret becomes a credential reference and the reverse becomes a literal, an entry an earlier run left on the placeholder is finished, a value the manifest kept as a mapping or a list becomes the string `.env` has for it, and a name `.env` no longer mentions is removed.

That reverses the additive rule the import followed, which is why the run shows its work and asks. Removals are rows in the same table, named before anything is written.

Underneath, `manifest.SyncEnvVars` replaces the import and update editors. The file is edited in place, so comments, entry order and hand-tuned keys survive, and the edit is rendered and thrown away once before any credential is created, so a file the write would refuse never leaves a credential behind that nothing points at. An entry carrying an anchor anywhere inside it is refused a removal or a change of form, the way the rewrite already refused it, since the alias reading through it would be left pointing at nothing.

The classifier now decides only what form an entry takes and no longer whether the variable travels. A stale local copy is therefore enough to delete a variable a running workload depends on, or to send a value the classifier once withheld, so a value that lands in the manifest in the clear is disclosed by name.

A project with no `.env` at all is left alone rather than emptied, because the ordinary CI checkout has none, and so is one whose `.env` declares nothing: a touch, or a secrets step that wrote an empty file, is not an instruction to delete the environment. A manifest sharing its environment through a YAML anchor takes no edit either: rewriting or dropping one entry would change it for every container reading the same node, so a run with only secrets to re-send goes through, after the same question every other re-send gets, and one with file work to do is refused before the question rather than half-applied.

The drift notice a bare deploy prints covers removals on the same terms, and `manifest.DirFlag` quotes a project path a shell would act on, in double quotes so it survives both the single quotes the notice wraps the command in and cmd.exe.

`--yes` is the only consent, as the flag PR says: the variable that suppresses wizards in CI is not agreement to remove a variable or overwrite a tenant credential, and each command has a test that a run with it set and no `--yes` is refused naming the flag.

---

> [!NOTE]
> **High Risk**
> Opt-in but irreversible when confirmed: stale `.env` can drop manifest variables workloads rely on, write formerly withheld values in the clear, and re-send secrets to the tenant-wide credential store.
> 
> **Overview**
> **`--sync-env` on `dr workload config` and `dr workload up` now treats `.env` as the full source of truth**, not just for adds and value updates. Manifest entries can be **removed** when a variable disappears from `.env`, **retyped** (literal ↔ credential reference), **finished** from credential placeholders, and **flattened** when YAML had mappings/lists. That replaces the old additive `ImportEnvVars` / `UpdateEnvVars` split with a single **`SyncEnvVars`** path wired through the wizard.
> 
> The **preview table includes remove rows**, and **JSON/text output expose `envNamesRemoved` / `namesRemoved`** so pipelines can tell adds from destructive edits. **Classifier “local-only” no longer blocks sync**—it only picks secret vs literal; values the classifier used to withhold can now land in the manifest (with literal disclosure). **Missing `.env` does not wipe the manifest**; **YAML-anchor shared environments** still block file edits but allow secret-only rotation.
> 
> **Confirmation is tightened**: non-interactive runs need **`--yes`** (CI non-interactive env is not consent for credential overwrites); JSON runs get a **silent refusal** without the table. Deploy **drift notices** now flag removals and other cases that `--sync-env` can fix. **`manifest.DirFlag`** uses safer shell quoting for suggested commands.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 68cc778832c481a792a7bab12f71fc422a5fdae2. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


cmd/workload/config/cmd.go
```diff
@@ -282,9 +287,13 @@ func run(cmd *cobra.Command, f flags, format outputformat.OutputFormat) error {
 		// suppresses wizards in CI is not consent to overwrite a value on the
 		// tenant, which is the line `dr workload delete` already draws.
 		Confirm: envconfirm.Ask(cmd.ErrOrStderr(), cmd.InOrStdin(), envconfirm.Policy{
-			Yes:         f.yes,
+			// The flag or the environment variable, matching what the rest of
+			// this command treats as an answer. What is not an answer is the
+			// mere absence of a terminal, which is the case this refuses.
+			Yes:         yes,
```
chasdr: hi, i think this flipped the sync-env consent gate, and i may be missing context so tell me if it was on purpose.

#889 changed `Yes: f.yes` to `Yes: yes` here and in up/cmd.go, where `yes := cli.IsNonInteractive(cmd)`. that feeds DATAROBOT_CLI_NON_INTERACTIVE into Policy.Yes, so the env var counts as consent now. the comment right above still says "f.yes rather than the merged signal", and envconfirm's doc says the var is "deliberately not read here".

i built the stack and ran it. with DATAROBOT_CLI_NON_INTERACTIVE=1 and no --yes, `config --sync-env` rewrote the manifest and removed a var, exit 0, no prompt. with an empty-but-present .env it stripped environmentVars to [] and said done. that's the CI default, so a scheduled `up --sync-env` after a failed .env step wipes the committed env and deploys it.

revert to f.yes on both?

```suggestion
			Yes:         f.yes,
```

(same line in up/cmd.go, and worth a test per entrypoint: --sync-env piped with no --yes should error naming --yes.)

internal/workload/wizard/run.go
```diff
@@ -743,328 +721,330 @@ func (o Options) warnEnvHeldBack(parsed *manifest.Manifest, detected Detected, b
...
-			local = append(local, v.Name)
-		}
+	// No file, no reconciliation. .env winning on every variable means an
+	// empty set of them would empty the manifest, so a project that simply
+	// has no .env, which is the ordinary CI checkout, would have its whole
+	// environment deleted by a flag that promised to bring the two into line.
+	// Asked of the filesystem rather than of the parse: a .env holding only
+	// comments is a file the user is looking at and can answer for, and the
+	// table shows them what emptying it would cost.
+	if !fsutil.FileExists(filepath.Join(filepath.Dir(path), EnvFileName)) {
```
chasdr: separate from the consent thing: the "no .env, leave it alone" guard keys on the file existing, not on it having vars. an empty-but-present .env parses to zero vars, so droppedNames returns every declared name and the plan removes them all. a `touch .env` or a secret step that wrote nothing empties the manifest. treat a present-but-empty .env the same as a missing one?

internal/workload/wizard/run.go
```diff
@@ -743,328 +721,330 @@ func (o Options) warnEnvHeldBack(parsed *manifest.Manifest, detected Detected, b
...
+	}
 
-		return
+	if !errors.Is(blocked, manifest.ErrSharedEnvVars) || o.hasFileWork(parsed, detected, wanted) {
+		return false, blocked
 	}
 
-	o.warnNamesHeldBack(parsed, detected)
-}
+	result.EnvSecretsRotated, result.EnvSecretsNotRotated = o.rotateSecrets(parsed, detected, wanted)
```
chasdr, line 853: one more, pre-existing (same shape in #888, so a follow-up not a blocker): an anchored/aliased env node gets an all-skip plan, so the confirm check returns before o.Confirm() runs, but this still calls rotateSecrets and PATCHes the credential. so a secrets-only --sync-env on an anchored manifest re-sends to the tenant with no prompt, even interactively. gate the rotate on consent even when the file plan is all-skip?

cmd/workload/config/cmd.go
wojtekwdr: not on purpose, thanks for running it. back to f.yes on both, and each command has a test now: variable set, no --yes, --sync-env is refused naming the flag and the file is untouched.

internal/workload/wizard/run.go
wojtekwdr: agreed. a present-but-empty .env is treated like a missing one now, for the sync and for the drift notice, and the run says which of the two it found.

internal/workload/wizard/run.go
wojtekwdr: fixed here rather than as a follow-up, it was small: the shared table keeps its re-send rows, so it stops and asks like any other, and a run with file work on a shared block is refused before the question instead of after it. on #888 the re-send row was already in the table, valueActions ran regardless of the block, so that one did ask; the all-skip plan came in with this PR.

damjandr (approved):

PR #890, opened 2026-09-02
[RAPTOR-19939] fix(workload): restart on rotation, settle the waits, recover a deleted catalog

wojtekwdr:
# RATIONALE

Stacked on the removals PR below it. Two halves: what a rotation does to the running workload, and what a deploy does when the build path fails under it. Both were found by running the rest of this stack against staging.

## CHANGES

**The rotation restart.** A rotation writes to the credential store and to no file, so a run with nothing else to do produced an empty plan, replaced no container, and left the workload serving the old value under a summary calling it up to date. It was, about the manifest. It was not about the thing taking requests. The deploy now restarts the workload itself, the rolling swap a resize gets, because the platform carries out a settings change by rolling the workload onto the artifact it already runs. The endpoint answers throughout, where the stop and start it used to recommend is a real outage. A restart refused before the settings call, by a workload in a state no deploy can act on or a swap already in flight, still says the secret was re-sent and is not being served yet, and a manifest that leaves the runtime block to the platform restarts with the block the workload already runs with.

**The two waits.** Both were wrong in opposite directions. The restart returned as soon as the replacement completed, which says the new generation came up and not that the old one stopped answering, and the old one is the generation still serving the key this run replaced. It now waits for the drain. The drain wait had the opposite fault: it waited for the outgoing generation to reach `stopped`. Draining is the old container finishing requests it already had, and on staging, with the new generation promoted and the old one draining, the endpoint answered from the new build every time. Waiting it out cost the platform's whole termination grace, seven minutes on a swap whose handover took thirty seconds. Where a role is reported, a draining predecessor no longer holds the wait. One still `running` does, because the router may not have moved, and where no role is reported draining still holds, because nothing else tells a predecessor from its replacement.

**A timed-out build poll.** One status request hitting the client timeout ended the deploy with `poll build: context deadline exceeded` while the platform's own log had already said `COMPLETED`. The image was built, no workload was created, the manifest was left unbound. The poll now forgives a few consecutive transient errors, reset by any successful one, which is the budget the workload and replacement polls already had.

**A deleted code catalog.** The sync decides what to upload by comparing the tree against its own record of what was pushed, and nothing asks whether the catalog that record names still exists. With it deleted on the platform, an unchanged tree uploaded nothing and the version the run had just minted was pointed at a catalog the platform called previously deleted, and a changed tree heard the same answer one request earlier, on the upload into that catalog. The run died there either way, with the project re-pointed at an artifact holding no code, and every later run failed the same way. The files are on disk, so wherever the run hears that answer it forgets the catalog, empties the sync engine's index and uploads into a fresh one. Emptying the index is the half easy to miss: with no catalog there is no remote, so the engine takes its index for the remote and every file reads as unchanged. A build found in an artifact's history is also recorded as the image's origin, or every roll after a recovery spends a build again and mints a version needing the very catalog a recovery may no longer have.

Why the platform deleted a catalog behind a live draft is not answered here.

---

> [!NOTE]
> **Medium Risk**
> Changes core `dr workload up` deploy, polling, secret rotation, and code-sync recovery paths that affect production rollouts; behavior is well covered by tests but mistakes could cause failed deploys, premature success, or duplicate uploads.
> 
> **Overview**
> **`dr workload up --sync-env`** now performs a rolling restart when re-sending secrets is the only change, instead of telling users to run stop/start manually. It uses the same settings-driven replacement as a resize, waits for the drain so the old generation is not still serving the prior credential, and updates the empty-plan verdict (including dry-run tense).
> 
> **Deploy reliability:** `WaitForBuild` tolerates a few consecutive transient poll failures (same policy as workload/replacement polls), avoiding aborted deploys when the build already completed. Proton serving logic stops blocking on a **draining** predecessor once an **active** generation is reported, cutting long post-roll waits while still waiting on a still-**running** predecessor or when no role is reported.
> 
> **Deleted code catalog recovery:** When inheriting code fails because the platform says the catalog was previously deleted (or project state is inconsistent with the sync index), the run clears catalog/built metadata, resets the local sync index, re-uploads from disk, and continues. Reusing an existing completed build from history now records **last built** like a run-triggered build, avoiding unnecessary rebuilds after recovery.
> 
> CHANGELOG documents these fixes; tests cover build poll forgiveness, proton drain behavior, rotation restart paths, and catalog recovery.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 6b6bb3b51e114c8f7a97fb77bbd09b47bbde333e. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


internal/workload/proton.go
```diff
@@ -230,12 +230,22 @@ func anyServingPredecessor(protons []Proton, wantArtifactID string) bool {
...
-		if IsDrainingProtonStatus(p.Status) {
+		// Draining holds the wait only when nothing else can say the
+		// handover happened. With a role reported, the active generation is
+		// the one the router sends to, and a draining predecessor is
+		// finishing the requests it already had: measured on staging with the
+		// old generation draining and the new one active, the endpoint
+		// answered from the new build every time. Waiting for it to reach
+		// stopped as well costs the platform's whole termination grace, which
+		// ran to seven minutes on a swap whose handover took thirty seconds.
+		if IsDrainingProtonStatus(p.Status) && activeAt < 0 {
```
chasdr: the header still says "Draining and running both count", but with the `&& activeAt < 0` you added here, a draining predecessor no longer counts once a role is reported. the two inline comments you added are spot on, just the top-of-function line reads stale now. reword it to match?

internal/workload/up/version.go
```diff
@@ -776,43 +779,188 @@ func relinkFailure(artifactID, projectDir string, err error) error {
...
+
+	switch httpErr.StatusCode {
+	case http.StatusGone:
+		return true
+	case http.StatusUnprocessableEntity, http.StatusNotFound:
+		// The detail when the client lifted one, the raw body when it did not:
+		// which of the two carries the sentence depends on the envelope.
+		said := strings.ToLower(httpErr.Detail + " " + string(httpErr.Body))
+
+		return strings.Contains(said, "previously deleted")
```
chasdr, line 960: the 410 branch is solid. for this 422/404 one, catalogGone leans on `strings.Contains(said, "previously deleted")`, and HTTPError.Detail's own doc warns it's display-only and can come back empty or localized. you read Body too so the empty case is covered, but if the platform ever rewords or localizes that sentence, recovery quietly stops firing and you're back to the dead-catalog death this fixes. is that phrase the only signal the platform gives here, or is there a structured code to key off?

chasdr (approved):
LGTM 👍 traced the paths that worried me and they hold: dry-run returns before restartForRotation so no accidental restart, the catalog recovery can't loop, and WaitForBuild returns the last good build on give-up. ran the affected packages (all green) and task lint (0 issues, all three GOOS). two non-blocking questions inline, nothing that gates the merge.

PR #896, opened 2026-09-07
[BUZZOK-31834] Decode versions in order to install key dependencies first

tsdaemon:
# RATIONALE
I want to add `pulumi-datarobot` to `versions.yaml` to unify all the dependencies, and move away from a brittle fixes in `infra/Taskfile.yaml`. But when I tested it on a clean system, `dr dependency install` failed because there was no `pulumi` because it does not respect the order of dependencies installation in the YAML (which is expected for a random map). This updates the decode to iterate over config as it is defined.

## PR Automation

> [!IMPORTANT]
>

---

> [!NOTE]
> **Low Risk**
> Localized change to YAML parsing in the tools package; install order becomes deterministic and matches the YAML, with a regression test and no auth or data-handling impact.
> 
> **Overview**
> **`GetRequirementsFromDir`** no longer unmarshals `versions.yaml` into a map, which made prerequisite order nondeterministic and could run `dr dependency install` steps before their dependencies (e.g. `pulumi-datarobot` before `pulumi`). It now decodes with **`yaml.Node`** and walks top-level keys in **file order**, with checks for an empty document and a top-level mapping.
> 
> A regression test runs **`GetRequirementsFromDir`** many times and asserts keys stay in declaration order (`tool-z`, `tool-a`, `tool-m`).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 6dce629c7a17423f13ec6ab0e983ab053fadf007. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/internal/tools/versions.go
+++ b/internal/tools/versions.go
@@ -65,9 +65,18 @@ func GetRequirementsFromDir(dir string) ([]Prerequisite, []string, error) {
 	var violations []string
 
 	versions := make([]Prerequisite, 0, len(mapping.Content)/2)
+	seen := make(map[string]bool, len(mapping.Content)/2)
 
 	for i := 0; i+1 < len(mapping.Content); i += 2 {
 		key := mapping.Content[i].Value
+
+		// yaml.Node traversal doesn't reject duplicate keys the way decoding into a
+		// map does, so check explicitly to keep that same fail-fast behavior.
+		if seen[key] {
+			return nil, nil, fmt.Errorf("versions yaml file %s: duplicate key %q (line %d)", yamlFile, key, mapping.Content[i].Line)
+		}
+
+		seen[key] = true
 
 		var version Prerequisite
 		if err := mapping.Content[i+1].Decode(&version); err != nil {
--- a/internal/tools/versions.go
+++ b/internal/tools/versions.go
@@ -24,8 +24,6 @@ import (
 	"gopkg.in/yaml.v3"
 )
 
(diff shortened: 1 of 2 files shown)


chasdr (approved):
lgtm 👍 traced the order-preservation through GetRequirements -> CheckPrerequisiteList into MissingTools, install order matches the file end to end, and the regression test defeats both sort and coincidence.

@ajalon1 @taras-pokornyy tagging you two for a second look, you both know the dependency install path better than i do. anything off about walking yaml.Node for order here?

taras-pokornyy (approved):

taras-pokornyy:
> lgtm 👍 traced the order-preservation through GetRequirements -> CheckPrerequisiteList into MissingTools, install order matches the file end to end, and the regression test defeats both sort and coincidence.
> 
> @ajalon1 @taras-pokornyy tagging you two for a second look, you both know the dependency install path better than i do. anything off about walking yaml.Node for order here?

Confirmed `Node.Decode` per entry uses the same decode path as before, so struct tags/validation/anchors within an entry are unaffected - only top-level ordering changed. Two harmless edge-case shifts: `null` content now errors instead of silently returning empty, and duplicate keys now produce two entries instead of last-wins. LGTM.

tsdaemon:
@taras-pokornyy good catch on duplicate key. Actually it was not last wins, Yaml schema v3 would fail on it when deconding into a map, and my implementation didn't. Added a fix and a test.

PR #897, opened 2026-09-08
[RAPTOR-20092] fix(workload): a rebuilt wizard draft no longer shows the suggested name as typed

wojtekwdr:
# RATIONALE
Follow-up from the review of #856. The wizard tracks whether the name on screen was typed or only suggested, and that flag survived the two places that rebuild the draft: picking a different directory on the way back, and choosing "set up a new workload" after having looked at an existing one. Either way the fresh draft's suggestion came back as a typed value, one Enter away from naming a deployed workload after its directory, which is what the name screen exists to prevent.

## CHANGES
The draft and the flag now move together through one small flow method, so the two cannot drift again. A changed directory is a restart, so everything answered for the old tree's draft goes with it, the typed name included. Re-choosing the same directory keeps it.

## TESTING
Run `dr workload config --dry-run` from a directory whose only project sits in a subdirectory, choose "Use this directory anyway", type a name, then Escape back to the directory question and pick the subdirectory. The name field must come back empty with its placeholder rather than showing the subdirectory's name as a value.

---

> [!NOTE]
> **Low Risk**
> Interactive wizard state only; no auth, persistence, or deploy-path changes beyond preventing accidental one-Enter name confirmation.
> 
> **Overview**
> Fixes a workload config wizard bug where **rebuilding the draft** (different directory or “create new workload” after binding) left `nameGiven` true, so the new tree’s **suggested name appeared as a typed value** and one Enter could deploy with a bad name (e.g. `src`).
> 
> Draft install and **`nameGiven` reset now go through `startFrom`**: only a **`--name` flag** still counts as user-given; interactive names typed for the old draft are cleared with the restart. **`acceptDirectory`** and **create-new-after-bind** call `startFrom` instead of assigning the draft alone. Comments spell out that changing directory discards post-directory answers including the typed name; re-selecting the same directory keeps them.
> 
> **Regression tests** cover directory switch vs stay-put and create-new after bind.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 8775ab4dbc2d91ed91eaf8eae38d4283b831e2a9. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


internal/workload/wizard/model.go
```diff
@@ -210,6 +207,18 @@ func newFlow(detected Detected, workloads []workload.Workload, answers Answers)
...
 
+// startFrom installs a draft built from this run's flags, at the start and
+// again whenever the flow starts over. The name question starts over with it:
+// a name typed for the draft being replaced went with that draft, so only a
+// --name still counts as given. That flag is an answer the user gave and the
+// screen shows it as a value; the suggestion a fresh draft carries is a
+// placeholder, and left marked as given it would come back looking typed,
+// one reflexive Enter away from naming a deployed workload "src".
+func (f *flow) startFrom(draft manifest.Draft) {
+	f.draft = draft
```
adamalpi, line 218: **[Low] The third draft rebuild is exempt from `startFrom`, and nothing says why**

(Anchored here; the site itself is `model.go:1145`, outside this PR's hunks.)

There is a third place the draft is rebuilt — `f.draft = f.answers.partialApplyTo(live.Defaults(), f.detected)` in the live-workload handler — and it deliberately does not go through `startFrom`, so `nameGiven` survives it.

That is correct today, but only because of an invariant two functions away: `branch()` returns `screenKind` when `f.live != nil`, so `screenName` is unreachable while bound, and a stale `nameGiven` can never be read — `screens.go:266` is its only consumer. Nothing at 1145, and nothing in this doc comment, records that the site is exempt or what makes it safe.

Add a rename screen for a bound workload, or otherwise make `screenName` reachable when `f.live != nil`, and this PR's bug comes back in the one place with no test covering it. A one-line note at 1145, or a sentence here naming the exempt site, is cheap insurance for a fix whose whole point is that draft rebuilds and `nameGiven` must not drift apart.

internal/workload/wizard/model.go
```diff
@@ -210,6 +207,18 @@ func newFlow(detected Detected, workloads []workload.Workload, answers Answers)
...
+	f.nameGiven = f.answers.Name != ""
```
adamalpi, line 219: **[Low] No CHANGELOG entry for a user-facing fix**

(Anchored here; `CHANGELOG.md` is not in this PR.)

The repo convention looks firm: 6 of the last 12 commits on `main` touch `CHANGELOG.md`, and the ones that don't are dependency or CI bumps. There is a close precedent for a UX-only fix getting an entry — `c62d47a fix(workload): stop a refused deploy announcing work it will not do` — and the sibling PR #898 adds one.

The user-visible behaviour here is worth a line: before this, one reflexive Enter after changing directory could put a name like `src` on a deployed workload, which is the outcome the name screen exists to prevent.

adamalpi (approved):
Traced every path that touches nameGiven; the invariant holds and the two new tests cover both reset sites with the right control case. Two non-blocking notes inline, both anchored here because neither has an exact line in the diff.

PR #898, opened 2026-09-08
[RAPTOR-20129] fix(filesapi): send zip-path overwrite in the multipart form

wojtekwdr:
# RATIONALE

`dr artifact code sync` and `dr workload up` upload through `POST /files/{id}/fromFile/` whenever a change set exceeds 20 files or 50 MB. The overwrite mode went only in the query string, which the Files API accepts and ignores, so the server's rename default applied: every existing path came back as `name (2).ext` holding the new bytes while the original kept the old ones, and an image built from that version ran stale code. Reproduced on staging: a 25-file re-sync produced a 53-file version with 25 duplicates. The Terraform provider carried the same code and fixed it the same way.

## CHANGES

- Multipart uploads can carry form fields, framed before the file part so a streaming parser has them before the file.
- `UploadFromZipExisting` sends `overwrite` in the form and keeps the query copy until the API documents which location is authoritative.
- Tests pin the field order, the query copy, and an exact Content-Length with fields present.
- Changelog entry under Unreleased / Fixed.

## TESTING

Staging, fixed binary, 25 modified files: `dr artifact code sync` re-sync gives 28 files and no duplicates (53 with 25 duplicates before), and a `dr workload up` re-deploy gives 27 files and no duplicates, with the server checksum of the edited file matching the local one. `task lint` on all three GOOS legs and `task test` are green.

## RELATED

PR #867 bundles the same fix with unrelated integrity work and is currently red; this is the standalone extract so it can ship on its own, and #867 can rebase onto it. `useArchiveContents` staying in the query is RAPTOR-19915.

---

> [!NOTE]
> **Medium Risk**
> Changes deploy/sync upload behavior for large change sets; incorrect framing could break uploads, but scope is limited to the zip fromFile path and is covered by new tests.
> 
> **Overview**
> Fixes **catalog zip uploads** (`dr artifact code sync` and large-change `dr workload up` paths) so existing files are **replaced** instead of duplicated. The Files API was ignoring `overwrite` when it was only on the query string; the server’s rename default produced `name (2).ext` copies and left old bytes in place, so built images could run stale code.
> 
> **`UploadFromZipExisting`** now sends `overwrite` as a multipart form field **before** the zip part (query param kept for compatibility). Streaming multipart framing was extended to emit optional form fields with correct **Content-Length**; stage uploads pass no extra fields.
> 
> Tests assert form field order, query copy, payload integrity on large bodies, and that new-catalog zip uploads still send only the file part.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 8787c8e705459c53d21ee902700371e055a6086c. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


internal/drapi/filesapi/multipart.go
```diff
@@ -34,13 +35,19 @@ const multipartFormField = "file"
 // by the pipe (one chunk in flight) plus the small envelope, regardless
 // of file size — important because the engine may upload multi-GiB zips.
 //
+// fields are written as ordinary form parts ahead of the file part. The
+// Files API binds a POST's parameters from the parsed body only, so an
```
adamalpi: **[Medium] The stated rule is wrong, and the counter-example is two lines above the parameter being fixed**

"The Files API binds a POST's parameters from the parsed body only, so an option that must reach the server has to travel here rather than in query" is too strong. `useArchiveContents=true` is sent **query-only** on this exact request — `fromfile.go:54`, and `:29` for the new-catalog path — and it is demonstrably honoured:

- The reported symptom was per-path `name (2).ext` duplicates, which means the archive *was* extracted into individual catalog entries. Had `useArchiveContents` been ignored along with `overwrite`, the upload would have stored a single file called `changes.zip`.
- `ApplyStage` (`stage.go:92`) sends `overwrite` in a JSON body and works, which is exactly why the CHANGELOG's "smaller change sets, which go through a stage, were never affected" holds.

So the real rule is per-parameter — plausibly this endpoint's schema binds `overwrite` from the body and `useArchiveContents` from the query — not "body only".

The risk is concrete rather than cosmetic: a reader who acts on the rule as written moves `useArchiveContents` into the form and drops the query copy, and zip extraction silently stops. That is a quieter failure than the one this PR fixes, because the upload still succeeds and still produces a catalog version — just one holding a zip instead of the code.

Two follow-ons: the same wording landed in the Terraform provider (`internal/client/filesapi/fromfile.go`, datarobot-community/terraform-provider-datarobot#400), so the correction belongs in both; and the RAPTOR-20129 ask to the Files API team should be narrowed to `overwrite` specifically, since the same request already depends on the query for a sibling parameter.

internal/drapi/filesapi/multipart.go
```diff
@@ -34,13 +35,19 @@ const multipartFormField = "file"
...
+// query; query is kept for parameters the server reads from the URL.
+//
 // Trade-off: the request has no GetBody, so http.Transport cannot
 // transparently retry the body on connection reset. Callers needing
 // retry must redo the call from scratch (re-opening the source if it
 // isn't seekable).
 func newStreamingMultipartRequest(
 	requestURL string,
 	query url.Values,
+	fields url.Values,
```
adamalpi, line 57: **[Low] Two adjacent `url.Values` parameters with opposite destinations, and every call site passes a nil**

After this change `newStreamingMultipartRequest` takes `query` and `fields` back to back, and both live call sites pass at least one nil — `newStreamingMultipartRequest(requestURL, nil, nil, name, size, body)` at `stage.go:63`, and `(requestURL, nil, fields, ...)` at `fromfile.go:69`.

Transposing the two compiles cleanly and silently sends URL parameters as form parts and form fields as URL parameters, which is the same class of bug this PR exists to fix, and the tests here would not catch it on the stage path (both are nil there).

A small options struct, or distinct named types (`type queryValues url.Values` / `type formFields url.Values`), makes the swap a compile error for a few lines of cost.

adamalpi (approved):
Fix is correct and the tests are thorough. Two inline notes, neither blocking: the explanatory comment states a rule the same function disproves, and the new parameter is swappable with the one beside it.

PR #904, opened 2026-09-09
fix: correct dr self completion help examples

tsdaemon:
# RATIONALE

The `dr self completion --help` output showed incorrect usage examples: it printed
`dr completion bash` instead of the real `dr self completion bash`, since the
command's `Example` text was written before completion was nested under `self`
and never updated. The zsh example was also broken: it concatenated `_dr` +
`CliName` ("dr"), producing the invalid filename `_drdr` instead of `_dr`.

## CHANGES

- Fixed `cmd/self/completion/cmd.go` help `Example` text to use `dr self completion <shell>` for bash/zsh/fish/powershell.
- Fixed the zsh completion filename example from `_dr` + CliName (`_drdr`) to `_` + CliName (`_dr`).

---

> [!NOTE]
> **Low Risk**
> Help text only; no runtime behavior, auth, or data handling changes.
> 
> **Overview**
> Updates **`dr self completion --help`** examples so they match the real command path: all shell snippets now use **`self completion`** instead of the outdated top-level **`completion`**.
> 
> Also fixes the **zsh** install path example so the completion file is **`_` + CliName** (e.g. `_dr`) instead of the broken **`_dr` + CliName** concatenation that produced `_drdr`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 3426f9b3b2d34d34732bf34e0bfcdd4d790f788e. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/cmd/self/completion/cmd.go
+++ b/cmd/self/completion/cmd.go
@@ -39,12 +39,12 @@ You can also use the 'install' subcommand to install completions interactively.`
 
 Bash:
 
-  $ source <(` + version.CliName + ` completion bash)
+  $ source <(` + version.CliName + ` self completion bash)
 
   # To load completions for each session, execute once:
 
   # Linux:
-  $ ` + version.CliName + ` completion bash > /etc/bash_completion.d/` + version.CliName + `
+  $ ` + version.CliName + ` self completion bash > /etc/bash_completion.d/` + version.CliName + `
 
 Zsh:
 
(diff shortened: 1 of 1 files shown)


ajalon1:
Thanks @tsdaemon -- I had opened a similar PR myself for the install and uninstall subcommands.

ajalon1 (approved):
going to merge myself

chasdr (approved):
LGTM 👍

chasdr:
verified both fixes, self path and `_dr` are both right. 👍

heads up on a sibling of the same bug, out of scope for this PR: the powershell installer at `cmd/self/completion/install/cmd.go:485` generates `dr completion powershell` (no `self`) into the user's profile, which won't resolve now that the default completion cmd is disabled. `cmd_test.go:451/481` pin the broken string too. worth a follow-up?

PR #912, opened 2026-09-11
[CFX-7608] Reject a non-http/https URL in dr auth set-url

chasdr:
PR 3 of 3: #776 was PR 1, #911 is PR 2

## Summary

`dr auth set-url` accepted any URL scheme and wrote it straight into `drconfig.yaml`, so `dr auth set-url ftp://host` stored a dead endpoint that only failed later at read time with a vaguer error. It now rejects a non-http/https URL at write time and exits non-zero. A bare host like `app.datarobot.com` is still accepted and defaults to `https`.

## Notes for review

The http/https rule is one helper, `config.RequireHTTPScheme`, applied at both write paths that persist the endpoint: `SetURLToConfig` (set-url arg, the interactive picker, stdin, and `dr auth login <url>`) and `SaveURLToConfig` (the template-setup custom-host picker). `SchemeHostOnly` stays scheme-agnostic because `export` and `GetBaseURL` share it, so the rule sits on top rather than inside it.

## Output

```
$ dr auth set-url ftp://app.datarobot.com
ERROR unsupported URL scheme "ftp", use https://
$ echo $?
1
```

## Technical Changes

- internal/config/api.go: add `RequireHTTPScheme`. `SetURLToConfig` and `SaveURLToConfig` both call it after `SchemeHostOnly`.
- internal/auth/auth.go: `ValidateEndpoint` delegates its inline scheme check to the same helper. message byte-identical, no duplication.
- cmd/auth/seturl/cmd.go: a bad arg prints the reason and exits non-zero instead of silently opening the picker. `Run` becomes `RunE` with `SilenceErrors`/`SilenceUsage`.
- docs/commands/auth.md: note the rejection under `set-url`.
- tests: `RequireHTTPScheme` table, both writers' scheme rows, `set-url` arg exits non-zero.

### Breakdown

- code: +43 / -17
- tests: +84 / -0
- docs: +4 / -0

---

> [!NOTE]
> **Low Risk**
> Localized URL validation on auth/config write paths with tests; no change to token handling or API auth flow beyond earlier rejection of bad endpoints.
> 
> **Overview**
> **`dr auth set-url`** now rejects endpoints whose scheme is not `http` or `https` at configuration time, with a clear error and non-zero exit, instead of saving unusable values (e.g. `ftp://`) into `drconfig.yaml` or falling back to the interactive picker when the user passed an explicit bad URL.
> 
> Validation is centralized in **`config.RequireHTTPScheme`**, applied after normalization in **`SetURLToConfig`** and **`SaveURLToConfig`** (set-url, login, template setup). **`ValidateEndpoint`** uses the same helper. Bad schemes in the interactive picker still re-prompt because the error wraps **`ErrInvalidURL`**. Bare hosts still default to `https` via **`SchemeHostOnly`**.
> 
> Docs under **`set-url`** describe the rejection behavior; tests cover the helper, both write paths, and the set-url command argument path.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit e8e4738b8b1d03ad73a36b4ff382c663880411db. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/cmd/auth/seturl/cmd.go
+++ b/cmd/auth/seturl/cmd.go
@@ -16,15 +16,19 @@ package seturl
 
 import (
 	"github.com/datarobot/cli/internal/auth"
+	"github.com/datarobot/cli/internal/cli"
 	"github.com/datarobot/cli/internal/config"
+	"github.com/datarobot/cli/internal/log"
 	"github.com/datarobot/cli/internal/telemetry"
 	"github.com/spf13/cobra"
 )
 
 func Cmd() *cobra.Command {
 	cmd := &cobra.Command{
-		Use:   "set-url [url]",
-		Short: "🌐 Configure your DataRobot environment URL.",
+		Use:           "set-url [url]",
+		SilenceErrors: true,
+		SilenceUsage:  true,
+		Short:         "🌐 Configure your DataRobot environment URL.",
 		Long: `Configure your DataRobot environment URL with an interactive selection.
 
 This command helps you choose the correct DataRobot environment:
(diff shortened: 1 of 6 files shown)


chasdr:
bugbot run

PR #913, opened 2026-09-11
[RAPTOR-20140] fix(filesapi): name the File Registry entry a sync creates

wojtekwdr:
# RATIONALE

Code pushed to the File Registry showed up as `Untitled Dataset`, or as `wapi-sync.zip` when the change set was big enough to take the zip route, because the CLI never sent a name: one route falls back to a class constant, the other titles the entry after the file it was built from, which is our temp archive. The Registry accepts duplicate names silently, so a tenant ends up with a column of identical rows and no way to tell whose code is whose.

The Files API gained an optional `name` on its create routes in DataRobot#158141. This sends it.

## CHANGES

Entries are named `Artifact: <artifact id>` now, on both upload routes: JSON on `POST /files/`, a multipart form field on `POST /files/fromFile/`, since that API binds a POST from the parsed body and drops query parameters.

The id rather than the artifact's name, because naming is create-time only: the entry keeps whatever it was created with, so a name the project is free to change would go stale in place, and artifact names are not unique anyway (`up` derives one from the directory). The artifact's own id rather than its repository's, since a repository is a different entity. The cost of that is the id names the artifact that first pushed the code rather than the one deploying from it later, which keeps resolving because locked artifacts are retained.

Existing entries keep their names. No rename on sync, deliberately: it would mean writing over a name someone set by hand.

Checked on staging against both routes, before and after.

---

> [!NOTE]
> **Low Risk**
> Label-only change on catalog create for code upload paths, with tests and no rename of existing entries; behavior depends on the platform accepting the new `name` parameter.
> 
> **Overview**
> **`dr artifact code sync` and `dr workload up` now set a meaningful File Registry title on first catalog creation** instead of leaving `Untitled Dataset` (stage path) or `wapi-sync.zip` (large zip uploads).
> 
> New entries are named **`Artifact: <artifact-id>`**. The sync engine’s `newCatalogName` uses the artifact id (not the renamable artifact name or repo id) because Files API naming is create-only; later syncs do not rename existing rows.
> 
> The Files API client sends that name on **both** create paths: JSON `name` on `CreateCatalog`, and a **`name` multipart field before the file** on `UploadFromZipNew` (query params are ignored). Names are clamped to 255 characters via `clampCatalogName` / `CatalogNameMaxLen`. `UploadFromZipNew` now takes separate `catalogName` and zip `filename` arguments.
> 
> Docs (`CHANGELOG`, `artifact.md`) and tests cover JSON body, zip multipart, omission when empty, clamping, and engine stage vs zip first-sync behavior.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit eb420badf760a239340b027696ca938ae8d21d8c. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


internal/drapi/filesapi/types.go
```diff
@@ -33,6 +33,20 @@ const (
...
+// title>` and a blank string is falsy there: sending "" would fall through
+// to the platform's own title, which is the behaviour this parameter
+// exists to replace.
+//
+// This is the client's contract and not a back-compatibility story. The
+// sync engine always has an artifact to name, so a real first sync always
+// sends a name, and one against a server predating the parameter fails on
+// the unknown key rather than degrading.
+type CreateCatalogReq struct {
+	Name string `json:"name,omitempty"`
```
adamalpi, line 49: **[High] This makes the first sync of every project fail against a pre-#158141 Files API, for a cosmetic label**

This doc states the consequence plainly — *"one against a server predating the parameter fails on the unknown key rather than degrading"* — and I think that decision is worth surfacing outside a code comment, because it is a new hard version coupling between the CLI and the platform on a core path.

Why it bites in practice rather than in theory:

- `omitempty` does not save it. `newCatalogName` returns `""` only when `e.artifact == nil`, and the engine's own comment says that never happens ("Phase 1 always fetches the artifact before Phase 5 uploads"), so the key is always sent on a real sync.
- A rejection is fatal, not degraded: `drapi.Post` (`post.go:71-73`) turns any non-2xx into `ErrFromResp`. `dr workload up` on a fresh project dies at catalog creation with a raw API error, and the user has no way to connect that to a label they never asked for.
- The CLI documents self-managed/on-prem instances as a first-class target (`README.md:60`, `docs/commands/auth.md:429`, the "Custom/On-Prem" option in `dr auth set-url`), and those clusters routinely trail cloud. I grepped for any feature detection or minimum-version check in the repo and there is none — `docs/commands/start.md` version-checks the *CLI* against a template, never the platform against the CLI.

I could not test an old server from here, so whether it 422s or quietly ignores the key is unverified — which is exactly why this is worth confirming rather than assuming. Note the asymmetry while you do: an unknown **JSON key** is far more likely to be rejected than an unknown **multipart field**, so `CreateCatalog` is the exposed route and the zip path probably is not.

Three placements:

- **Retry without the name on rejection**, localized to `CreateCatalog` — every cluster keeps working, costs one wasted request on old servers, and the label falls back to `Untitled Dataset`, i.e. exactly today's behaviour, so nothing regresses.
- **Document a minimum platform version** in the CHANGELOG and `docs/commands/artifact.md` — zero code, but turns a cosmetic improvement into a support constraint operators discover from release notes.
- **Ship as-is** — correct only if every instance this CLI is expected to reach is already past #158141. That is a product call.

My read is that hard failure is the wrong default when the benefit is a label, but it is your call; it just should not be made only here.

internal/drapi/filesapi/multipart.go
```diff
@@ -39,6 +39,8 @@ const multipartFormField = "file"
 // The Files API binds a POST's parameters from the parsed body alone
 // and drops unrecognized query parameters without complaining, so an
 // option that has to reach the server travels here and not in the URL.
+// That covers the upload's overwrite mode and the name a newly created
+// catalog gets, both of which are silently lost from a query string.
 //
 // useArchiveContents on the fromFile routes reads like a counter-example
```
adamalpi, line 48: **[Medium] This documents a live silent-failure dependency and defers the one-line fix into code the PR is already rewriting**

First: this is a better answer than the one I gave on #898. "The query rule is over-general" was wrong in an interesting way — the query parameter really is dropped, and extraction survives only on a server-side default. Good dig.

But that makes the deferral the problem. By this comment's own account, extraction happens "only because the server's declared form default for that field is already true … a flip of that default would stop extraction with no error." The failure mode is a catalog holding a `wapi-sync.zip` blob instead of extracted source, an image built from that blob, and a zero exit code — quieter than the overwrite bug #898 just fixed.

And the fix is in code this PR is editing. `UploadFromZipNew` now builds a `fields` map at `fromfile.go:39` to carry `name`, and `UploadFromZipExisting` has had one since #898, so this is `fields.Set("useArchiveContents", "true")` in two functions already being rewritten, next to a parameter being moved for the identical reason.

"Moving it into the form is a separate change" is the only record that it needs doing — I grepped the repo and there is no ticket, TODO, or issue reference anywhere. Either do it here, or file it and cite the id in this comment so it does not rest on a sentence in a doc block.

internal/workload/sync/engine.go
```diff
@@ -306,3 +306,73 @@ func resolveExistingCatalogID(e *Engine) string {
...
+	}
+
+	// Label plus a platform id is nowhere near the API's ceiling, so unlike
+	// the fallback this needs no budgeting.
+	return catalogNameLabel + e.artifact.ID
+}
+
+// clampRunes truncates to a character count, matching how the Files API
+// measures a name.
+func clampRunes(s string, limit int) string {
```
adamalpi: **[Low] Two implementations of one clamping rule**

`clampRunes` duplicates `filesapi.clampCatalogName` (`catalogname.go:35`) exactly — same rune-slice truncation, same semantics. The duplication exists because the filesapi one is unexported while `CatalogNameMaxLen` beside it is exported.

Exporting `ClampCatalogName` instead of the bare constant would remove the copy and keep the "counted in characters rather than bytes because the server validates a decoded Python string" reasoning in one place. As it stands, a change to how the API measures a name has two call sites to find, and the one over here has no comment explaining why it counts runes.

internal/drapi/filesapi/fromfile.go
```diff
@@ -24,16 +24,29 @@ import (
...
-func (c *httpClient) UploadFromZipNew(name string, size int64, body io.Reader) (*FromFileResp, error) {
+// UploadFromZipNew creates a catalog from a zip's contents, named after
+// catalogName when it is non-empty. The name is a multipart form field
+// for the same reason overwrite is: the Files API binds a POST from the
+// parsed body alone. An empty name leaves the entry on the platform's
+// default, which the File Registry shows as "Untitled Dataset".
+//
+// filename is the zip's own name in the file part, which the server reads
+// as the source of the bytes rather than as a label for the entry.
+func (c *httpClient) UploadFromZipNew(catalogName, filename string, size int64, body io.Reader) (*FromFileResp, error) {
```
adamalpi: **[Low] Two adjacent same-typed strings whose transposition produces exactly the bug this PR fixes**

`UploadFromZipNew(catalogName, filename string, …)` — swap the two and it compiles cleanly, names the Registry entry `wapi-sync.zip`, and labels the file part `Artifact: <id>`. That is the precise defect being fixed here, reintroduced as a footgun. `UploadFromZipExisting` below takes three adjacent strings (`catalogID, filename, overwrite`).

The one call site (`upload_zip.go:135`) is correct and the new tests would catch a swap — but only because they assert both values, which is easy to lose in a later refactor. Distinct named types or a small request struct makes it a compile error instead. Same shape as the `query`/`fields` note on #898, now with strings; worth fixing once rather than noting each time the signature grows.

adamalpi (commented):
Reasoning and tests are strong; four inline notes, one of which is a rollout decision worth settling before merge.

internal/drapi/filesapi/types.go
wojtekwdr: It is the other way round, which changes the fix.

`POST /files/` had no `validator_class` before #158141, just a bare `post()` that never read the body, so an unknown key there is ignored and the entry keeps the platform default. `/files/fromFile/` already had `FormValidationMixin` and a validator without `name`, so that is the one that 422s. I put a junk key through both against staging and they reject unknown keys identically today, so the strictness is real; what differs is only which of them was strict back then.

That makes retry the worst option for the route that actually breaks, since re-sending means streaming the whole archive a second time. So the zip path now creates its catalog first and uploads into it, and the name only ever travels on the JSON create. There is no version coupling left to document or retry around. It costs one extra request on a first sync, and the empty-catalog-if-the-upload-fails case is what the stage path has always done, so the two paths are now one shape.

internal/drapi/filesapi/multipart.go
wojtekwdr: RAPTOR-19915, filed before this PR. It is not in the comment because we do not put ticket ids in code comments, which is why that line reads as bare as it does.

Leaving it rather than folding it in: it is a separate defect with its own ticket, and unlike `name` it is safe against every server version, since `use_archive_contents` has been a declared form field for years. Nothing about it is blocked on this change.

internal/workload/sync/engine.go
wojtekwdr: Fixed: `ClampCatalogName` is exported now and the copy is gone. Your read of the cause was right, exporting the constant while hiding the function is what forced the duplication.

internal/drapi/filesapi/fromfile.go
wojtekwdr: Gone, along with the function. The zip path creates its catalog separately now, so `UploadFromZipNew` does not exist and neither does the pair of adjacent strings.

`UploadFromZipExisting` still takes three, which is the #898 shape and a wider change than this one.

internal/workload/sync/upload_zip.go
```diff
@@ -120,14 +125,29 @@ func addToZip(zw *zip.Writer, src, archivePath string) error {
...
+// second time. The JSON route ignores what it does not recognize, so
+// there the same server just leaves its own default on the entry.
+//
+// The cost is an extra round trip on first sync and, if the upload then
+// fails, an empty catalog with nothing in it. Both are what the stage
+// path has always done, so this is one shape rather than two.
 func postZip(e *Engine, body io.Reader, size int64) (*filesapi.FromFileResp, error) {
-	if id := resolveExistingCatalogID(e); id != "" {
-		return e.files.UploadFromZipExisting(id, "wapi-sync.zip", filesapi.OverwriteReplace, size, body)
+	catalogID, err := ensureCatalog(e)
```
adamalpi, line 145: **[Low] The last asymmetry: the zip path holds the catalog id and then returns the server's echo instead**

Now that `postZip` gets `catalogID` from `ensureCatalog`, line 72 still returns `resp.CatalogID, resp.CatalogVersionID`, while the stage path returns its own id and takes only the version from the response (`upload_stage.go:52`). Returning `catalogID` here would make the two genuinely one shape — which is the stated goal of this change — and drop a dependency on `/files/<id>/fromFile/` echoing `catalogId` back.

That echo now matters on a **first** sync, where it did not before: previously the first sync went through the create-from-file route, whose response necessarily carried the id it had just minted, so the existing route's response only ever had to be right for subsequent syncs.

I checked before raising it and this is **not** a live break — `phase5_execute.go:282` already assigns `newCatalogID` from that same field on every subsequent zip sync in production, so the field is populated. It is robustness and consistency, not a regression.

One note on the test: `engine_test.go:466` sets `zipResp.CatalogID` to `"cid-zip"`, the same value as `fake.catalogID`, so both implementations pass identically. Making the two differ would pin which one the code actually uses.

internal/workload/sync/upload_zip.go
```diff
@@ -120,14 +125,29 @@ func addToZip(zw *zip.Writer, src, archivePath string) error {
...
+// Files API's create-from-file route, and used to. Creating separately
+// keeps the catalog's name on the JSON create call, which is the only
+// route where naming is safe against every server: the create-from-file
+// route validates its multipart form strictly, so a name reaching a
+// server that predates the parameter fails the whole upload, and the
+// retry that would paper over it means streaming the entire archive a
+// second time. The JSON route ignores what it does not recognize, so
+// there the same server just leaves its own default on the entry.
+//
+// The cost is an extra round trip on first sync and, if the upload then
```
adamalpi, line 141: **[Low] The empty-catalog case is parity with the stage path, but it is new to this one**

"Both are what the stage path has always done" is accurate about the stage path and worth stating — the wrinkle is that `UploadFromZipNew` was atomic. No catalog existed unless the upload succeeded, so the zip path is *acquiring* this failure mode rather than matching one it already had.

`phase6_state.go:38` writes `cfg.CatalogID` only after phase 5 succeeds, so a failed first sync leaves an entry nothing has recorded, and the retry's `resolveExistingCatalogID` comes back empty and creates another.

What makes it worth a line in this PR specifically: those leftovers now carry an identical `Artifact: <id>` name, so repeated failures reproduce exactly the indistinguishable column of rows this change exists to remove. Previously they would at least have been the same `Untitled Dataset` nobody expected to be unique.

Not worth blocking on, and I would not pin the id before the upload — that trades a leak for a wrong binding, which is worse. Just that the trade reads better stated as what it is than as parity.

adamalpi (approved):
Thanks for the correction on the route strictness — deleting the create-from-file route rather than retrying around it is a better fix than any of the three I offered. Two small leftovers from the restructure inline, neither blocking.

PR #917, opened 2026-09-14
feat(workload): create with --use-case-id, render typed placement errors

alexromanyuk:
## Why

Enclave placement is opt-in per workload and governed by a Use Case: an Enclave-targeted create must carry a top-level `useCaseId`, and placement is restricted to the Enclaves an administrator has granted to that Use Case. The CLI's existing `--enclave` pin predates this contract, so a pin without a Use Case is now rejected by the server, and there was no way to opt in to availability placement at all. Placement rejections are also typed (`{"code": ..., "message": ...}`) and rendered here as a raw JSON dump.

## What

- `dr workload create --use-case-id <id>`: writes the top-level `useCaseId`. On its own it sets `runtime.enclaveSelectionPolicy` to `"availability"` (DataRobot picks among the Use Case's granted Enclaves); combined with `--enclave <name>` the pin stays `"manual"`. Same conventions as the pin flag: refuses to override a spec that already sets the field, spec is re-encoded with numbers preserved.
- `--enclave` now requires `--use-case-id`, refused locally instead of paying a round trip for a guaranteed server rejection.
- Typed placement rejections render as the server's sentence with the machine code in parentheses, e.g. `... (MISSING_USE_CASE)`, via a generic `{code, message}` detail shape in the Workload API error envelope. Other detail shapes are unchanged.
- Help text and constants updated: "availability" is no longer described as the default; without a policy the workload is not placed on an Enclave.

## Testing

- `internal/workload`: `ApplyUseCase` unit tests (defaulting, pin composition, spec conflicts, blank/trim, non-object specs).
- `cmd/workload/create`: wire-level tests for `--use-case-id` alone and with `--enclave`, the local pairing refusal, and spec-conflict errors; existing pin tests updated for the pairing rule.
- `internal/workload/apiclient`: typed-detail rendering, including the guard that other object shapes stay raw JSON.
- `go test ./...`: green except `cmd/dotenv` Pulumi-login tests, which fail identically on `main` on this machine (environment-dependent).

---

> [!NOTE]
> **Medium Risk**
> Breaking change for scripts that pass `--enclave` without `--use-case-id`, plus altered placement semantics on workload create; changes are well-tested and align with the server contract.
> 
> **Overview**
> **`dr workload create`** now supports **Use Case–governed Enclave placement** via `--use-case-id <id>`, which sets top-level `useCaseId` and, by itself, `runtime.enclaveSelectionPolicy` to `"availability"` so DataRobot chooses among Enclaves granted to that Use Case. **`--enclave`** still pins a named Enclave (`manual` policy) but **must be paired with `--use-case-id`**; the CLI fails locally instead of relying on a server rejection.
> 
> Enclave placement is documented as **opt-in**: without these flags (or equivalent spec fields), creates are not targeted to an Enclave. Flag/spec conflict rules are unchanged—flags refuse to override fields already set in the spec.
> 
> Workload API errors whose `detail` is a typed `{code, message}` object (e.g. placement rejections) are shown as **`message (CODE)`** instead of a raw JSON blob; other detail shapes behave as before.
> 
> `ApplyUseCase` and a shared `decodeSpecObject` helper were added alongside the existing enclave pin logic; create command order applies the pin first so a manual policy from `--enclave` is preserved when the Use Case is applied.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit e56b6663509c22abcc3abf095cc13204948939f4. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


cmd/workload/create/cmd.go
```diff
@@ -142,6 +151,13 @@ Example:
 				return err
 			}
 
+			// The server requires a Use Case for any Enclave-targeted create;
+			// fail here instead of a guaranteed round-trip rejection.
+			if cmd.Flags().Changed("enclave") && !cmd.Flags().Changed("use-case-id") {
+				return errors.New(
+					"--enclave requires --use-case-id: Enclave placement is governed by a Use Case")
+			}
```
ajalon1, line 159: I'm surprised cobra doesn't have this modeled out.

I would suggest pulling this out into its own helper method, similar to other cobra helpers, something that we might be able to pull up into a shared module. You don't have to pull it up into its own module; but I'd love to see this function here.

```
func RequireFlagWhenChanged(cmd *cobra.Command, flagName, requiredFlagName string)
error {
  if cmd.Flags().Changed(flagName) && !cmd.Flags().Changed(requiredFlagName) {
          return fmt.Errorf("--%s requires --%s", flagName, requiredFlagName)
  }

  return nil
}
```

internal/workload/enclave.go
```diff
@@ -56,6 +92,35 @@ func ApplyEnclavePin(spec []byte, enclave string) (json.RawMessage, error) {
...
+
+// ApplyUseCase links a workload create spec to a Use Case: the top-level
+// useCaseId field the server reads at create time. Enclave placement is
+// opt-in per workload and always governed by a Use Case, so when neither
+// the spec nor an already-applied pin has chosen an enclaveSelectionPolicy,
+// the policy becomes "availability" and DataRobot picks among the Enclaves
+// granted to the Use Case. It errors if the spec already sets useCaseId
+// rather than silently rewriting it. Re-encodes the spec the same way
+// ApplyEnclavePin does.
+func ApplyUseCase(spec []byte, useCaseID string) (json.RawMessage, error) {
```
ajalon1, line 106: We try to not DRY code in this codebase until we hit 3 distinct callers, but I know that Use Cases as a concept will be valuable in other places. I'd like you to consider splitting this up so that useCaseID parsing is its own function, maybe under `internal/workload/usecase` or something like that, and applying said ID to the spec object is something that can stay here.

internal/workload/apiclient/apiclient.go
ajalon1, line 1: Some potential improvements here, esp. since these are specific to the Workload API which is FastAPI based. I'm going to avoid being prescriptive about how far you might go, and propose only that you use better typing.

internal/workload/apiclient/apiclient.go
```diff
@@ -72,9 +73,12 @@ func LiftDetail(err error) error {
...
-// they carry the field-level messages the caller needs.
+// body, or "" when the body is not such a document. A typed rejection --
+// a {"code": ..., "message": ...} detail, the shape workload placement
+// errors use -- renders as the message with the machine code in
+// parentheses. Other non-string details (e.g. validation error arrays) are
+// re-encoded as JSON rather than dropped: they carry the field-level
+// messages the caller needs.
 func ErrorDetail(body []byte) string {
 	var payload struct {
 		Detail any `json:"detail"`
```
ajalon1, line 84: I would suggest using `json.RawMessage` over `any` here.

internal/workload/apiclient/apiclient.go
```diff
@@ -88,6 +92,15 @@ func ErrorDetail(body []byte) string {
 		return detail
 	}
 
+	if typed, ok := payload.Detail.(map[string]any); ok && len(typed) == 2 {
+		code, codeOK := typed["code"].(string)
+		message, messageOK := typed["message"].(string)
```
ajalon1, line 97: These could be a typed struct, like

```
var typed struct {
        Code    string `json:"code"`
        Message string `json:"message"`
}
````

File: cmd/workload/create/cmd.go [hunk 5]
```diff
@@ -170,12 +195,16 @@ Example:
...
+		"Pin the workload to the named Enclave (sets runtime.enclaveSelectionPolicy=manual); requires --use-case-id")
+
+	cmd.Flags().StringVar(&useCaseID, "use-case-id", "",
+		"Link the workload to this Use Case and place it on the Enclaves granted to it")
 
 	telemetry.TrackWith(cmd, func(_ *cobra.Command, _ []string) map[string]any {
 		return map[string]any{
 			"output_format": string(outputFormat),
 			"enclave":       enclave,
+			"use_case_id":   useCaseID,
```
ajalon1, line 207: You might consider avoiding raw `use_case_id` and `enclave` in telemetry. These might be treated as customer-identifying or sensitive values, and they are high cardinality. You may want to get opinions from your PM, analytics and legal -- I'm just raising the point from my experience here.

Most of our existing telemetry records things like

>    •  sanitized argument/source helpers, e.g. idargs.TelemetryID(...), ref.Source
   •  boolean flag values, e.g. "dry_run": f.dryRun
   •  enum-ish values, e.g. "output_format", "status"
   •  some raw IDs in older/current code, e.g. "workload_id", "artifact_id", "template_id"

I'd suggest adding

```
     "use_case_id_provided": cmd.Flags().Changed("use-case-id"),
     "enclave_provided":     cmd.Flags().Changed("enclave"),
     "placement_mode":       placementMode,
```

since those would be "safer" telemetry markers in the unlikely event you have to filter or remove this information.

ajalon1 (commented):
LGTM for changelog. I would strongly suggest typing improvements in apiclient.go, but nothing is blocking.

I leave final approval to the Workload team.

internal/workload/apiclient/apiclient.go
```diff
@@ -88,6 +92,15 @@ func ErrorDetail(body []byte) string {
 		return detail
 	}
 
+	if typed, ok := payload.Detail.(map[string]any); ok && len(typed) == 2 {
```
ajalon1, line 95: oh, `len(typed) == 2` is brittle and will cause this CLI error handling to break down as soon you add something on the API side.

internal/workload/apiclient/apiclient.go
```diff
@@ -88,6 +92,15 @@ func ErrorDetail(body []byte) string {
...
+		code, codeOK := typed["code"].(string)
+		message, messageOK := typed["message"].(string)
+
+		if codeOK && messageOK {
+			return fmt.Sprintf("%s (%s)", message, code)
```
wojtekwdr, line 100: [high] This never runs for `dr workload create`: `CreateWorkload` still posts through `drapi.PostJSON`, and only the build trigger goes through `apiclient.PostJSON`. Serving the create command a 422 with `{"detail":{"code":"MISSING_USE_CASE","message":"useCaseId is required"}}` still prints the raw `body={...}` dump, and `dr workload up` shares the same path. Routing `CreateWorkload` through `apiclient.PostJSON` fixes both, but it needs a create-side test with a typed body, since the existing 422 test only matches a substring and passes either way.

internal/workload/enclave_test.go
```diff
@@ -125,3 +125,83 @@ func TestApplyEnclavePin_RejectsNullSpec(t *testing.T) {
...
+
+	runtime, ok := doc["runtime"].(map[string]any)
+
+	require.True(t, ok)
+	assert.Equal(t, EnclaveSelectionPolicyManual, runtime["enclaveSelectionPolicy"])
+	assert.Equal(t, []any{"prod-east"}, runtime["enclaves"])
+}
+
+func TestApplyUseCase_KeepsASpecChosenPolicy(t *testing.T) {
+	spec := `{"name":"my-app","artifactId":"art-1","runtime":{"enclaveSelectionPolicy":"availability"}}`
```
wojtekwdr, line 164: [low] The spec here already uses `availability`, the same value the default writes, so this still passes if `ApplyUseCase` overwrites the policy unconditionally. I tried that mutation and only `KeepsAnAppliedPin` failed. A spec with `"enclaveSelectionPolicy":"manual","enclaves":["prod-east"]` would make it test what the name says.

cmd/workload/create/cmd.go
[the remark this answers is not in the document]
alexromanyuk: Fixed in 427d11b. `--enclave` now works when the spec already has `useCaseId`. Added a test for it.

internal/workload/apiclient/apiclient.go
```diff
@@ -88,6 +92,15 @@ func ErrorDetail(body []byte) string {
...
+		}
+	}
```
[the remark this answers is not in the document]
alexromanyuk: Fixed in 427d11b. `CreateWorkload` now uses `apiclient.PostJSON`, so create errors show "message (CODE)".

cmd/workload/create/cmd.go
alexromanyuk: Done in 427d11b. Added `cli.RequireFlagWhenChanged` in `internal/cli`, with tests. The create command uses it.

internal/workload/enclave.go
alexromanyuk: Done in 427d11b. Parsing is now `usecase.ParseID` in `internal/usecase`. It returns a typed `usecase.ID` and checks the 24-char hex format. Applying the ID to the spec stays in `internal/workload`.

internal/workload/apiclient/apiclient.go
alexromanyuk: Thanks. Went with better typing: `json.RawMessage` for the detail and a typed struct for code/message.

internal/workload/apiclient/apiclient.go
alexromanyuk: Done in 427d11b, the detail is now `json.RawMessage`.

internal/workload/apiclient/apiclient.go
alexromanyuk: Done in 427d11b, using a typed `{Code, Message}` struct.

cmd/workload/create/cmd.go
alexromanyuk: Good point, changed in 427d11b. Telemetry now sends `use_case_id_provided`, `enclave_provided` and `placement_mode`. No raw ids.

internal/workload/apiclient/apiclient.go
alexromanyuk: Agreed, removed in 427d11b. Extra fields are now ignored. Added a test with an extra field.

internal/workload/apiclient/apiclient.go
alexromanyuk: Fixed in 427d11b. `CreateWorkload` goes through `apiclient.PostJSON`, which also covers `dr workload up`. The new test checks the exact rendered message and that there's no raw JSON. I checked it fails on the old code.

internal/workload/enclave_test.go
alexromanyuk: Fixed in 427d11b. The test now uses a `manual` spec with `enclaves: ["prod-east"]` and checks both are kept.

internal/workload/enclave.go
```diff
@@ -56,6 +94,45 @@ func ApplyEnclavePin(spec []byte, enclave string) (json.RawMessage, error) {
...
+// error.
+func SpecSetsUseCase(spec []byte) bool {
+	doc, err := decodeSpecObject(spec)
+	if err != nil {
+		return false
+	}
+
+	id, ok := doc["useCaseId"]
+
+	return ok && id != nil
```
wojtekwdr: [medium] This checks the value, but `ApplyUseCase` on line 129 checks only that the key exists, and a spec with `useCaseId: null` falls in the gap: `--enclave prod-east` refuses with `--enclave requires --use-case-id`, and adding the flag refuses with `spec already sets useCaseId`. I ran both, so there's no flag combination that creates that spec. The same value check also lets `useCaseId: ""` or `useCaseId: 123` count as a named Use Case, so `--enclave` skips the local refusal and pays the round trip this gate exists to save.

internal/workload/enclave.go
```diff
@@ -65,18 +142,10 @@ func ApplyEnclavePin(spec []byte, enclave string) (json.RawMessage, error) {
 		}
 	}
 
-	if _, ok := runtime["enclaveSelectionPolicy"]; ok {
-		return nil, errors.New(
-			"spec already sets runtime.enclaveSelectionPolicy; remove it from the spec or drop --enclave")
+	if _, ok := runtime["enclaveSelectionPolicy"]; !ok {
```
wojtekwdr: [medium] A spec carrying `runtime.enclaves` with no policy comes out of here as `{"enclaveSelectionPolicy":"availability","enclaves":["prod-east"]}`, so the enclave list the user wrote becomes dead weight and nothing says so. `ApplyEnclavePin` refuses that same spec with `spec already sets runtime.enclaves`. Refusing here too would at least be consistent, rather than writing a policy that contradicts what's already in the file.

cmd/workload/create/cmd.go
```diff
@@ -170,14 +178,76 @@ Example:
...
+	cmd *cobra.Command, payload json.RawMessage, enclave, useCaseID string,
+) (json.RawMessage, error) {
+	var err error
+
+	// The server requires a Use Case for any Enclave-targeted create;
+	// fail here instead of a guaranteed round-trip rejection. A spec
+	// that already names one satisfies the requirement.
+	if !workload.SpecSetsUseCase(payload) {
+		if err := cli.RequireFlagWhenChanged(cmd, "enclave", "use-case-id"); err != nil {
+			return nil, fmt.Errorf(
```
wojtekwdr: [low] The `(or useCaseId in the spec)` half is the part that tells someone the way out, and nothing pins it. I swapped this wrap for a plain `return nil, err` and the package stayed green, because the test only does `Contains` on the prefix.

CHANGELOG.md
```diff
@@ -7,6 +7,8 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
...
 
+- `dr workload create --use-case-id <id>` places the workload on an Enclave through a Use Case: the workload is linked to the Use Case at create time and may only run on the Enclaves an administrator has granted to it. On its own the flag sets `runtime.enclaveSelectionPolicy` to `"availability"` and DataRobot picks among the granted Enclaves; combined with `--enclave <name>` the named Enclave is pinned instead. `--enclave` now requires `--use-case-id` unless the spec already sets `useCaseId`, because the server rejects any Enclave-targeted create without a Use Case; the CLI refuses locally instead of paying a round trip for a guaranteed rejection. A malformed `--use-case-id` (not a 24-character hex id) is also refused locally. Placement rejections are typed on the server, and the CLI now renders them as the server's sentence with the machine code in parentheses -- for example `useCaseId is required: this workload will be placed on an Enclave, and Enclave placement is governed by a use case. (MISSING_USE_CASE)` -- instead of a raw JSON dump. Enclave placement is opt-in per workload: without these flags (or the fields in the spec) the workload is not placed on an Enclave.
```
wojtekwdr: [low] The `--enclave` requirement breaks scripts that pass it alone, but it's filed under Added. The Unreleased block has a Changed section at line 40, where the `--import-env` removal and the `--accept-remote` CI change went, and that's where someone hunting for what broke their pipeline will look.

wojtekwdr (approved):
LGTM with some comments

internal/workload/enclave.go
alexromanyuk: Good catch, fixed in cf040fd. Both checks now use the same rule: `useCaseId` counts only if it's a non-empty string. `null` or `""` counts as not set, so `--use-case-id` fills it in. Any other type is refused as an invalid spec. Added tests for all three.

internal/workload/enclave.go
alexromanyuk: Fixed in cf040fd by removing the default: `--use-case-id` no longer sets a policy at all. A spec with `enclaves` and no policy now goes to the server as written, and the server rejects it with a clear error.

cmd/workload/create/cmd.go
alexromanyuk: Fixed in cf040fd. The test now checks the full message, including the `(or useCaseId in the spec)` part.

CHANGELOG.md
alexromanyuk: Agreed, fixed in cf040fd. Moved it to Changed and called out that scripts passing `--enclave` alone will now fail.

PR #918, opened 2026-09-14
[RAPTOR-20217] fix(sync): re-lock a stale uv.lock instead of uploading it

adamalpi:
# RATIONALE

Adding a dependency to `pyproject.toml` without re-running `uv lock` synced one file, built green, and deployed an image without the dependency in it. The generated Dockerfile installs the lockfile as given (`uv sync --frozen`), so the edited `pyproject.toml` was ignored and the failure arrived at container start as an `ImportError`, with nothing pointing back at the sync that caused it.

The check belongs in the CLI because it can only be correct here: `uv lock --check` needs the whole project tree, and the builder stage has just `pyproject.toml` and `uv.lock` copied in. Doing it there was tried and reverted — it rejects changes the image never consumes, and a `dynamic = ["dependencies"]` project can never be made to pass. Full write-up on the ticket.

## CHANGES

`phaseLockfile` now compares `uv.lock` against `pyproject.toml` before the walk, so anything it writes flows through the normal diff/upload pipeline:

| situation | behaviour |
| --- | --- |
| no `pyproject.toml` | untouched |
| `uv.lock` missing | generate, warn, continue (unchanged) |
| `uv.lock` current | continue; uv is invoked for the check only |
| `uv.lock` stale, re-lock works | regenerate, warn, upload it |
| `uv.lock` stale, `uv lock` fails | refuse, nothing uploaded |
| uv not installed | warn that it could not be checked, continue |
| uv workspace member | left alone, with a warning naming the workspace root |

Refusing is the one place this differs from a missing lockfile, which warns and continues: a missing lockfile already fails the build loudly, so continuing costs nothing, while a stale one fails nowhere. Not knowing is not the same as knowing it is wrong — with no uv on PATH there is nothing to compare, and refusing there would break every project whose lock is current and whose CI has no uv.

A workspace member is skipped entirely: every uv lock operation there acts on the root's lockfile rather than the `uv.lock` this directory uploads, so uv's answer is about a different file in both directions.

Residual, unchanged by this: callers that do not go through the CLI — the API, the UI, Pulumi — can still submit a stale lock, and a user-provided Dockerfile is outside this entirely.

## PR Automation

> [!IMPORTANT]
>

## BREAKING

No flag or output-shape change. One documented behaviour changes: `docs/commands/artifact.md` promised *"An existing `uv.lock` is never modified"*, which is no longer true — updated here. The same sentence appears in `build-from-code.md` in the GA docs PR and needs a companion update there.

## TESTING

```bash
# stale lock is refreshed
cd <python project with a committed uv.lock>
# add a dependency to pyproject.toml, do NOT run uv lock
dr artifact code sync
#   WARN  uv.lock was out of date with pyproject.toml — regenerated it

# stale lock that cannot be fixed refuses, uploading nothing
dr artifact code sync    # exits non-zero
```

`task test`, `task lint` (0 issues across linux/darwin/windows) and `task dupcheck` all pass.

## NOTES

- **`--dry-run` and `--diff` are affected**, because the phase runs inside `Plan()`. That matches the existing missing-lockfile behaviour, but it does mean a preview can now write `uv.lock` and can fail. A preview showing an upload set that would build the wrong image seemed worse than an error. Flag help and docs updated to say so.
- A refusal reaching `dr workload up` arrives behind `run.go`'s "cannot compare the working tree with the last deploy:" prefix. The actionable text survives; making it read cleanly needs a typed error plus unwrapping in `up`, which is wider than this fix. Happy to do it if preferred.
- `uv lock --check` against a current lockfile resolves from the lock itself — measured at ~16 ms, no network. It reaches the index only when something changed.

## RELATED

- https://datarobot.atlassian.net/browse/RAPTOR-20217
- `datarobot/workload-api#1283` — the build-side attempt, reverted
- `datarobot/workload-api#1288` — records why the check cannot live in the build


internal/workload/sync/phase_lockfile.go
```diff
@@ -100,19 +127,140 @@ func phaseLockfile(e *Engine) error {
...
+// where the sync has to continue rather than refuse.
+//
+// Cheap enough to run on every sync of a Python project: against a
+// current lockfile it resolves from the lock itself, with no network and
+// no measurable cost. It reaches the index only when something has
+// actually changed, which is the run that was about to be wrong.
+func runUvLockCheck(dir string) (bool, error) {
+	if _, err := runUv(dir, "uv lock --check", "lock", "--check"); err != nil {
+		var exitErr *exec.ExitError
+		if errors.As(err, &exitErr) {
```
wojtekwdr: [medium] uv exits 1 when the lock is outdated but 2 when it can't run at all, and this maps both to stale. On a uv older than 0.5.8, which has no `--check`, the check exits 2, `uv lock` exits 0, the re-check exits 2 again, and every sync of a project with a perfectly current lock is refused with the workspace message (same for a current lock whose `requires-python` has no local interpreter and downloads are off). Treating only `exitErr.ExitCode() == 1` as stale and returning the error otherwise puts those on the warn-and-continue path, like uv missing.

internal/workload/sync/phase_lockfile.go
```diff
@@ -100,19 +127,140 @@ func phaseLockfile(e *Engine) error {
...
+		return fmt.Errorf("uv.lock is out of date with pyproject.toml and could not be regenerated: %w. "+
+			"The image would be built from the old lock, so nothing was uploaded — "+
+			"run `uv lock` in the project and re-sync", err)
+	}
+
+	// Same reason generateLockfile re-checks the filesystem rather than
+	// trusting the exit code: `uv lock` in a workspace member succeeds
+	// while writing the lockfile at the workspace root, leaving this
+	// directory's copy exactly as stale as it was.
+	if current, err := e.lockfileCheckFn(e.projectDir); err == nil && !current {
```
wojtekwdr: [medium] This re-check can't catch the workspace case the comment describes. From a member directory `uv lock --check` judges the root lockfile that the `uv lock` above just rewrote, so it passes: I tried it with uv 0.11.26 and the member's own uv.lock was byte-identical afterwards, yet this logs "regenerated it" and uploads the stale copy. Comparing the file's contents before and after the runner would be independent of uv, the way the re-stat in `generateLockfile` is, and the test at line 196 only passes because its fake checker never reads the file.

internal/workload/sync/phase_lockfile.go
```diff
@@ -100,19 +127,140 @@ func phaseLockfile(e *Engine) error {
...
+// right is a wrong image nothing else will stop, so it stops the sync.
+// The exception is not knowing at all — with no uv on PATH there is
+// nothing to compare, and refusing would break every project whose
+// lockfile is current and whose CI has no uv, which synced fine before
+// this check existed.
+func refreshLockfile(e *Engine) error {
+	current, err := e.lockfileCheckFn(e.projectDir)
+	if err != nil {
+		if errors.Is(err, errUvNotFound) {
+			e.lockfileHint = "uv is not installed, so uv.lock could not be checked against pyproject.toml. " +
```
wojtekwdr: [low] Setting `lockfileHint` here trips the once-only guard in `warnIfLockfileIgnored`, which never fired before for a project that has a uv.lock. With no uv on PATH and `uv.lock` in `.drignore`, the "excluded by .drignore, the image build requires it" warning is gone and only the could-not-check one shows. Warning without assigning the field would keep both.

internal/workload/sync/phase_lockfile_test.go
```diff
@@ -100,16 +118,135 @@ func TestEngine_Plan_LeavesExistingLockfileAlone(t *testing.T) {
...
+	check := func(string) (bool, error) { return relocked, nil }
+
+	e := lockfileEngineWithCheck(t, dir, runner, check)
+
+	plan, err := e.Plan()
+	require.NoError(t, err)
+
+	assert.True(t, relocked, "a stale uv.lock is re-locked")
+	assert.True(t, e.lockfileGenerated)
+	assert.Contains(t, uploadPathsOf(plan), "uv.lock", "the refreshed lock is uploaded with the rest")
```
wojtekwdr: [low] nit: this assertion passes even if nothing was re-locked, since a first sync uploads the fixture's uv.lock regardless. Checking the upload's `LocalHash` against the rewritten `version = 2` content would tie it to the refresh.

docs/commands/artifact.md
```diff
@@ -177,7 +177,8 @@ dr artifact code checkout [<ver>] [--dir <path>] [--clean]
...
+- An existing `uv.lock` is kept in step with `pyproject.toml`. `sync` checks the two against each other and re-runs `uv lock` when they have diverged, uploading the refreshed lock — commit it. This matters because nothing downstream catches a stale lock: the image build installs the lockfile as given, so editing `pyproject.toml` without re-locking used to produce an image missing the dependency you added, with sync and the build both reporting success and the failure arriving at container start as an `ImportError`. The check needs your whole project to be accurate, which is why it lives here rather than in the build. If the lock is out of date and `uv lock` cannot put it right, sync stops and uploads nothing, rather than shipping code the image will not match. If `uv` is not installed there is nothing to compare with, so sync says so and continues — a project whose lock is already current is unaffected. `sync` also warns if your `.drignore` excludes `uv.lock`.
```
wojtekwdr: [low] The bullet above says generation also runs on `--dry-run`/`--diff`, and this one needs the same sentence, because here it means a preview (and the comparison `dr workload up` runs before asking anything) can rewrite a committed uv.lock or exit non-zero. Both flags' help still says no writes.

wojtekwdr (approved):

internal/workload/sync/phase_lockfile.go
adamalpi: Confirmed and fixed in 703fabb. Checked the exit codes rather than trusting either of us:

```
current lock, --check   -> 0
stale lock,   --check   -> 1
unknown flag            -> 2      # stands in for a uv with no --check
broken pyproject.toml   -> 2
```

So exactly as you said. `runUvLockCheck` now treats only exit 1 as an answer; 2 returns an error carrying uv's own output, which puts it on the warn-and-continue path beside uv-missing. The old-uv case you describe was the worst failure this phase could have had — refusing every sync of a project whose lock is perfectly current — and it would not have shown up in any test I had, since they all inject the checker.

Added `TestRunUvLockCheck_TellsStaleFromCannotRun`, which puts a stub `uv` on PATH exiting 0/1/2 and asserts the mapping (skipped on Windows, the stub is a shell script). That is the one test that exercises the real checker.

internal/workload/sync/phase_lockfile.go
adamalpi: Right, and worse than you put it — the check does not even reach the runner. I built a workspace with a member holding its own `uv.lock`:

```
from members/app:  uv lock --check  -> exit 0     # judging the ROOT lockfile
                   uv lock          -> exit 0
                   uv lock --check  -> exit 0
member uv.lock: byte-identical
```

So the first check answers "current" about a file we are not uploading, and my guard never runs. Fixed in 703fabb by comparing this directory's bytes before and after the runner, as you suggested — independent of where uv decided to work, the way `generateLockfile`'s re-stat is.

I have also corrected the comment, which claimed the re-check caught the workspace case. It did not, and the content comparison only catches the half where we reach the runner: a member whose check passes against the root is still undetectable from here without reimplementing uv's workspace resolution. The comment now says what the guard actually covers rather than implying the whole case is handled.

The test is rewritten so it fails without the fix: the runner leaves the file alone and the checker reports current afterwards, which is the shape you measured.

internal/workload/sync/phase_lockfile.go
adamalpi: Fixed in 703fabb — the warning no longer assigns `lockfileHint`.

You are right that this was a new suppression rather than the existing pattern. I had looked at this during self-review and talked myself out of it, reasoning that the generate path does the same thing; what I missed is that there it is harmless, because when generation fails there is no `uv.lock` and `warnIfLockfileIgnored` returns at its own second guard anyway. Mine is the first path that sets the hint while a `uv.lock` exists, which is precisely when that warning is wanted.

The message moved into `uncheckableLockfileWarning(err)` so the wording stays testable without capturing logs, and `TestEngine_Plan_UncheckableLockfileStillWarnsWhenIgnored` pins that both warnings survive.

internal/workload/sync/phase_lockfile_test.go
adamalpi: Fixed in 703fabb. The test now compares the upload's `LocalHash` against the sha256 of the `version = 2` content the runner writes, so it fails if the fixture's copy is what got uploaded. You are right that the old assertion was vacuous on a first sync.

docs/commands/artifact.md
adamalpi: Fixed in 703fabb, and I took the flag help with it since you are right that it was the bigger lie.

- `--dry-run` / `--diff` help: "Show plan; no remote writes (may still refresh uv.lock)."
- The long help now spells out that the lockfile is the one local write a preview makes, and that a lock it cannot put right stops the preview too.
- The docs bullet says the same, and names the comparison `dr workload up` runs before asking anything, which is the case where it is least expected.

internal/workload/sync/phase_lockfile.go
```diff
@@ -100,19 +132,165 @@ func phaseLockfile(e *Engine) error {
...
+		// for. Suppressing it here would lose "uv.lock is excluded by
+		// .drignore" on every machine without uv.
+		log.Warn(uncheckableLockfileWarning(err))
+
+		return nil
+	}
+
+	if current {
+		return nil
+	}
```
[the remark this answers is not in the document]
adamalpi: Confirmed and fixed in d0aa4c8. This is a real hole and the second half of it is the part I had missed.

I had documented that `uv lock --check` from a member judges the root lockfile, and left it at "the content comparison only catches the half where we reach the runner". What I did not follow through on is that **the check makes the bad case likelier rather than less**, which is your point. Reproduced against uv 0.11.23 on a real workspace, with a stale root and a stale member lock:

```
sync 1:  uv lock --check -> 1   (root is stale)
         uv lock         -> 0   (rewrites the ROOT)
         member uv.lock unchanged -> byte comparison refuses  ✅
sync 2:  uv lock --check -> 0   (root is now current)
         member uv.lock still stale -> returns early, uploads it  ❌
```

So the refusal was a one-shot that degraded into the exact silent-wrong-image this PR exists to prevent. Worse, the refusal message says "run `uv lock` in the project and re-sync" — the action that creates the state where sync 2 passes.

Fixed by not asking the question at all in a workspace. A member is detected up front (walking ancestor `pyproject.toml` files for the `[tool.uv.workspace]` table header) and left alone with a warning naming the root, which is what every sync did before this check existed. One honest warning per run beats one refusal that turns into a wrong image on the retry — and nothing here can resolve it properly, since uv owns where the lockfile lives.

Tests: `TestEngine_Plan_WorkspaceMemberIsNotChecked` (checker and runner both untouched), `TestEngine_Plan_AncestorPyprojectWithoutWorkspaceIsStillChecked` (no false positive on an unrelated ancestor package), and `TestUvWorkspaceRoot` (a commented-out or quoted mention is not a declaration; a root is not a member of itself).

PR #906, opened 2026-09-09
CMPT-7184 CMPT-7759 feat(pipeline): remove feature gate for GA and publish docs

sunny2get:
# RATIONALE

Make `dr pipeline` generally available and publish its docs. The command was hidden from the command tree, shell completion, and `dr --help` unless `DATAROBOT_CLI_FEATURE_PIPELINE=true` was set, and its reference pages were absent from the site nav.

Only the `pipeline` gate is removed — `dr workload` / `dr artifact` stay gated. No subcommands or flags changed, so scripts that set the variable keep working; it's just no longer read for this command.

## CHANGES

- **Code:** drop `features.SetGate(cmd, "pipeline")` (and its import) from `cmd/pipeline/cmd.go`; remove the "Private Preview" note from the help text.
- **Tests:** flip `TestCmd_FeatureGate` → `TestCmd_NotFeatureGated`; add `TestPipelineCommandPresentByDefault`; register the 38 telemetry-wired `dr pipeline …` paths in `expectedTrackedCommands`.
- **Docs:** add `pipeline.md` + `pipelines-reference.md` to the MkDocs nav and `docs/index.md`; strip the gate/preview notes; fix the command tree (`environment` → `image`, add `clone`/`source`, and the missing `run task` subgroup).
- CHANGELOG entry under `[Unreleased] → Changed`.

## TESTING

- `task lint` — 0 issues (linux/darwin/windows); `task test` — full suite passes.
- Confirmed `dr pipeline --help` is visible with `DATAROBOT_CLI_FEATURE_PIPELINE` unset, and all 38 leaves carry the `telemetry` annotation.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #919, opened 2026-09-15
[RAPTOR-20300] feat(workload): remove the feature gate except for up and config

wojtekwdr:
# RATIONALE

`dr workload` and `dr artifact` are ready for general use, but the feature gate keeps both trees out of `dr --help`, shell completion and dispatch unless every user exports `DATAROBOT_CLI_FEATURE_WORKLOAD=true`. Two subcommands are not ready: `dr workload up` and `dr workload config` are still being finished and should stay out of the tree until they are. The gate sits on the two roots, so it cannot express that today; the only way to hide the two unfinished commands is to hide everything. This moves the gate down to exactly those two.

## CHANGES

- `dr workload` and `dr artifact` register unconditionally, the same shape as the pipeline GA (#906).
- The existing `workload` gate moves onto `config` and `up` only, through `cli.CommandAdder` on the workload root. Same variable, so nothing that exports it today changes. Without it the two commands are absent from help, completion and dispatch (`unknown command`), not merely hidden: a hidden-but-runnable deploy command would be more available than it is today.
- Tests: the root "not present by default" tests are inverted, a new one asserts `up` and `config` stay absent, the workload package pins which verbs the gate covers, and the workload/artifact telemetry paths join the live list.
- Docs: gate notes removed from the workload and artifact pages and the command index; five sentences on those public pages that named `dr workload up` reworded, since the command is not public; the feature-gates page points at the live nested example instead of presenting `dr workload` as gated. Changelog entry under Changed.
- `internal/workload/build.go`: the trigger-timeout message, also reached from `dr artifact build create`, no longer says to re-run `dr workload up`.
- Docs catch-up for flags that shipped without a mention: `--enclave` (workload create and list), `--offset` (both list commands), and `--accept-remote` on `code sync`, whose conflict paragraph still described remote-wins. `docs/index.md` now lists both command groups beside pipeline.

Left alone on purpose: CLI strings that name `dr workload up` behind a manifest binding (`workload delete --help`, id resolution, locked-artifact sync) and contributor docs under `docs/development/`.

Checked on a local build: `dr --help` lists both roots; `dr workload --help` shows nine verbs without the variable and eleven with it; `dr workload up` answers `unknown command: up`; `__complete` offers neither.

Ticket: https://datarobot.atlassian.net/browse/RAPTOR-20300

---

> [!NOTE]
> **Low Risk**
> Localized CLI registration and documentation changes with no auth or data-path impact; the only behavioral shift is broader command discovery, with unfinished deploy/setup verbs still gated.
> 
> **Overview**
> **`dr workload` and `dr artifact` are now visible by default** in `dr --help`, shell completion, and dispatch without `DATAROBOT_CLI_FEATURE_WORKLOAD`. The workload feature flag now applies only to **`dr workload up`** and **`dr workload config`**, which stay **unregistered** (unknown command) when the variable is unset—not merely hidden.
> 
> Registration changes: the artifact root drops its gate entirely; the workload root registers most verbs unconditionally and wraps only `config` and `up` with `features.SetGate` via `cli.CommandAdder`.
> 
> Tests and telemetry wiring are updated to expect workload/artifact in the default tree, assert `up`/`config` stay absent without the env var, and track GA workload/artifact paths on the live root command.
> 
> Docs and changelog drop feature-gate callouts for these commands, add navigation/index entries, document some existing flags (`--enclave`, `--offset`, `--accept-remote`), and reword public pages that assumed `up` is always available. The build-trigger timeout message now says to re-run **the command** instead of naming `dr workload up` (also used from `dr artifact build create`).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit ce80cae9e4362332c44636db36aaeeed634d544f. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


cmd/workload/cmd.go
```diff
@@ -60,7 +71,7 @@ Manage and monitor workloads in your deployment infrastructure.`,
 		start.Cmd(),
 		status.Cmd(),
 		stop.Cmd(),
-		up.Cmd(),
+		gated(up.Cmd()),
```
adamalpi, line 74: **[Medium] Two runtime error paths send a GA user to a command they cannot run**

`cmd/workload/internal/idargs/idargs.go:280` and `:313` both name `dr workload up`, and the PR lists them as deliberate because they sit "behind a manifest binding". I think that criterion does not hold, because `.datarobot.yaml` is a **committed** file — a manifest binding implies a teammate had the gate, not that this user does.

The documented flow makes it concrete: `config` writes the manifest, you commit it, `up` deploys and stamps the `workloadId`. A teammate who commits between those two steps leaves a manifest with no `workloadId`. A GA colleague clones, runs `dr workload status`, and gets:

```
<path> specifies no workloadId yet. Run 'dr workload up', or pass an id
$ dr workload up
Error: unknown command: up
```

Line 313 is the same shape behind a 404 — right manifest, wrong endpoint or organisation — which is exactly the case that message exists to explain.

This is the one of my three notes where the remedy genuinely dead-ends rather than just reading oddly: `dr workload create` **is** GA, so there is a correct message available. Naming the id flag alone would also do. (Everything under `cmd/workload/up/` and `cmd/workload/config/` is fine as-is — those strings are only reachable with the gate on.)

cmd/workload/cmd.go
```diff
@@ -41,13 +42,23 @@ func Cmd() *cobra.Command {
 Manage and monitor workloads in your deployment infrastructure.`,
 	}
 
-	features.SetGate(cmd, "workload")
+	// The gate that used to hide this whole tree now hides only the two
+	// commands that are not finished. Everything else is generally available;
+	// `config` and `up` stay behind DATAROBOT_CLI_FEATURE_WORKLOAD=true, and
+	// the adder leaves them out at registration while it is unset, so they are
+	// absent from help, completion and dispatch rather than merely hidden.
```
adamalpi, line 49: **[Low] `dr workload delete --help` names `dr workload up`**

Same "behind a manifest binding" rationale, but `--help` is behind nothing at all — no manifest, no binding, no auth. It is the most reachable string in the set. Verified against a build of this branch:

```
$ dr workload delete --help
  ... workloadId is removed too, so the next 'dr workload up' creates a new
      workload rather than pointing at one that is gone.
$ dr workload up
Error: unknown command: up
```

That is `cmd/workload/del/cmd.go:54`. Either reword the clause or drop it — it needs revisiting when `up` ships either way, so the cost is the same now as later.

This comment is anchored here rather than on the file itself because `del/cmd.go` is not in the diff; the decision that creates the dangling reference is the one above.

cmd/telemetry_wiring_test.go
```diff
@@ -90,6 +89,29 @@ var expectedTrackedCommands = []string{
...
+	"dr artifact list",
+	"dr artifact delete",
+	"dr artifact lock",
+	"dr artifact build create",
+	"dr artifact build get",
+	"dr artifact build list",
+	"dr artifact build logs",
+	"dr artifact code init",
+	"dr artifact code sync",
+	"dr artifact code versions",
```
adamalpi, line 114: **[Low] `dr artifact code checkout` goes GA untracked, and this test cannot catch it**

I checked every leaf under `cmd/artifact` and `cmd/workload` for a `telemetry.Track`/`TrackWith` call. `checkout` is the only one without one — the other misses are group commands, which legitimately have none. It is also the only leaf missing from the list this PR otherwise extends carefully.

The reason it slipped is structural and worth more than the missing line: `TestTelemetryWiring_AllCoreCommandsTracked` iterates `expectedTrackedCommands` and asserts each entry **is** tracked. It never asserts the converse, so a registered leaf absent from this hand-maintained list is invisible to it. The test is named `AllCoreCommandsTracked` but measures "all listed commands", and the gap it just missed is precisely the one its name promises to cover.

Adding `"dr artifact code checkout"` fixes today. Walking the tree and failing on an untracked leaf would stop the next one — that list only grows, and this PR is the moment it grew most.

adamalpi (approved):
Verified the gating on a local build — 9 verbs bare, 11 with the variable, `up` answers `unknown command`. Three inline notes, none blocking. Heads-up unrelated to the review: #918 also edits `docs/commands/artifact.md`, so whichever of us lands second needs a rebase on that file.

PR #922, opened 2026-09-15
[RAPTOR-19939] feat(workload): one --sync-env flag, with a table you approve

wojtekwdr:
# RATIONALE

`--import-env` and `--update-env` were two flags for one intent, and the reader had to work out which half applied to what. One flag does both. Because collapsing them means a single word rewrites a committed file and overwrites values on a tenant-wide credential store, the run shows its work and asks.

Re-lands #888, which was merged into the head branch of #882 after that branch had already been squash-merged into main, so main never received it: v0.5.0, v0.5.1 and main all still offer the two old flags. The code here is as it was reviewed, rebased onto main.

## CHANGES

- `--sync-env` on `dr workload config` and `dr workload up` replaces `--import-env` and `--update-env`. One flag brings `.datarobot.yaml` into line with `.env`: a name the manifest does not declare is added, a literal whose value has moved is rewritten, and the credential behind a secret is re-sent.
- The run prints a table of every variable the two files have anything to say about, what it will do to each and why, and asks before it writes a line or sends a value. `--yes` is the only consent that counts; `DATAROBOT_CLI_NON_INTERACTIVE` is not, so an unattended run without `--yes` refuses and names the flag. A dry run prints the table and applies nothing.
- Whether the file will take a rewrite is settled before the table is shown rather than inside the apply, so nothing is asked about a run that was never going to happen.
- A re-sent secret that is all the deploy had to do still leaves the workload serving the old value, because a container reads its credentials at startup. The run says so and names the stop and start that fix it.

## BREAKING

`--import-env` and `--update-env` shipped in the v0.5 releases and are removed here; passing either fails with `unknown flag`. Replace both with `--sync-env`, and note that the run asks before writing, so a script needs `--yes`. The commands sit behind `DATAROBOT_CLI_FEATURE_WORKLOAD=true`.

## RELATED

The CHANGELOG entry here covers the whole of what the three stacked PRs do, not only this one. The Unreleased section is appended to rather than rewritten in the middle of a stack, and the release it is read at ships all three together.

Bottom of a stack of three. The two above it are the one-pass rewrite and the reversal that lets `.env` win outright.

---

> [!NOTE]
> **Medium Risk**
> Breaking CLI flag removal affects scripts; `--sync-env` can rewrite committed manifests and overwrite tenant credentials, with mitigation via preview table and explicit `--yes` only.
> 
> **Overview**
> **`--import-env` and `--update-env` are removed** from `dr workload config` and `dr workload up`; **`--sync-env`** is the single opt-in way to reconcile `.datarobot.yaml` with `.env` (add missing names, rewrite changed literals, re-send secrets). Drift notices and JSON telemetry now point at that one flag.
> 
> **Before any write or credential update**, the wizard builds a per-variable **preview table** (`envplan`) and runs it through shared **`envconfirm`** logic: interactive **Apply?** unless `--yes` or `--dry-run`. **`DATAROBOT_CLI_NON_INTERACTIVE` does not count as consent** for sync—pipelines without a TTY must pass **`--yes`**. Refusing the prompt is a no-op, not a hard failure; deploy can continue with the existing manifest.
> 
> **`editEnv`** runs **`provenRewrite`** before the prompt so shared/anchored manifests that cannot take file edits fail early; updates run before adds; secret-only syncs on shared env still require confirmation. **`dr workload up`** passes the same confirm hook and surfaces **secret re-sends** in the empty-plan “already up to date” message when rotation did not trigger a roll.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit b5bb7a0afd8b3393863ae298978e0a4558257bf4. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


cmd/workload/internal/envconfirm/envconfirm.go
```diff
@@ -0,0 +1,106 @@
...
+	}
+
+	return func() (bool, error) {
+		fmt.Fprintf(stderr, "\n  %s\n", tui.HintStyle.Render(
+			"A rewrite lands in a file you commit, and a re-send overwrites the value on the tenant."))
+		fmt.Fprint(stderr, "\n  Apply this? [y/N] ")
+
+		// One line, like the deploy's own confirmation. An answer that cannot
+		// be read at all is a no, which leaves both files as they are.
+		line, err := bufio.NewReader(stdin).ReadString('\n')
```
adamalpi: **[Low] Two buffered readers over one stdin**

This builds its own `bufio.Reader` on stdin, and `up`'s `typedConfirm` (`cmd/workload/up/cmd.go:424`) builds a `bufio.Scanner` on the same stdin. Both can fire in a single run — the env question, then the locked-production roll — and a buffered reader throws away whatever it read past the line it returned, so the roll prompt can lose the answer that was meant for it.

Narrow in practice, and I want to be straight about how narrow: both gate on `isStdinTerminalFn()`, so a pipe never reaches either prompt. I tried to reproduce it through a pipe for exactly that reason and could not. What is left is type-ahead or a pasted `y⏎ROLL⏎`, where the first read can take both lines.

Threading one reader through both prompts removes the class rather than the instance. Worth it mainly because the cost is a parameter.

Separately, and pre-existing rather than yours: neither prompt goes through `internal/misc/reader`, whose comments describe the cancelreader and Windows console-restoration handling it exists to provide. `typedConfirm` already bypassed it, so this is consistent with its neighbour — noting it only because a third prompt would be the moment to fix the pattern rather than match it.

adamalpi (approved):
Checked the consent model, the CHANGELOG replacement and the BREAKING claim against the v0.5–v0.8 tags; all hold up. One non-blocking note inline.

PR #924, opened 2026-09-15
[RAPTOR-19939] feat(workload): let .env win on the value, the form and the existence

wojtekwdr:
# RATIONALE

The reconcile settled a value and a missing name but never a name you had deleted, so a variable dropped from `.env` went on reaching the container and the table said as much on every run. `.env` now wins on all three questions: the value, the form the entry takes, and whether the entry exists at all.

That reverses the additive, one-directional rule the import followed, which is exactly what the confirmation table two PRs below is for.

Re-lands #889, which went the same way as #888: merged into a base branch that had already been squash-merged into main.

## CHANGES

- A name `.env` no longer mentions is removed from the manifest. A literal it now reads as a secret becomes a credential reference and the reverse becomes a literal, an entry an earlier run left on the credential placeholder is finished, and a value the manifest kept as a mapping or a list becomes the string `.env` has for it.
- The classifier now decides only what form an entry takes and no longer whether the variable travels, so a value it once withheld can land in the manifest in the clear. The run discloses those by name.
- A project with no `.env` is left alone rather than emptied, and so is one whose `.env` declares nothing: a `touch`, or a secrets step that wrote an empty file before a scheduled run passed `--yes` on its behalf, is not an instruction to delete an environment.
- A manifest sharing its environment through a YAML anchor takes no edit at all. Removal and a change of form refuse an entry carrying an anchor anywhere below it rather than only one that is itself shared, because a value another entry reads through an alias would otherwise go out with its entry and leave the alias pointing at nothing in a file that no longer parses.
- The drift notice a bare deploy prints covers removals on the same terms, and the five reports that named what no flag would settle collapse to one. A reconciliation applies all of them now, so they are drift like any other; what is left is the credential whose value cannot be compared at all.

## BREAKING

A name dropped from `.env` is now removed from the manifest instead of being left alone. A stale local `.env` is therefore enough to delete a variable a running workload depends on, which is why the run shows its work and asks, and why `--yes` is the only thing that counts as agreement.

## RELATED

Top of a stack of three. No CHANGELOG change: the entry in the bottom PR covers the stack.

---

> [!NOTE]
> **High Risk**
> Breaking `--sync-env` removes manifest variables absent from `.env`; combined with `--yes`/CI this can delete committed environment config workloads rely on, despite confirmation and empty-file guards.
> 
> **Overview**
> **`--sync-env` is now a full reconciliation: `.env` wins on value, entry shape, and whether a variable exists.** Names dropped from `.env` are removed from the manifest (with preview/confirm), and changes that used to be “no flag applies”—literal↔secret, structured YAML values, finishing credential placeholders—are applied like ordinary drift.
> 
> Reporting and APIs surface **removals separately** (`envNamesRemoved` / `namesRemoved`) so JSON and human output do not look like “only adds.” Deploy drift notices point removals at the same `--sync-env` remedy; wizard planning uses **`syncVars`** (all `.env` names, not classifier “local-only” withholds).
> 
> **Safety rails:** reconciliation is skipped when `.env` is missing or declares no variables; shared/anchored YAML env entries refuse removal or form rebuilds that would break aliases. Empty `.env` files do not trigger mass deletion.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit f9019cba92faee3ef5321041cb13d39025f0fb9d. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


internal/workload/wizard/envplan.go
```diff
@@ -155,73 +159,52 @@ func (o Options) sharedPlan(parsed *manifest.Manifest, detected Detected, wanted
...
+// names, which is the half of a reconciliation that loses configuration. Both
+// the table and the drift notice ask this, so they ask it in one place.
+func droppedNames(parsed *manifest.Manifest, detected Detected) []string {
+	// No file, nothing dropped, and a file naming nothing is the same. An
+	// absent .env names nothing, so every variable would read as removed,
+	// which is the reading that would have a bare deploy in a fresh CI clone
+	// announce the deletion of an environment nobody asked to change; an
+	// empty one is what a touch or a secrets step that wrote nothing leaves,
+	// and the sync it advertises leaves such a file alone.
+	if !fsutil.FileExists(filepath.Join(detected.Dir, EnvFileName)) || len(detected.EnvVars) == 0 {
```
adamalpi, line 220: **[Medium] The guard covers an empty `.env` but not a partly-written one**

This catches an absent `.env` and one declaring zero variables, and the comment above it names exactly why: "a secrets step that wrote an empty file before a scheduled run passed `--yes` on its behalf".

A secrets step that wrote 2 of 50 variables before failing is the same accident one step along, and nothing catches it. It passes `len(detected.EnvVars) == 0`, and under `--yes` the sync removes the other 48 from a running workload's environment. The table still prints, but `--yes` is precisely the mode where nobody reads it — the table is the mitigation everywhere except the one place this fires.

The PR's own BREAKING note is what makes it worth raising rather than shrugging at: "a stale local `.env` is enough to delete a variable a running workload depends on". A partial write is a stale `.env` produced by a mechanism you have already thought about, not by user error.

A proportion check would close it on the same terms the zero-check was written on — refuse, or require something more than `--yes`, when a run would remove most of what the manifest declares. A file that shrank from 50 names to 2 is likelier to be a truncated write than an intended deletion of 48 variables; the reverse reading gets a clear error and one extra flag, which seems the right way round for an irreversible edit to a deployed environment.

internal/workload/wizard/envplan.go
```diff
@@ -28,6 +30,7 @@ const (
 	actAdd    = "add"
 	actUpdate = "update"
 	actResend = "re-send"
+	actRemove = "remove"
```
adamalpi, line 33: **[Medium] Top of a stack that can be lost the way #888 and #889 were**

Anchored here for want of a better line — it is about the PR's base.

This PR's base is `wojtekw/RAPTOR-19939-sync-env-one-pass` (#923), whose own base is #922's branch. That is the shape your rationale identifies as what went wrong before: merged into a base branch that had already been squash-merged into main, so main never received it. I checked the tags — v0.5.0 through v0.8.0 all still ship the old flags, so the loss was real and went unnoticed across four releases.

Being the top of the stack, this is the PR that disappears most quietly if the bases are squash-merged in the wrong order. Worth a stated procedure rather than care: land bottom-up retargeting each base to `main` as the one below merges, or squash all three onto `main` together.

Raised on #923 as well, since both PRs carry a branch base.

adamalpi (approved):
The hard parts — anchor rules, the empty-.env guard, the classifier demotion — are handled well. Two notes inline, one about a gap in the guard.

PR #926, opened 2026-09-15
[RAPTOR-20311] fix(workload): wait until the built image is the one a deploy gets

adamalpi:
# RATIONALE

`--wait` stopped too early. A build reports `COMPLETED` when the image exists, but the artifact only starts pointing at that image a couple of seconds later. Anything that waited and then deployed landed in that gap and got the **previous** build's image — or no image at all.

The platform already publishes `imageApplied` to close this for REST callers. The CLI never read it.

## CHANGES

- **`--wait` now waits for `imageApplied` as well as `COMPLETED`**, so when it returns, deploying is safe. Failed and cancelled builds still stop the wait immediately.
- **If the build finishes but never gets applied, the error says so** instead of a generic timeout that sounds like a slow build.
- **The printed `image_uri` now comes from the build, not from the artifact.** The artifact points at the last-applied build, so inside that same gap it named the wrong image.
- Covers `dr artifact build create --wait`, `dr artifact build get --wait`, and `dr workload up` — all three share one wait.

Minor, spotted while reading the current API schema: `BUILT` is a real build status the CLI didn't know about, and the `UNKNOWN` constant claimed the server never sends it, which is no longer true. Both corrected. Failure reasons from the platform now appear in the summary.

## PR Automation

> [!IMPORTANT]
>

## TESTING

```bash
task test && task lint && task dupcheck
```

All pass. New tests cover the wait polling through `COMPLETED`-but-not-applied, the distinct timeout when it never applies, and `null`/`false`/`true` for `imageApplied`.

## BREAKING

None. `--wait` just takes a couple of seconds longer on success — those seconds were the bug.

## RELATED

- https://datarobot.atlassian.net/browse/RAPTOR-20311 — full analysis, including a schema review confirming nothing else in the CLI is broken by recent `workload-api` changes
- Independent of #918; they only share `CHANGELOG.md`

---

> [!NOTE]
> **Medium Risk**
> Changes shared build-wait logic on the deploy path; behavior is stricter (longer success wait) but well covered by tests and aligns with platform semantics.
> 
> **Overview**
> Fixes **RAPTOR-20311**: `--wait` on artifact builds (and the same wait used by `dr workload up`) no longer treats `COMPLETED` as done. It now polls until the build is **deployable**—`COMPLETED` plus the platform’s `imageApplied`—so scripts that wait and then deploy do not land in the gap where the registry has the new image but the artifact still points at the previous one.
> 
> **Wait behavior:** `FAILED` / `CANCELLED` still fail immediately. If the build finishes but `imageApplied` never becomes true, timeout errors call out that the artifact was not updated rather than a generic slow-build timeout. The `--wait` summary takes `imageUri` from the **build** record (and drops the extra artifact GET), and can include the platform `failureReason` alongside the log tail.
> 
> **Model cleanup:** `BUILT` is an explicit non-terminal status; `Build` gains `imageUri`, `imageApplied`, and helpers that treat null `imageApplied` as “not yet.” Tests cover the COMPLETED-but-not-applied poll window and summary image sourcing.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit c17244b5715545871cec993814e28da6d0a05e9c. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


internal/workload/build.go
```diff
@@ -385,15 +436,27 @@ func WaitForBuild(
...
+		if IsBuildErrorStatus(build.Status) {
+			return build, fmt.Errorf("build %s ended with status %s; run 'dr artifact build logs %s' to inspect", buildID, build.Status, buildID)
+		}
 
+		// COMPLETED is not the end of the wait. It means the image exists;
+		// the artifact is repointed at it a moment later, and a caller that
+		// deploys in between gets the build before this one, or an artifact
+		// with no runtime image at all. Waiting for the artifact to catch up
+		// is the whole point of --wait (RAPTOR-20311).
+		if build.IsDeployable() {
```
wojtekwdr: [high] On a workload-api that doesn't send `imageApplied` yet (no tagged release has it; v11.12.0 doesn't), the field decodes to nil and this is never true. Every successful `build create --wait` and every `dr workload up` that builds then sits through the 30 minute default with nothing new on screen and fails with "not updated", although those servers repoint the artifact inside the build GET itself. Could the wait fall back to checking that the artifact's image ends in `:{buildId}` when `imageApplied` isn't `true`? That's how the server computes the field anyway.

internal/workload/build.go
```diff
@@ -440,15 +505,13 @@ func BuildSummaryFor(build *Build, tailLen int) (BuildSummary, error) {
...
-	}
-
-	summary.ImageURI = GetPrimaryContainerImageURI(*artifact)
+	// From the build, not from the parent artifact. The artifact points at
+	// whichever build was applied last, so fetching it here reported the
+	// previous build's image whenever the summary was composed before the
+	// artifact caught up — the same window --wait now waits through. The
+	// build's own record cannot be wrong about what the build produced,
+	// and reading it costs no second request.
+	summary.ImageURI = build.ImageURI
```
wojtekwdr, line 573: [medium] The build-level `imageUri` shipped separately from `imageApplied` and isn't in a tagged release either, so on a server without it a successful `--wait` now prints no image and the JSON summary has `"imageUri": ""`. Both commands' help text promises the resulting image. Once the wait has seen the image applied, the artifact's primary container names the same image, so falling back to it when `build.ImageURI` is empty would keep the old output there.

internal/workload/build.go
```diff
@@ -385,15 +436,27 @@ func WaitForBuild(
...
+		// is the whole point of --wait (RAPTOR-20311).
+		if build.IsDeployable() {
 			return build, nil
 		}
 
 		if time.Now().After(deadline) {
+			if IsBuildCompleted(build.Status) {
+				return build, fmt.Errorf(
+					"build %s finished but the artifact was not updated to use its image within %s; "+
+						"deploying now would use the previous image. Check 'dr artifact get %s'",
```
wojtekwdr: [low] `dr artifact get` doesn't show the container image in text or JSON, so this hint points at a command that can't confirm or rule out what the error says. Nothing in the CLI surfaces `imageApplied` today, so dropping the pointer may be more honest than naming another command.

CHANGELOG.md
```diff
@@ -18,6 +18,9 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
...
+- `--wait`'s summary reads the resulting `imageUri` off the build rather than off the parent artifact. The artifact points at whichever build was applied last, so for the same reason as above the printed URI could name the previous build's image; it also cost an extra request, since the build record carries its own. `BUILT` is now a named non-terminal status rather than one the poll loop skipped past because it did not recognise it, and the `UNKNOWN` sentinel no longer claims the server never sends it — it is in the platform's enum and is what the status field defaults to.
```
wojtekwdr: [low] `BUILT` was already recognised: `BuildStatusLine` in build_logs.go narrates it as "image built; pushing it to the registry". That's also the server's meaning (build done, push still running, before `COMPLETED`), so the new comment on `BuildStatusBuilt` describing it as the image existing without the artifact pointing at it mixes it up with the window this PR is about.

internal/workload/build_test.go
```diff
@@ -647,6 +723,7 @@ func TestBuildSummaryFor_SuccessSkipsLogs(t *testing.T) {
 		ID:         "b-1",
 		ArtifactID: "art-1",
 		Status:     BuildStatusCompleted,
+		ImageURI:   "ecr/img:tag",
```
wojtekwdr, line 846: [low] The summary tests set `ImageURI` by hand, and nothing decodes `imageUri` or `failureReason` from a server body, so a typo in either JSON tag still passes. Extending `TestGetArtifactBuild_Decodes` with both fields would cover it.

internal/workload/build.go
adamalpi: You are right, and this would have been the worst thing in the PR. Verified before fixing: `v11.12.0`, the latest tag, has **zero** occurrences of `imageApplied` — and its `ImageBuildFormatted` carries none of `imageApplied`, `imageUri` or `failureReason`. So the gate would never be satisfied, and every successful `build create --wait` and every `dr workload up` that builds would have sat through the 30 minute default and then failed, on every released server. Worse than the race it fixes.

Fixed in c6b0395, taking your suggestion. A nil `imageApplied` — the server never sent one, or sent null for "not determined" — falls back to the check the server itself performs: the image tag is the build id, so an artifact whose primary container ends in `:{buildId}` is running that build's image. An explicit `false` is the server answering and is believed rather than second-guessed, so a newer platform is not overridden by a tag comparison.

Worth saying plainly: the ticket described this check ("that is exactly how `imageApplied` is computed server-side") and I did not use it. That is the miss, not the availability of the idea.

Three tests: fallback returns immediately on a released-server shape, the fallback keeps waiting while the tag still names the previous build, and an explicit `false` never reads the artifact at all.

internal/workload/build.go
adamalpi: Confirmed and fixed in c6b0395 — `v11.12.0` has no build-level `imageUri` either, so a successful `--wait` would have printed no image and the JSON summary would have carried `"imageUri": ""`, against help text that promises the resulting image.

The summary now prefers the build's `imageUri` and falls back to the artifact's primary container when it is absent, which is where this always read from. Your reasoning is what makes the fallback safe rather than a reintroduction of the bug: by the time a summary is composed the wait has established the artifact is running this build's image, so on those servers the two name the same thing. Test: `TestBuildSummaryFor_FallsBackToTheArtifactImage`.

internal/workload/build.go
adamalpi: Fixed in c6b0395 by dropping the pointer. You are right that it named a command which cannot answer the question — `dr artifact get` shows no container image in either format, so the reader would run it and be none the wiser.

The error now says the image was built and the artifact is still not using it, and stops there. Nothing in the CLI surfaces `imageApplied` today, so there is no honest command to point at; inventing one would be worse than the silence.

CHANGELOG.md
adamalpi: You are right and I had the evidence in front of me — my own grep returned `build_logs.go:85` and I still wrote that the CLI did not know the status. `BuildStatusLine` has narrated it as "image built; pushing it to the registry" all along, which is the server's meaning: a stage *before* `COMPLETED`, not the window after it that this PR is about. The comment described the wrong thing.

Corrected in c6b0395, in both the constant's comment and the CHANGELOG. I kept the constant, for the narrower reason that the classifiers were relying on the string falling through their unknown-status default — but that is a tidiness argument, not the discovery I wrote it up as.

internal/workload/build_test.go
adamalpi: Fixed in c6b0395. `TestGetArtifactBuild_Decodes` now decodes `imageUri`, `imageApplied` and `failureReason` from the server body and asserts on them, so a typo in any of the three json tags fails there rather than passing everywhere because the summary tests construct the struct by hand.

internal/workload/build.go
```diff
@@ -440,15 +505,13 @@ func BuildSummaryFor(build *Build, tailLen int) (BuildSummary, error) {
...
 	}
 
-	artifact, err := GetArtifact(build.ArtifactID)
-	if err != nil {
-		// Surface the partial summary so callers can still render
-		// duration and status even when the artifact fetch fails.
-		// ImageURI stays empty in that case.
-		return summary, fmt.Errorf("fetch parent artifact for build %s: %w", build.ID, err)
-	}
-
```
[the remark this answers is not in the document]
adamalpi: Correct, and fixed in c6b0395 — though one level up from where this is anchored. The skip is in `cmd/artifact/build/get/cmd.go`, which only called `WaitForBuild` when `!IsTerminalBuildStatus(build.Status)`. COMPLETED is terminal, so `get --wait` returned success without ever reaching the apply check the rest of the PR adds.

Taken the shape of your suggestion rather than the literal line: the guard is now `!IsBuildErrorStatus(build.Status)`, so anything that has not failed goes through `WaitForBuild`. That keeps the deployability logic — including the fallback for servers without `imageApplied` — in one place instead of duplicating it at the call site. An already-deployable build costs one poll that returns immediately.

internal/workload/build.go
```diff
@@ -137,6 +182,12 @@ type BuildSummary struct {
 	DurationSeconds int64           `json:"durationSeconds"`
 	ImageURI        string          `json:"imageUri"`
 	LogTail         []BuildLogEntry `json:"logTail"`
+
+	// FailureReason is the platform's own account of a failure, when it
+	// gave one. Unlike LogTail it is omitted when empty: the log tail has a
+	// stable shape so consumers can jq it either way, while a reason that
+	// was never given is better absent than present and blank.
+	FailureReason string `json:"failureReason,omitempty"`
```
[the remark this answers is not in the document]
adamalpi: Fixed in c6b0395. `RenderBuildSummary` now prints `Reason: <failureReason>` above the log tail on stderr, so text and JSON carry the same facts.

Placed before the tail deliberately: the platform's own reason is the shortest account of why a build failed, and a reader who has to find it among fifty log lines mostly does not.

wojtekwdr (approved):

